PR #31, opened 2023-03-22
OSS CI Updates

prateekdesai04:
***Description of changes:***
Making changes to CI, removing pre-cloned repository pre-OSS.

### DIFF
--- a/graphstorm_job.sh
+++ b/graphstorm_job.sh
@@ -18,6 +18,9 @@ git clone https://github.com/awslabs/graphstorm.git
 
 cd graphstorm
 
+echo "REMOTE: $REMOTE"
+echo "SOURCE-REF: $SOURCE_REF"
+
 if [ ! -z $REMOTE ]; then
     git remote set-url origin $REMOTE
 fi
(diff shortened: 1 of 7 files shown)


graphstorm_job.sh
```diff
@@ -1,30 +1,38 @@
...
+git clone https://github.com/awslabs/graphstorm.git
+
+cd graphstorm
+
+if [ ! -z $REMOTE ]; then
+    git remote set-url origin $REMOTE
+fi
+
+git fetch origin $SOURCE_REF:working
+git checkout working
```
classicsong, line 26: Why checkout working?

classicsong (approved):

graphstorm_job.sh
prateekdesai04: We create a new branch called working, fetch the PR code in that branch and git checkout working switches the current working branch to 'working', which is now pointing to the same commit as the source branch. This allows us to perform subsequent Git operations, such as building or testing the code, on the source branch.

classicsong: But what if the branch name is not working?

prateekdesai04: working is a branch which we create in the github runners after cloning the repository and we checkout code on this 'working' branch and execute tests in this branch

PR #35, opened 2023-03-22
Add support for tuning language model with a separate learning rate

classicsong:
*Issue #, if available:*

*Description of changes:*


python/graphstorm/model/gnn.py
```diff
@@ -40,34 +40,56 @@ class GSOptimizer():
 
     Parameters
     ----------
-    dense_opts : list
-        A list of optimizer objects for dense model parameters.
+    dense_opts : th.optim.Optimizer, list or dict
```
zheng-da, line 43: i don't understand. why do we want to accept a dict and explicitly tell the dense optimizer is for LM or for other models? I don't see how the optimizer wrapper handles them differently.

python/graphstorm/model/gnn.py
```diff
@@ -202,22 +230,40 @@ def __init__(self):
         self._optimizer = None
 
     def get_dense_params(self):
-        """retrieve the all dense layers' parameters as a parameter list.
+        """retrieve the all dense layers' parameters as a parameter list except for
+        language model related parameters.
 
-        TODO(zhengda) we don't need this. We only need to call self.parameters().
```
zheng-da, line 207: actually, we should consider removing this method. why do we need this method? Pytorch's `parameters` can already do the same thing.

python/graphstorm/model/gnn.py
classicsong: Because we need to use different learning rate for dense model and language model.

python/graphstorm/model/gnn.py
classicsong: When you call parameters(), parameters of language models will also be collected.

python/graphstorm/model/gnn.py
```diff
@@ -40,34 +40,56 @@ class GSOptimizer():
...
+        If list, only the first optimizer is treated as the single dense
+        optimizer, the rest optimizers are ignored.
+        If dict, we assume it has multiple optimizers, currently, we support
+        'dense' and 'lm'. The 'dense' usually includes input projection
+        weights, GNN weights, decoder weights, etc. 'lm' is for language
+        models.
     sparse_opts : list
         A list of optimizer objects for sparse model parameters.
     """
     def __init__(self, dense_opts, sparse_opts=None):
```
classicsong, line 55:
```suggestion
    def __init__(self, dense_opts, lm_opts, sparse_opts=None):
```

classicsong: Updated

python/graphstorm/trainer/ep_trainer.py
```diff
@@ -133,10 +136,6 @@ def fit(self, train_loader, n_epochs,
                     num_input_nodes += nodes.shape[0]
 
                 t2 = time.time()
-                if freeze_input_layer_epochs <= i:
-                    self._model.unfreeze_input_encoder()
-                # TODO(xiangsx) Support unfreezing gnn encoder and decoder
```
zheng-da, line 138: this is a bug fix?

classicsong: I created another PR for that.

python/graphstorm/model/gnn.py
```diff
@@ -202,22 +243,40 @@ def __init__(self):
...
+        -------
+        list of Parameters: the dense parameters
         """
         params = []
         if self.gnn_encoder is not None and isinstance(self.gnn_encoder, nn.Module):
             params += list(self.gnn_encoder.parameters())
-        if self.node_input_encoder is not None and isinstance(self.node_input_encoder, nn.Module):
-            params += list(self.node_input_encoder.parameters())
+        if self.node_input_encoder is not None:
+            params += list(self.node_input_encoder.get_general_dense_parameters())
```
zheng-da, line 257: i think it's better check the type of the encoder.

python/graphstorm/model/gnn.py
```diff
@@ -202,22 +243,40 @@ def __init__(self):
...
 
+    def get_lm_params(self):
+        """ get the language model related parameters
+
+        Returns
+        -------
+        list of Parameters: the language model parameters.
+        """
+        params = []
+        if self.node_input_encoder is not None:
```
zheng-da, line 273: the same here.

zheng-da (approved):

PR #38, opened 2023-03-24
Add use_your_own_data example and fix HGT example inconsistency

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
This PR includes two parts:

1. Add a new folder, named "examples/use_your_own_data/" and add a yaml file that can use ACM data as an example of customer data and train on built-in RGCN model.
2. Modify the HGT model's README to fit the OSS structure, move acm_data.py to the examples/ folder.


examples/acm_data.py
```diff
@@ -80,11 +87,14 @@ def create_acm_graph(dowload_path='/tmp/ACM.mat', output_path=None):
     
     parser.add_argument('--download-path', type=str, default='/tmp/ACM.mat',
                         help="The path of folder to store downloaded ACM raw data")
+    parser.add_argument('--dataset-name', type=str, default='acm',
```
classicsong, line 90: Are you going to support other datasets?

classicsong: The name of the example is already acm_data.py I don't think we need this argument.

classicsong (approved):
LGTM. Left one comment.

examples/acm_data.py
zhjwy9343: I provide this argument to let users know that they can change the name.

classicsong: If they change the name, it is still acm, right?

zhjwy9343: yes, the data is still acm data, but the saved file will be <theirname>.dgl. And in the rest steps they need to use their name to replace "acm" in all scripts.

PR #39, opened 2023-03-24
[BugFix] save_topk does save the right model after evaluation.

classicsong:
When both save_model_per_iters and evaluation_frequency are set to the same value, save_topk actually does not save the best model according to the evaluation, as there is a mismatch in its way of counting the iterations.

*Description of changes:*
This PR fixes that bug.


zhjwy9343 (approved):
LGTM

PR #45, opened 2023-03-29
[Fix CI] Fix CI out of disk space problem

classicsong:
We save too much temporal models and outputs during CI 

*Description of changes:*
Remove them temporal models and outputs from disk

### DIFF
--- a/tests/end2end-tests/graphstorm-ec/mgpu_test.sh
+++ b/tests/end2end-tests/graphstorm-ec/mgpu_test.sh
@@ -119,6 +119,8 @@ python3 check_infer.py --train_embout /data/gsgnn_ec/emb/ --infer_embout /data/g
 
 error_and_exit $?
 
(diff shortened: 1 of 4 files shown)


tests/end2end-tests/graphstorm-ec/mgpu_test.sh
```diff
@@ -119,6 +119,8 @@ python3 check_infer.py --train_embout /data/gsgnn_ec/emb/ --infer_embout /data/g
 
 error_and_exit $?
 
+rm -fr /data/gsgnn_ec/*
```
zheng-da, line 122: do you try to remove the files after every test?

classicsong: You are right. Those files will not be used after the test.

zheng-da (approved):

Issue #37, opened 2023-03-23
`predict` does not have `input_nodes` while `forward` has `input_nodes`


zheng-da:
The interface of `predict` and `forward` should be consistent.

PR #53, opened 2023-03-31
[Bug fix] Fix the case when GNN encoder is not used.

classicsong:
When GNN encoder is not used, node and edge classification/regression task will fail due to empty blocks returned by dataloader.

#37 

*Description of changes:*
Fix the bug

### DIFF
--- a/tests/unit-tests/data_utils.py
+++ b/tests/unit-tests/data_utils.py
@@ -21,6 +21,12 @@ import dgl
 import numpy as np
 import torch as th
 import dgl.distributed as dist
+import tempfile
+
+from transformers import AutoTokenizer
+from graphstorm import get_feat_size
+from graphstorm.model.lm_model import TOKEN_IDX, ATT_MASK_IDX, VALID_LEN
+from util import create_tokens
 
 
 def generate_mask(idx, length):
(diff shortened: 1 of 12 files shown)


tests/unit-tests/data_utils.py
```diff
@@ -156,6 +162,66 @@ def generate_dummy_dist_graph(dirname, size='tiny', graph_name='dummy'):
     return partion_and_load_distributed_graph(hetero_graph=hetero_graph, dirname=dirname,
                                               graph_name=graph_name)
 
+def create_lm_graph(tmpdirname):
```
zheng-da, line 165: can you add docstring to explain the difference between the two LM graphs?

tests/unit-tests/test_embed.py
```diff
@@ -203,75 +204,15 @@ def test_compute_embed(dev):
...
-        input_ids1, attention_mask1, g
-
 def test_lm_infer():
     # initialize the torch distributed environment
     th.distributed.init_process_group(backend='gloo',
                                       init_method='tcp://127.0.0.1:23456',
                                       rank=0,
                                       world_size=1)
-    lm_config, feat_size, input_ids, attention_mask, g = create_lm_graph()
+    with tempfile.TemporaryDirectory() as tmpdirname:
```
zheng-da, line 213: is it right that the temporary file will no longer be needed after the graph is constructed?

tests/unit-tests/test_gnn.py
```diff
@@ -93,6 +96,19 @@ def check_node_prediction(model, data):
     assert_almost_equal(pred1[0:len(pred1)].numpy(), pred2[0:len(pred2)].numpy(), decimal=5)
     assert_equal(labels1.numpy(), labels2.numpy())
 
+def check_mlp_node_prediction(model, data):
```
zheng-da, line 99: maybe our test should explain why we want to have this test case.

tests/unit-tests/test_embed.py
classicsong: Once we have g and lm_config, we do not need the graph files.

tests/unit-tests/data_utils.py
classicsong: Updated.

tests/unit-tests/test_gnn.py
classicsong: Agree.
I put some docstring to tests in test_gnn.py

zheng-da: i don't see docstring in this function.

zheng-da (approved):

Issue #57, opened 2023-04-03
Link Prediction's infer evaluated MRR is much larger than trainer evaluated MRR


zhjwy9343:
I tested with MAG data.

For MAG, the best trainer test MRR is around 0.38, but the infer test MRR is about 0.83.

No clue about why the test MRRs have such big difference.

zhjwy9343:
After debugging, found the cause of the difference. It is the difference between the following two arguments.
    num_negative_edges: 4
    num_negative_edges_eval: 200

During training, the evaluator uses "num_negative_edges" for validation and testing. While inference, the evaluator use the "num_negative_edges_eval".

So for a large number of negative edges, the model performs much better.

zheng-da:
is this problem fixed?

classicsong:
The problem is fixed.

PR #58, opened 2023-04-04
[Fix] fix a bug: using the num_negative_edges for test evaluation when do inference

zhjwy9343:
*Issue #57,:*

*Description of changes:*
The LP inference script use "num_negative_edges" for test evaluation, but it should use "num_negative_edges_eval" same as the trainer scripts.
After fix this bug, the MRR values of trainer and infer are similar.


classicsong (approved):
LGTM

PR #66, opened 2023-04-06
Increase `eval_batch_size` to accelerate lp inference

isratnisa:
Resolves issue #81 

This PR updates the default `eval_batch_size` to 10K, replacing the previous value of 1024 as commonly set in most `.yaml` files. The increase in batch size from 1K to 10K results in a significant reduction in end-to-end inference time, from 45 minutes to 19 minutes. In order to enhance the usability of the default value of 10K, the eval_batch_size flag is removed from the `.yaml` files. Users who encounter out-of-memory (OOM) issues can easily adjust the `eval_batch_size` to a lower value as needed.

Following table has the timings for link prediction on OGBN-papers100M (number of nodes = 111M, number of edges = 1.6B). The dataset is partitioned into 3. g4dn.metal instances are used.

#### Environment setup: 
* CUDA 11.3
* DGL 1.0.1
* Pytorch 1.12.0

#### Hyperparameter:
* model_encoder_type: rgcn
* n_layers: 1, 
* n_hidden: 16, 
* fanout: "20", 
* batch_size: 128

| eval_batch_size | Time (min) lp_mini_batch_predict | Time (min) full inference pipeline |
| ---------------- | ---------------------------------- | -------------------------------- |
| 1K  | 31  | 45 |
| 10K  | 5 | 19 |
| 100K  | 2 | 14 |


inference_scripts/lp_infer/lp_infer_gnn.py
```diff
@@ -64,8 +64,15 @@ def main(args):
             'Supported test negative samplers include '
             f'[{BUILTIN_LP_UNIFORM_NEG_SAMPLER}, {BUILTIN_LP_JOINT_NEG_SAMPLER}]')
 
+    # Increase eval batch size to maximize reuse of node embeddings, and also increase
+    # network and GPU efficiency
+    avail_cuda_mem = th.cuda.get_device_properties(0).total_memory - th.cuda.memory_reserved(0)
+    req_cuda_mem_per_edge = config.num_negative_edges_eval * config._n_hidden * 4 * 2
```
classicsong, line 70: Can you explain why we choose *4*2?

isratnisa: 4 is the number of bytes in a float. 2 was for 2 node embeddings for each edge. Anyway, we are not using this solution anymore.

tests/end2end-tests/graphstorm-lp/test.sh
```diff
@@ -32,6 +32,11 @@ python3 -m graphstorm.run.gs_link_prediction --workspace $GS_HOME/training_scrip
 
 error_and_exit $?
 
+echo "**************dataset: Movielens, RGCN layer 1, node feat: fixed HF BERT, BERT nodes: movie, inference: mini-batch, negative_sampler: joint, exclude_training_targets: false, eval_batch_size: 1024, no_validation: false"
```
classicsong, line 35: Can you just modify 1 train test and 1 inference test instead of adding new tests?

classicsong: You can try to change the first 4 tests in mgpu_test.sh by adding --eval-batch-size 1024

tests/unit-tests/test_config.py
```diff
@@ -480,6 +480,7 @@ def test_train_info():
         check_failure(config, "batch_size")
         config._batch_size = 32
         assert config.batch_size == 32
+        config._eval_batch_size = 32
         assert config.eval_batch_size == 32
```
classicsong, line 484:
```suggestion
        assert config.eval_batch_size == 10000
```
You need to test eval_batch_size default return value as 10000.

python/graphstorm/config/argument.py
```diff
@@ -733,7 +733,9 @@ def eval_batch_size(self):
         if hasattr(self, "_eval_batch_size"):
             assert self._eval_batch_size > 0
             return self._eval_batch_size
-        return self.batch_size
+        # (Israt): This number is heuristically chosen. Benchmarks can be found at PR:
+        # https://github.com/awslabs/graphstorm/pull/66.
+        return 10000
```
classicsong, line 738: I think we need to provide more information in our argument doc about the selection of eval_batch_size.

isratnisa: > 

Like the numbers or more reasons?

isratnisa: Added.

tests/end2end-tests/graphstorm-lp/test.sh
isratnisa: Done.

tests/unit-tests/test_config.py
```diff
@@ -480,7 +480,8 @@ def test_train_info():
         check_failure(config, "batch_size")
         config._batch_size = 32
         assert config.batch_size == 32
-        assert config.eval_batch_size == 32
```
classicsong, line 483: You do not need to set it to 10000.
By default, it will be 10000. You are testing its default value.

classicsong (approved):
LGTM

tests/unit-tests/test_config.py
isratnisa: Gotcha!

PR #67, opened 2023-04-06
Asynchronous data loading in mini-batch prediction 

isratnisa:
Load  the node embeddings of different batches asynchronously in the inference pipeline for link prediction.

Update: This optimization is not very impactful anymore as in [this](https://github.com/awslabs/graphstorm/pull/101) PR we are fusing multiple batches to load embeddings from remote machines in one pull.


isratnisa:
This optimization is not impactful anymore as in PR https://github.com/awslabs/graphstorm/pull/101 we are fusing multiple batches to load embeddings from remote machines in one pull.

PR #68, opened 2023-04-06
Add more optimizations in the graph construction pipeline

zheng-da:
*Description of changes:*
This PR adds small optimizations and handles many errors in the input data properly.

### DIFF
--- a/python/graphstorm/gconstruct/construct_graph.py
+++ b/python/graphstorm/gconstruct/construct_graph.py
@@ -388,7 +388,7 @@ def map_node_ids(src_ids, dst_ids, edge_type, node_id_map, skip_nonexist_edges):
     edge_type : tuple
         It contains source node type, relation type, destination node type.
     node_id_map : dict
-        The key is the node type and value is IdMap or IdentityMap.
+        The key is the node type and value is IdMap or NoopMap.
     skip_nonexist_edges : bool
         Whether or not to skip edges whose endpoint nodes don't exist.
 
@@ -466,8 +466,8 @@ def parse_edge_data(in_file, feat_ops, node_id_map, read_file, conf, skip_nonexi
                                     skip_nonexist_edges)
     return (src_ids, dst_ids, feat_data)
 
-class IdentityMap:
-    """ Identity map
+class NoopMap:
+    """ It doesn't map IDs.
 
     This is an identity map. It doesn't do any mapping on IDs.
 
@@ -504,7 +504,7 @@ class IdentityMap:
         return None
 
 class IdMap:
-    """ ID map
+    """ Map an ID to a new ID.
 
     This creates an ID map for the input IDs.
 
@@ -745,7 +745,7 @@ def process_node_data(process_confs, remap_id, num_processes):
                 and np.all(type_node_id_map == np.arange(len(type_node_id_map))) \
                 and not remap_id:
             num_nodes = len(type_node_id_map)
-            type_node_id_map = IdentityMap(num_nodes)
+            type_node_id_map = NoopMap(num_nodes)
         else:
             type_node_id_map = IdMap(type_node_id_map)
             num_nodes = len(type_node_id_map)
@@ -890,8 +890,10 @@ def verify_confs(confs):
     ntypes = {conf['node_type'] for conf in confs["node"]}
     etypes = [conf['relation'] for conf in confs["edge"]]
     for src_type, _, dst_type in etypes:
-        assert src_type in ntypes, f"source node type {src_type} does not exist."
-        assert dst_type in ntypes, f"destination node type {dst_type} does not exist."
+        assert src_type in ntypes, \
+                f"source node type {src_type} does not exist. Please check your input data."
+        assert dst_type in ntypes, \
+                f"dest node type {dst_type} does not exist. Please check your input data."
 
 def process_graph(args):
     """ Process the graph.
--- a/python/graphstorm/gconstruct/construct_graph.py
+++ b/python/graphstorm/gconstruct/construct_graph.py
@@ -85,7 +85,9 @@ def write_data_json(data, data_file):
             for i, val in enumerate(data[key]):
                 records[i][key] = val.tolist()
     with open(data_file, 'w', encoding="utf8") as json_file:
-        json.dump(records, json_file)
+        for record in records:
+            record = json.dumps(record)
+            json_file.write(record + "\n")
 
 def read_data_parquet(data_file, data_fields=None):
     """ Read data from a parquet file.
--- a/python/graphstorm/gconstruct/construct_graph.py
+++ b/python/graphstorm/gconstruct/construct_graph.py
@@ -1,0 +1,732 @@
+"""
+    Copyright 2023 Contributors
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+    Generate example graph data using built-in datasets for node classifcation,
+    node regression, edge classification and edge regression.
+"""
+
+import multiprocessing
+from multiprocessing import Process
+import glob
+import os
+import json
+import argparse
+import pyarrow.parquet as pq
+import pyarrow as pa
+import numpy as np
+
+from transformers import BertTokenizer
+import torch as th
+import dgl
+
+##################### The I/O functions ####################
+
+def read_data_parquet(data_file):
+    """ Read data from the parquet file.
+
+    A row of a multi-dimension data is stored as an object in Parquet.
+    We need to stack them to form a tensor.
+
+    Parameters
+    ----------
+    data_file : str
+        The parquet file that contains the data
+
+    Returns
+    -------
+    dict : map from data name to data.
+    """
+    table = pq.read_table(data_file)
+    data = {}
+    for key, val in table.to_pandas().items():
+        d = np.array(val)
+        # For multi-dimension arrays, we split them by rows and
+        # save them as objects in parquet. We need to merge them
+        # together and store them in a tensor.
+        if d.dtype.hasobject:
+            d = [d[i] for i in range(len(d))]
+            d = np.stack(d)
+        data[key] = d
+    return data
+
+def write_data_parquet(data, data_file):
+    """ Write data in parquet files.
+
+    Normally, Parquet cannot support multi-dimension arrays.
+    This function splits a multi-dimensiion array into N arrays
+    (each row is an array) and store the arrays as objects in the parquet file.
+
+    Parameters
+    ----------
+    data : dict
+        The data to be saved to the Parquet file.
+    data_file : str
+        The file name of the Parquet file.
+    """
+    arr_dict = {}
+    for key in data:
+        arr = data[key]
+        assert len(arr.shape) == 1 or len(arr.shape) == 2, \
+                "We can only write a vector or a matrix to a parquet file."
+        if len(arr.shape) == 1:
+            arr_dict[key] = arr
+        else:
+            arr_dict[key] = [arr[i] for i in range(len(arr))]
+    table = pa.Table.from_arrays(list(arr_dict.values()), names=list(arr_dict.keys()))
+    pq.write_table(table, data_file)
+
+def parse_file_format(fmt):
+    """ Parse the file format blob
+
+    Parameters
+    ----------
+    fmt : dict
+        Describe the file format.
+    """
+    assert 'name' in fmt, "'name' field must be defined in the format."
+    if fmt["name"] == "parquet":
+        return read_data_parquet
+    else:
+        raise ValueError('Unknown file format: {}'.format(fmt['name']))
+
+############## The functions for parsing configurations #############
+
+class Tokenizer:
+    """ A wrapper to a tokenizer.
+
+    It is defined to process multiple strings.
+
+    Parameters
+    ----------
+    tokenizer : a tokenizer
+    max_seq_length : int
+        The maximal length of the tokenization results.
+    """
+    def __init__(self, tokenizer, max_seq_length):
+        self.tokenizer = tokenizer
+        self.max_seq_length = max_seq_length
+
+    def __call__(self, strs):
+        """ Tokenization function.
+
+        Parameters
+        ----------
+        strs : list of strings.
+            The text data to be tokenized.
+
+        Returns
+        -------
+        a dict of tokenization results.
+        """
+        tokens = []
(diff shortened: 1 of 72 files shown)


python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -838,4 +950,6 @@ def process_graph(args):
     argparser.add_argument("--num_partitions", type=int, default=1,
                            help="The number of graph partitions. " + \
                                    "This is only valid if the output format is DistDGL.")
+    argparser.add_argument("--skip_nonexist_edges", action='store_true',
```
thvasilo, line 953: Which are the cases where this happens? May be better to raise an error, depends on what we expect the user intent to be.

tests/unit-tests/test_construct_graph.py
```diff
@@ -116,7 +117,102 @@ def test_label():
     assert np.sum(res['val_mask']) == 1
     assert np.sum(res['test_mask']) == 1
 
+def test_id_map():
```
thvasilo, line 120: Might be better to break out individual tests instead of testing all functionality in one test unit.

tests/unit-tests/test_construct_graph.py
```diff
@@ -116,7 +117,102 @@ def test_label():
...
+    str_ids = np.array([i for i in range(10)])
+    id_map = IdMap(str_ids)
+    try:
+        rand_ids = np.array([str(random.randint(0, len(str_ids))) for _ in range(5)])
+        remap_ids, idx = id_map(rand_ids)
+        raise ValueError("fails")
+    except:
+        pass
+
+def test_map_node_ids():
```
thvasilo, line 161: Same here, unit test maintenance is much easier when they are testing a single-input single-output scenario.

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -409,35 +465,82 @@ def parse_edge_data(in_file, feat_ops, src_id_col, dst_id_col, edge_type,
...
-    dict : the key is the original ID and the value is the new ID.
     """
-    return {id1: i for i, id1 in enumerate(ids)}
+    def __init__(self, ids):
+        self._ids = {id1: i for i, id1 in enumerate(ids)}
+
+    def __len__(self):
+        return len(self._ids)
+
+    def __call__(self, ids):
```
thvasilo, line 512: Not clear what the expected behavior is here, better to add a docstring. Why not make this into a function with a descriptive name instead of using `__call__`?

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -370,8 +372,60 @@ def parse_node_data(in_file, feat_ops, node_id_col, label_conf,
...
+        if skip_nonexist_edges:
+            print(f"dest nodes of {src_type} do not exist: {dst_ids[bool_mask]}")
+        else:
+            raise ValueError(f"dest nodes of {src_type} do not exist: {dst_ids[bool_mask]}")
+        # We need to remove the source nodes as well.
+        src_ids = src_ids[idx]
+    dst_ids = new_dst_ids
+    return src_ids, dst_ids
+
 def parse_edge_data(in_file, feat_ops, src_id_col, dst_id_col, edge_type,
```
thvasilo, line 427: This function now has 9 arguments. Refactoring this will be difficult. Consider creating a wrapper configuration object that groups together relevant arguments (using a `dataclass`) or breaking into multiple functions.

A good rule of thumb (that [pylint](https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/too-many-arguments.html) uses) is for function to not have more than 5 arguments if possible.

A good way to avoid having to pass state around like this is to wrap functions like these in a class that you can initialize with all the shared state.

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -166,8 +166,10 @@ def __call__(self, strs):
             t = self.tokenizer(s, max_length=self.max_seq_length,
                                truncation=True, padding='max_length', return_tensors='pt')
             tokens.append(t['input_ids'])
-            att_masks.append(t['attention_mask'])
-            type_ids.append(t['token_type_ids'])
+            # The masks are small integers. We don't need to use int4 or int8 to store them.
+            # This can signficantly reduce memory consumption.
```
thvasilo, line 170: Did we measure the memory consumed here? How much are we saving by using these representations? (vs. increasing the chances we run into overflow errors)

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -166,8 +166,10 @@ def __call__(self, strs):
             t = self.tokenizer(s, max_length=self.max_seq_length,
                                truncation=True, padding='max_length', return_tensors='pt')
             tokens.append(t['input_ids'])
-            att_masks.append(t['attention_mask'])
-            type_ids.append(t['token_type_ids'])
+            # The masks are small integers. We don't need to use int4 or int8 to store them.
```
classicsong, line 169: `We can use int4 or int8` instead of `We don't need to use int4 or int8` ?

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -370,8 +372,60 @@ def parse_node_data(in_file, feat_ops, node_id_col, label_conf,
             feat_data[key] = val
     return (data[node_id_col], feat_data)
 
+def map_node_ids(src_ids, dst_ids, edge_type, node_id_map, skip_nonexist_edges):
+    """ Map node IDs of source and destination nodes of edges.
+
+    In the ID mapping, we need to handle many errors in the input data. For example,
+    we need to handle the case that endpoint nodes of edges don't exist; we also need
```
classicsong, line 379: Can we directly list how many different errors we handle in this function?
`For example` is vague.

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -370,8 +372,60 @@ def parse_node_data(in_file, feat_ops, node_id_col, label_conf,
...
+        The key is the node type and value is IdMap or IdentityMap.
+    skip_nonexist_edges : bool
+        Whether or not to skip edges whose endpoint nodes don't exist.
+
+    Returns
+    -------
+    tuple of tensors : the remapped source and destination node IDs.
+    """
+    src_type, _, dst_type = edge_type
+    new_src_ids, idx = node_id_map[src_type](src_ids)
```
classicsong, line 401: Does `idx` represent the nodes that exists in node_id_map[src_type] ? Can we give it a better name?

python/graphstorm/gconstruct/construct_graph.py
classicsong: Agree with Theo. A function would be better.

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -409,35 +465,82 @@ def parse_edge_data(in_file, feat_ops, src_id_col, dst_id_col, edge_type,
...
+                assert np.issubdtype(ids.dtype, np.integer), \
+                        "The key of ID map is integer, input IDs should also be integers. " \
+                        + f"But get {type(ids[0])}."
+            else:
+                raise ValueError(f"Unsupported key data type: {type(id_)}")
+            break
+
+        new_ids = []
+        idx = []
+        for i, id_ in enumerate(ids):
```
classicsong, line 530: Can you add some docstring describing what you are doing?

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -755,6 +866,15 @@ def process_edge_data(process_confs, node_id_map, num_processes):
 
     return edges, edge_data
 
+def verify_confs(confs):
+    """ Verify the configuration of the input data.
+    """
+    ntypes = {conf['node_type'] for conf in confs["node"]}
+    etypes = [conf['relation'] for conf in confs["edge"]]
+    for src_type, _, dst_type in etypes:
```
classicsong, line 874: What if a node type does not have any node feature? Are we still ask users to provide node files for that node type?

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -802,18 +909,23 @@ def process_graph(args):
...
     for ntype in node_id_map:
-        map_data = {}
-        map_data["orig"] = np.array(list(node_id_map[ntype].keys()))
-        map_data["new"] = np.array(list(node_id_map[ntype].values()))
-        write_data_parquet(map_data, os.path.join(args.output_dir, ntype + "_id_remap.parquet"))
+        kv_pairs = node_id_map[ntype].get_key_vals()
+        if kv_pairs is not None:
+            map_data = {}
+            map_data["orig"], map_data["new"] = kv_pairs
+            write_data_parquet(map_data, os.path.join(args.output_dir,
```
classicsong, line 927: Can user choose the data format?
Or can we use the same format as input data?

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -802,18 +909,23 @@ def process_graph(args):
...
+                                                      ntype + "_id_remap.parquet"))
 
 if __name__ == '__main__':
     argparser = argparse.ArgumentParser("Preprocess graphs")
```
classicsong, line 931: We should also make a doc describing all the input arguments of construct_graph.

tests/unit-tests/test_construct_graph.py
classicsong, line 120: Please add a description for this test.

tests/unit-tests/test_construct_graph.py
```diff
@@ -116,7 +117,102 @@ def test_label():
...
+    # Test the case that destination node IDs don't exist.
+    src_ids = np.array([str(random.randint(0, len(str_src_ids) - 1)) for _ in range(15)])
+    dst_ids = np.array([str(random.randint(0, 20)) for _ in range(15)])
+    try:
+        new_src_ids, new_dst_ids = map_node_ids(src_ids, dst_ids, ("src", None, "dst"),
+                                                id_map, False)
+        raise ValueError("fail")
+    except:
+        pass
+    # Test the case that destination node IDs don't exist and we skip non exist edges.
```
classicsong, line 206: Shall we split it into multiple small unitests?

python/graphstorm/gconstruct/construct_graph.py
zheng-da: some data (e.g., MAG graph) has this problem that some of the end point nodes don't exist in the node files. this option is to tell the pipeline to ignore these edges. by default, it raises an error. however, i would like to give users the option of skipping these edges. otherwise, the users have to deal with this problem themselves.

python/graphstorm/gconstruct/construct_graph.py
zheng-da: i try to understand the problem of this. is it about the number of arguments or the amount of work done in this function?
if we wrap multiple arguments into an object, it doesn't change anything except the number of arguments.

we cannot break this function because this is a entry function that runs in process pool.

zheng-da: i guess the easiest solution is to pass `process_conf` to the function directly.

python/graphstorm/gconstruct/construct_graph.py
zheng-da: a lot. Memory consumption reduces by almost 3 times.

zheng-da: attention mask is bool. token type ids are small numbers according to the documentation of huggingface.

python/graphstorm/gconstruct/construct_graph.py
zheng-da: node location in `src_ids`

python/graphstorm/gconstruct/construct_graph.py
zheng-da: yes, we currently require users to provide node files for every node type. computing unique node IDs is very expensive operation. it requires appropriate optimization. I don't think I'll support this in v0.1. I want to see users' feedbacks first.

python/graphstorm/gconstruct/construct_graph.py
zheng-da: not sure what is the best option here. maybe we should allow users to provide an argument to specify the format. or maybe JSON is the best format here.

python/graphstorm/gconstruct/construct_graph.py
zheng-da: yes, just like training/inference script.

tests/unit-tests/test_construct_graph.py
zheng-da: do you mean split it into multiple `check_xxx` functions?

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -409,35 +462,109 @@ def parse_edge_data(in_file, feat_ops, src_id_col, dst_id_col, edge_type,
...
+        Returns
+        -------
+        tuple of tensors : the tensor of new IDs, the location of the IDs in the input ID tensor.
+        """
+        return ids, np.arange(len(ids))
+
+    def get_key_vals(self):
+        """ Get the key value pairs.
+        """
+        return None
```
thvasilo, line 504: What is the purpose of having this function if it just returns `None`? If we want to have the same exact interface for "IdentityMap" and "IdMap" they should probably belong to a class hierarchy with a common interface.

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -409,35 +462,109 @@ def parse_edge_data(in_file, feat_ops, src_id_col, dst_id_col, edge_type,
...
+                assert np.issubdtype(ids.dtype, np.integer), \
+                        "The key of ID map is integer, input IDs should also be integers. " \
+                        + f"But get {type(ids[0])}."
+            else:
+                raise ValueError(f"Unsupported key data type: {type(id_)}")
+            break
+
+        # If the input ID exists in the ID map, map it to a new ID
+        # and keep its location in the input ID array.
+        # otherwise, skip the ID.
```
thvasilo, line 551:
```suggestion
        # Otherwise, skip the ID.
```

thvasilo (commented):
A couple of minor changes requested, the logic seems fine.

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -755,6 +884,15 @@ def process_edge_data(process_confs, node_id_map, num_processes):
 
     return edges, edge_data
 
+def verify_confs(confs):
+    """ Verify the configuration of the input data.
+    """
+    ntypes = {conf['node_type'] for conf in confs["node"]}
+    etypes = [conf['relation'] for conf in confs["edge"]]
+    for src_type, _, dst_type in etypes:
+        assert src_type in ntypes, f"source node type {src_type} does not exist."
```
classicsong, line 893: Can we also tell user how to fix the problem or what is the root cause of the problem? For example, they do not add src_node_type in the config.

python/graphstorm/gconstruct/construct_graph.py
classicsong: I agree.

classicsong (approved):
LGTM.
But please resolve Theo's comments.

Issue #28, opened 2023-03-18
Have a scalable way of saving learnable embeddings


zheng-da:
Currently, trainer 0 saves the learnable embeddings to the disk. If the learnable embedding table is very large, trainer 0 will run out of memory. We need to distribute the saving of learnable embeddings.

PR #70, opened 2023-04-06
Save and Load learnable sparse embedding in a distributed manner

classicsong:
This PR saves and loads learnable sparse embeddings in a distributed manner. Each trainer (GPU) will be responsible to saving part of sparse embeddings. We assume the data is writen into a shared file system.

#28 

### DIFF
--- a/python/graphstorm/model/utils.py
+++ b/python/graphstorm/model/utils.py
@@ -110,6 +110,15 @@ def _get_sparse_emb_range(num_embs, local_rank, world_size):
 def save_sparse_embeds(model_path, embed_layer, local_rank, world_size):
     """ save sparse embeddings if any
 
+        Sparse embeddings are stored as:
+        $model_path/ntype0/sparse_emb_0.pt
+                           ...
+                           sparse_emb_N.pt
+        $model_path/ntype1/sparse_emb_0.pt
+                           ...
+                           sparse_emb_N.pt
+        ...
+
         Parameters
         ----------
         model_path: str
@@ -296,6 +305,15 @@ def load_model(model_path, gnn_model=None, embed_layer=None, decoder=None):
 
 def load_sparse_embeds(model_path, embed_layer, local_rank, world_size):
     """load sparse embeddings if any
+
+        Sparse embeddings are stored as:
+        $model_path/ntype0/sparse_emb_0.pt
+                           ...
+                           sparse_emb_N.pt
+        $model_path/ntype1/sparse_emb_0.pt
+                           ...
+                           sparse_emb_N.pt
+        ...
 
         Parameters
         ----------
--- a/python/graphstorm/model/utils.py
+++ b/python/graphstorm/model/utils.py
@@ -153,14 +153,7 @@ def save_sparse_embeds(model_path, embed_layer, local_rank, world_size):
 
             embs = th.cat(embs, dim=0)
             emb_path = os.path.join(model_path, ntype)
-            try:
-                if not os.path.exists(emb_path):
-                    os.mkdir(emb_path)
-            except OSError as e:
-                # if directory exists, ignore the error
-                if e.errno != errno.EEXIST:
-                    raise
-
+            os.makedirs(emb_path, exist_ok=True)
             th.save(embs, os.path.join(emb_path,
                                        f'sparse_emb_{local_rank}.pt'))
 
--- a/python/graphstorm/model/utils.py
+++ b/python/graphstorm/model/utils.py
@@ -19,6 +19,7 @@ import os
 import math
 import json
 import shutil
+import errno
 
 import torch as th
 from torch import nn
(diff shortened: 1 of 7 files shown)


classicsong:
@zhjwy9343 Can you take look of this PR? the path to save sparse embeddings is slightly changed.

python/graphstorm/model/utils.py
```diff
@@ -263,41 +302,71 @@ def load_model(model_path, gnn_model=None, embed_layer=None, decoder=None):
...
+                    # Get the target idx range for sparse_emb_{local_rank}.pt
+                    start, end = _get_sparse_emb_range(num_embs,
+                                                       local_rank=file_idx,
+                                                       world_size=num_files)
+
+                    # write sparse_emb back in an iterative way
+                    batch_size = 10240
+                    idxs = th.split(th.arange(end - start), batch_size, dim=0)
+                    for idx in idxs:
+                        # TODO: dgl.distributed.DistEmbedding should allow some basic tensor ops
```
zheng-da, line 355: what operations do you want?

python/graphstorm/model/utils.py
```diff
@@ -263,41 +302,71 @@ def load_model(model_path, gnn_model=None, embed_layer=None, decoder=None):
...
+        world_size : int
+            World size in a distributed env.
     """
     if embed_layer is None:
         return
     embed_layer = embed_layer.module \
         if isinstance(embed_layer, DistributedDataParallel) else embed_layer
 
     if len(embed_layer.sparse_embeds) > 0:
+        def load_sparse_emb(num_embs, ntype_path):
```
zheng-da, line 334: it's worthwhile to have a unit test on this function. it's better to implement it as a top function in the script.

python/graphstorm/trainer/gsgnn_trainer.py
```diff
@@ -186,12 +187,15 @@ def save_model(self, model, epoch, i, save_model_path):
...
         '''
         th.distributed.barrier()
-        if save_model_path is not None and self.rank == 0:
+        if save_model_path is not None:
+            assert isinstance(model.module, GSgnnModel), \
+                "Please make sure the model derives from GSgnnModel, " \
+                "which provides a scalable model saving implementation."
             save_model_path = self._gen_model_path(save_model_path, epoch, i)
             model.module.save_model(save_model_path)
             self.optimizer.save_opt_state(save_model_path)
```
zheng-da, line 196: do you need to change anything in optimizer to save the states of the learnable embeddings?

python/graphstorm/model/utils.py
```diff
@@ -87,37 +88,73 @@ def save_model_results_json(conf, test_model_performance, save_perf_results_path
         with open(json_path, 'w', encoding='utf-8') as f:
             json.dump(model_results_and_conf, f, ensure_ascii=False, indent=4)
 
-def save_sparse_embeds(model_path, embed_layer):
+def _get_sparse_emb_range(num_embs, local_rank, world_size):
```
zheng-da, line 91: we should have a unit test to test this function.

python/graphstorm/trainer/gsgnn_trainer.py
classicsong: We do not support loading the states of the learnable embeddings: https://github.com/awslabs/graphstorm/blob/main/python/graphstorm/model/utils.py#L336-L337. 
Let me add an issue about it. I will not support this in this PR.

python/graphstorm/model/utils.py
classicsong: like dtype, shape, etc.

python/graphstorm/model/utils.py
classicsong: Added

tests/unit-tests/test_model_save_load.py
```diff
@@ -0,0 +1,199 @@
...
+        os.mkdir(model_path)
+        saved_embs = \
+            {ntype: sparse_emb._tensor[th.arange(embed_layer.g.number_of_nodes(ntype))] \
+                for ntype, sparse_emb in embed_layer.sparse_embeds.items()}
+        #
+        for i in range(train_world_size):
+            save_sparse_embeds(model_path, embed_layer, i, train_world_size)
+
+        for i in range(infer_world_size):
+            load_sparse_embeds(model_path, embed_layer, i, infer_world_size)
```
zheng-da, line 133: it's not clear to me whether you have tested the following three cases:

# 1. N == K
# 2. N > K, some trainers/infers need to load more than one files
# 3. N < K, some trainers/infers do not need to load any files

classicsong: We have 
`
@pytest.mark.parametrize("infer_world_size", [3, 8, 16])
@pytest.mark.parametrize("train_world_size", [8])
`
So the combinations will be 
N = 8  and K = 3
N = 8  and K = 8
N = 8  and K = 16
To test different cases.

zheng-da (approved):

PR #71, opened 2023-04-07
Fix numpy split error: ValueError: array split does not result in an equal division

classicsong:
See error here: https://github.com/awslabs/graphstorm/actions/runs/4641522052/jobs/8214741442?pr=70

*Description of changes:*
use numpy.array_split instead of numpy.split.


tests/end2end-tests/data_process/data_gen.py
```diff
@@ -87,7 +87,7 @@ def write_data_parquet(data, data_file):
 def split_data(data, num):
     new_data_list = [{} for _ in range(num)]
     for key, val in data.items():
-        for i, val in enumerate(np.split(val, num)):
+        for i, val in enumerate(np.array_split(val, num)):
```
zheng-da, line 90: when does this error happen?

classicsong: You can take a look at this error log: https://github.com/awslabs/graphstorm/actions/runs/4641522052/jobs/8214741442?pr=70

classicsong: It happens when `node_id1 = np.unique(np.random.randint(0, 1000000000, 10000))` does not return a tensor with shape (10000,). as you use np.unique here.

zheng-da (approved):

PR #72, opened 2023-04-08
Add JSON support

zheng-da:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/python/graphstorm/gconstruct/construct_graph.py
+++ b/python/graphstorm/gconstruct/construct_graph.py
@@ -57,7 +57,10 @@ def read_data_json(data_file, data_fields):
     dict : map from data name to data
     """
     with open(data_file, 'r', encoding="utf8") as json_file:
-        data_records = json.load(json_file)
+        data_records = []
+        for line in json_file.readlines():
+            record = json.loads(line)
+            data_records.append(record)
 
     data = {key: [] for key in data_fields}
     for record in data_records:
--- a/python/graphstorm/gconstruct/construct_graph.py
+++ b/python/graphstorm/gconstruct/construct_graph.py
@@ -39,8 +39,36 @@ from graphstorm.utils import sys_tracker
 
 ##################### The I/O functions ####################
 
+def read_data_json(data_file, data_fields):
+    """ Read data from a JSON file.
+
+    Each row of the JSON file represents a data record. The function tries
+    to extract a set of values from the data record.
+
+    Parameters
+    ----------
+    data_file : str
+        The JSON file that contains the data.
+    data_fields : list of str
+        The data fields that we will read data from a data record.
+
+    Returns
+    -------
+    dict : map from data name to data
+    """
+    with open(data_file, 'r', encoding="utf8") as json_file:
+        data_records = json.load(json_file)
+
+    data = {key: [] for key in data_fields}
+    for record in data_records:
+        for key in data_fields:
+            data[key].append(record[key])
+    for key in data:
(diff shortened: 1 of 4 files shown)


python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -39,8 +39,58 @@
...
+    with open(data_file, 'r', encoding="utf8") as json_file:
+        data_records = []
+        for line in json_file.readlines():
+            record = json.loads(line)
+            data_records.append(record)
+
+    data = {key: [] for key in data_fields}
+    for record in data_records:
+        for key in data_fields:
+            data[key].append(record[key])
```
classicsong, line 68: How do we distinguish labels and features?

tests/end2end-tests/data_process/data_gen.py
```diff
@@ -56,6 +44,8 @@ def write_data_parquet(data, data_file):
 node_id3_str = np.array([str(nid) for nid in node_id3])
 node_data3 = {
```
classicsong, line 45: Can you add some docstring about how does the test graph looks like?

tests/unit-tests/test_construct_graph.py
```diff
@@ -41,6 +41,22 @@ def test_parquet():
     assert "data2" not in data1
     assert np.all(data1['data1'] == data['data1'])
 
+def test_json():
+    from graphstorm.gconstruct import write_data_json, read_data_json
+    handle, tmpfile = tempfile.mkstemp()
```
classicsong, line 46: Will the tmpfile be deleted after the test?

python/graphstorm/gconstruct/construct_graph.py
zheng-da: for read/write functions, we don't distinguish them.

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -39,8 +39,58 @@
...
+    for key in data:
+        data[key] = np.array(data[key])
+    return data
+
+def write_data_json(data, data_file):
```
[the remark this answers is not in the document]
zheng-da: i think this is for consistency. i plan to move all read/write functions to another file in another PR.

tests/end2end-tests/data_process/data_gen.py
```diff
@@ -96,7 +86,7 @@ def split_data(data, num):
 for i, node_data in enumerate(split_data(node_data2, 10)):
     write_data_parquet(node_data, os.path.join(in_dir, f'node_data2_{i}.parquet'))
 for i, node_data in enumerate(split_data(node_data3, 10)):
-    write_data_parquet(node_data, os.path.join(in_dir, f'node_data3_{i}.parquet'))
+    write_data_json(node_data, os.path.join(in_dir, f'node_data3_{i}.json'))
```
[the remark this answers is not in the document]
zheng-da: i want to make test all different cases. I just pick node3 to store the data with json. it's not necessary to store all node data with the same format.

classicsong (approved):

tests/end2end-tests/data_process/data_gen.py
zheng-da: i don't think there is any particular purpose of each node type. I create multiple node types to handle different cases, for example, to store node data with different formats, with or without node features, with or without labels, features of different types, node features with different dimensions. you can easily see this kind of information in the test code when node data are defined. I'm not sure what information should be documented.

PR #74, opened 2023-04-10
Graph construction with external memory support

zheng-da:
*Description of changes:*
When constructing a graph with large features, we can store the node/edge features in external memory during the graph construction. This allows us to construct a much larger graph.

### DIFF
--- a/python/graphstorm/gconstruct/construct_graph.py
+++ b/python/graphstorm/gconstruct/construct_graph.py
@@ -933,7 +933,7 @@ def process_edge_data(process_confs, node_id_map, convert2ext_mem,
             # If we allow to store features in external memory, we store edge features
             # that have large feature dimensions in a file and use memmap to access
             # the array.
-            etype_str = edge_type[0] + "-" + edge_type[1] + "-" + edge_type[2]
+            etype_str = "-".join(edge_type)
             type_edge_data[feat_name] = convert2ext_mem(type_edge_data[feat_name],
                                                         etype_str + "_" + feat_name)
             assert len(type_edge_data[feat_name]) == len(type_src_ids)
--- a/python/graphstorm/gconstruct/construct_graph.py
+++ b/python/graphstorm/gconstruct/construct_graph.py
@@ -978,8 +978,10 @@ def partition_graph(g, node_data, edge_data, graph_name, num_partitions, output_
         g.nodes[ntype].data['orig_id'] = th.arange(g.number_of_nodes(ntype))
     for etype in g.canonical_etypes:
         g.edges[etype].data['orig_id'] = th.arange(g.number_of_edges(etype))
+    sys_tracker.check('Before partitioning starts')
     dgl.distributed.partition_graph(g, graph_name, num_partitions, output_dir,
                                     part_method="None" if num_partitions == 1 else "metis")
+    sys_tracker.check('Graph partitioning')
     for i in range(num_partitions):
         part_dir = os.path.join(output_dir, "part" + str(i))
         data = dgl.data.utils.load_tensors(os.path.join(part_dir, "node_feat.dgl"))
@@ -988,6 +990,7 @@ def partition_graph(g, node_data, edge_data, graph_name, num_partitions, output_
             del data[ntype + "/orig_id"]
             for name, ndata in node_data[ntype].items():
                 data[ntype + "/" + name] = th.tensor(ndata[orig_ids])
+            sys_tracker.check(f'Get node data of node {ntype} in partition {i}')
         dgl.data.utils.save_tensors(os.path.join(part_dir, "node_feat.dgl"), data)
 
         data = dgl.data.utils.load_tensors(os.path.join(part_dir, "edge_feat.dgl"))
@@ -996,6 +999,7 @@ def partition_graph(g, node_data, edge_data, graph_name, num_partitions, output_
             del data[_etype_tuple_to_str(etype) + '/orig_id']
             for name, edata in edge_data[etype].items():
                 data[_etype_tuple_to_str(etype) + "/" + name] = th.tensor(edata[orig_ids])
+            sys_tracker.check(f'Get edge data of edge {etype} in partition {i}')
         dgl.data.utils.save_tensors(os.path.join(part_dir, "edge_feat.dgl"), data)
 
 def process_graph(args):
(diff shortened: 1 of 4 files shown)


thvasilo (commented):
As we've now hit 1k lines of code in this file let's consider how to modularize the code. Can move out general functions (e.g multi-process related) to another module, wrap non-public functions in a class, create a config class to avoid argument proliferation etc.

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -646,7 +646,32 @@ def close(self):
...
+    # We need to create the workspace directory if it doesn't exist.
+    os.makedirs(ext_mem_workspace, exist_ok=True)
+    tensor_path = os.path.join(ext_mem_workspace, name + ".npy")
+    em_arr = np.memmap(tensor_path, arr.dtype, mode="w+", shape=arr.shape)
+    em_arr[:] = arr[:]
+    return em_arr
+
+def process_node_data(process_confs, remap_id, num_processes,
+                      ext_mem_workspace=None,
+                      ext_mem_feat_size=64):
```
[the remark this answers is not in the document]
thvasilo: OK, as long as we agree that we expect users to call `process_node_data` and `process_edge_data` we can treat them as top-level entry points functions, and as such they can be expected  to have lots of arguments.

My previous understanding was that since this is a script the users would only use its main function as an entry point and not import these as library functions.

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -646,7 +646,32 @@ def close(self):
         for proc in self.processes:
             proc.join()
 
-def process_node_data(process_confs, remap_id, num_processes):
+def convert2ext_mem(arr, ext_mem_workspace, name):
```
classicsong, line 649: Will you remove those files? Or they will be automatically removed after the process exits.

tests/end2end-tests/data_process/test.sh
```diff
@@ -36,6 +36,16 @@ python3 $GS_HOME/tests/end2end-tests/data_process/test_data.py --graph_format Di
 
 error_and_exit $?
 
+# Test the DistDGL graph format with external memory support.
+echo "********* Test the DistDGL graph format with external memory support ********"
+python3 -m graphstorm.gconstruct.construct_graph --conf_file /tmp/test_data/test_data_transform.conf --num_processes 4 --output_dir /tmp/test_partition2 --graph_name test --ext_mem_workspace /tmp --ext_mem_feat_size 2
```
classicsong, line 41: Can you add a check to see whether the output of using ext_mem_workspace and with using ext_mem_workspace are same?

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -753,6 +782,14 @@ def process_node_data(process_confs, remap_id, num_processes):
 
         for feat_name in type_node_data:
             type_node_data[feat_name] = np.concatenate(type_node_data[feat_name])
```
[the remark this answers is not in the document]
zheng-da: we can support using multiple files to store the data afterwards. that requires much more complex implementations.

python/graphstorm/gconstruct/construct_graph.py
zheng-da: currently, i don't remove them. let me change the code to remove it.

tests/end2end-tests/data_process/test.sh
zheng-da: what is the question?

classicsong: Compare the dgl graphs you build with and without using ext_mem_workspace and check if there is any mismatch.
In theory, using ext_mem_workspace will not change the constructed dgl graph.

zheng-da: it's a bit difficult to compare. graph partition has some randomness.

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -646,7 +646,55 @@ def close(self):
...
+        The path of the directory where the array will be stored.
+    ext_mem_feat_size : int
+        The threshold of the feature size that triggers storing data on disks.
+    """
+    def __init__(self, ext_mem_workspace, ext_mem_feat_size):
+        self._ext_mem_workspace = ext_mem_workspace
+        self._ext_mem_feat_size = ext_mem_feat_size
+        self._tensor_files = []
+
+    def cleanup(self):
```
classicsong, line 664: Can we directly overload __del__() ?

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -871,6 +930,12 @@ def process_edge_data(process_confs, node_id_map, num_processes, skip_nonexist_e
 
         for feat_name in type_edge_data:
             type_edge_data[feat_name] = np.concatenate(type_edge_data[feat_name])
+            # If we allow to store features in external memory, we store edge features
+            # that have large feature dimensions in a file and use memmap to access
+            # the array.
+            etype_str = edge_type[0] + "-" + edge_type[1] + "-" + edge_type[2]
```
classicsong, line 936:
```suggestion
            etype_str = "-".join(edge_type)
```

tests/unit-tests/test_construct_graph.py
```diff
@@ -229,7 +231,84 @@ def test_map_node_ids():
...
+    converters = [ExtMemArrayConverter(None, 0),
+                  ExtMemArrayConverter("/tmp", 2)]
+    for converter in converters:
+        arr = np.array([str(i) for i in range(10)])
+        em_arr = converter(arr, "test1")
+        np.testing.assert_array_equal(arr, em_arr)
+
+        arr = np.random.uniform(size=(1000, 10))
+        em_arr = converter(arr, "test2")
+        np.testing.assert_array_equal(arr, em_arr)
```
classicsong, line 245: You can also check whether the file exists before converter.cleanup() and whether the file is deleted after converter.cleanup().

classicsong (approved):
LGTM

python/graphstorm/gconstruct/construct_graph.py
zheng-da: do we prefer to override del?

tests/unit-tests/test_construct_graph.py
zheng-da: this is difficult. i don't know their file names.

Issue #51, opened 2023-03-30
The preprocessing script support data split for link prediction.


zheng-da:
Currently, the labels on edges must be classification task.

PR #76, opened 2023-04-10
support different tasks in graph construction.

zheng-da:
*Description of changes:*
We need to support more tasks in graph construction.
https://github.com/awslabs/graphstorm/issues/51


python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -336,37 +356,74 @@ def process_labels(data, label_confs):
...
+        if valid_label_idx is None:
+            rand_idx = np.random.permutation(num_samples)
+        else:
+            rand_idx = np.random.permutation(valid_label_idx)
+        num_labels = len(rand_idx)
+        train_idx = rand_idx[0:int(num_labels * train_split)]
+        val_start = int(num_labels * train_split)
+        val_end = int(num_labels * (train_split + val_split))
         val_idx = rand_idx[val_start:val_end]
         test_idx = rand_idx[val_end:]
```
zhjwy9343, line 411: So here test_split does not take effect. We assume the rest will all become test set?

zhjwy9343, line 411: ditto here

zheng-da: this is to split the data with valid labels. What should we do with the rest of labelled samples?

zhjwy9343: I mean if users give [0.5, 0.1, 0.1]. It means 50% for train, 10% for val, and 10% for test. There will be 30% unused. Base on the current logic. The test set will be 40%, which is not users want.

zheng-da: i see. i'll fix it.

zhjwy9343 (approved):
LGTM

Issue #20, opened 2023-03-11
Verify the correctness of GraphStorm input config.


zheng-da:
* Don't require users to provide graph name.
* Verify the node/edge feature names and label names.
* If infer_data.test_idxs is empty, return an error in the inference script.
* We may want to warn users if they want to have mini-batch inference. (mini-batch inference on a large graph can take very long time).

PR #84, opened 2023-04-12
BugFix: evaluation_frequency needs to work with save_model_per_iters set to -1

classicsong:
By default save_model_per_iters is set to -1 which means GFS will not save models within an epoch. In this case, evaluation_frequency should be able to use any positive integer.

#20 

*Description of changes:*
Fix the bug


zhjwy9343 (approved):
LGTM

PR #85, opened 2023-04-12
[Fix] Test graph construction and fix a few naming bugs

zhjwy9343:
*Issue #, if available:*
The construct_graph.py has some naming bugs. 

*Description of changes:*

- Add a new dummy customer data generation codes

- Fix a few naming issues in construct_grapy.py and data_gen.py.

### DIFF
--- a/tests/end2end-tests/data_process/gen_test_graphs.py
+++ b/tests/end2end-tests/data_process/gen_test_graphs.py
@@ -115,6 +115,8 @@ def generate_data(args):
     e2_df = pd.DataFrame({'src_id': e2_src_ids,
                           'dst_id': e2_dst_ids})
     e2_df = e2_df.drop_duplicates().iloc[ :NUM_ETYPE2 * factor]
+    
+    # [TODO] James: Keep these commented codes for later test after construct graph fix this issue.
     # elabels = np.random.randint(0, 2, int(e2_df.shape[0] * 0.5))      # labels for half edges
     # e2_df['f1-label'] = np.nan
     # e_idx = np.random.randint(0, e2_df.shape[0], int(e2_df.shape[0] * 0.5))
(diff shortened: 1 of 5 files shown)


tests/end2end-tests/data_process/gen_test_graphs.py
```diff
@@ -0,0 +1,374 @@
...
+    e_dfs_dict[('n0', 'e0', 'n1')] = e1_df
+
+    # -------- build edge2 between node1 and node3 ---------
+    e2_src_ids = np.random.choice(n1_ids, NUM_ETYPE2 * factor + 100, replace=True)
+    e2_dst_ids = np.random.choice(n3_ids, NUM_ETYPE2 * factor + 100, replace=True)
+    e2_df = pd.DataFrame({'src_id': e2_src_ids,
+                          'dst_id': e2_dst_ids})
+    e2_df = e2_df.drop_duplicates().iloc[ :NUM_ETYPE2 * factor]
+    
+    # [TODO] James: Keep these commented codes for later test after construct graph fix this issue.
```
zheng-da, line 119: what is the issue?

tests/end2end-tests/data_process/gen_test_graphs.py
```diff
@@ -0,0 +1,374 @@
+"""
```
zheng-da, line 1: this script is not used in CI. I wonder if we need to have such a file checked in the test folder.

zheng-da: we already have a script that generates a test graph. https://github.com/awslabs/graphstorm/blob/main/tests/end2end-tests/data_process/data_gen.py
we run the pipeline on the test graph and verify the correctness of the partition result.

zhjwy9343: This script will be used for regression test.

zheng-da (approved):

tests/end2end-tests/data_process/gen_test_graphs.py
zhjwy9343: This is the issue about only partial nodes have labels.

PR #88, opened 2023-04-13
Fix a bug in pip install

zheng-da:
*Issue #, if available:*

*Description of changes:*
We should install `scikit-learn` instead of `sklearn`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ requirements = [
     'transformers>=4.3.0',
     'pandas',
     'pylint',
-    'sklearn',
+    'scikit-learn',
     'ogb',
     'psutil',
 ]


setup.py
```diff
@@ -34,7 +34,7 @@ def find_version(*file_paths):
     'transformers>=4.3.0',
     'pandas',
     'pylint',
-    'sklearn',
+    'scikit-learn',
     'ogb',
```
thvasilo, line 38: Should have a `>=` for the ogb version here as well since we had problems before?

classicsong: Can we fix it with ogb == 1.3.6?
Furthermore, we would better add torch >= 1.12

thvasilo: Yeah if it's going to be more than one change let's open another issue and PR to keep this one focused.

thvasilo: Opened #90 , if you can think of more _minimum_ requirements let's add them. I would suggest against pinning specific versions unless it's absolutely necessary as that would make installation for end users more difficult. "Larger or equal" tends to be good enough.

We can pin dependencies in our Dockerfiles to ensure working environment.

PR #98, opened 2023-04-17
Add hdf5 support

zheng-da:
*Description of changes:*
Add HDF5 support. We don't need to read the entire file into memory if there is only one file.


python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -66,7 +66,8 @@ def parse_node_data(in_file, feat_ops, node_id_col, label_conf, read_file):
         label_data = process_labels(data, label_conf, True)
         for key, val in label_data.items():
             feat_data[key] = val
-    return (data[node_id_col], feat_data)
+    node_ids = data[node_id_col] if node_id_col in data else None
```
classicsong, line 69: Why node_id_col does not exist in data?

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -163,23 +164,22 @@ def process_node_data(process_confs, convert2ext_mem, remap_id, num_processes=1)
...
+        assert 'format' in process_conf, \
+                "'format' must be defined for a node type"
+        # If there is only one file, we don't need to concatenate the data.
+        # Potentially, we don't need to read the data in memory if the input
+        # format supports it. Currently, only HDF5 supports this.
+        read_file = parse_node_file_format(process_conf, in_mem=len(in_files) > 1)
         feat_ops = parse_feat_ops(process_conf['features']) \
                 if 'features' in process_conf else None
         label_conf = process_conf['labels'] if 'labels' in process_conf else None
+        node_id_col = process_conf['node_id_col'] if 'node_id_col' in process_conf else None
```
classicsong, line 182: If 'node_id_col' not in process_conf, does it mean that the node id starts from 0?

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -233,12 +246,24 @@ def process_node_data(process_confs, convert2ext_mem, remap_id, num_processes=1)
...
 
-        # Some node types don't have data.
-        if len(type_node_data) > 0:
+        # If we didn't see the node data for this node type before.
+        if len(type_node_data) > 0 and node_type not in node_data:
             node_data[node_type] = type_node_data
+        # If we have seen the node data for this node type before.
+        elif len(type_node_data) > 0:
+            for key, val in type_node_data.items():
+                # Make sure the node data has duplicated names.
```
classicsong, line 255:
```suggestion
                # Make sure the node data does not have a naming conflict.
```

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -163,23 +164,22 @@ def process_node_data(process_confs, convert2ext_mem, remap_id, num_processes=1)
...
-        read_file = parse_node_file_format(process_conf)
         assert 'files' in process_conf, \
                 "'files' must be defined for a node type"
         in_files = get_in_files(process_conf['files'])
+        assert 'format' in process_conf, \
+                "'format' must be defined for a node type"
+        # If there is only one file, we don't need to concatenate the data.
+        # Potentially, we don't need to read the data in memory if the input
+        # format supports it. Currently, only HDF5 supports this.
+        read_file = parse_node_file_format(process_conf, in_mem=len(in_files) > 1)
```
classicsong, line 178: Is there anywhere in this function read_file is used?

python/graphstorm/gconstruct/construct_graph.py
zheng-da: This is to handle the case that multiple node data blobs are provided for one node type. We only need to specify the node ID column in one of the blobs.

python/graphstorm/gconstruct/construct_graph.py
zheng-da: yes, it is passed to the worker processes to read data from the input files.

classicsong (approved):
LGTM

PR #104, opened 2023-04-20
Support node data in separate files.

zheng-da:
*Description of changes:*
In some cases, the node data of the same node type are stored in separate files. We need to use multiple node blocks to specify the files. Currently, we only support this for node data.


python/graphstorm/gconstruct/file_io.py
```diff
@@ -140,6 +141,27 @@ def write_data_parquet(data, data_file):
...
+    Parameters
+    ----------
+    f : HDF5 file handle
+        The handle to access the HDF5 file.
+    fname : str
+        The path of the file.
+    """
+    def __init__(self, f, fname):
+        self._f = f
+        self._fname = fname
```
classicsong, line 159: It seems fname is not used.

tests/end2end-tests/data_process/data_gen.py
```diff
@@ -131,6 +136,17 @@ def split_data(data, num):
             },
         ],
     },
+    {
```
classicsong, line 139: What will happen if the block of node1 with `"node_id_col": "id"` comes after the one without `"node_id_col": "id"`? For example
```
node_conf = [
   {
       "node_type": "node1",
        "format": {"name": "hdf5"},
        "files": os.path.join(in_dir, "node_data1_2.hdf5"),
        ...
    },
    {
        "node_id_col": "id",
        "node_type": "node1",
        "format": {"name": "parquet"},
        "files": os.path.join(in_dir, "node_data1_*.parquet"),
        "features": [
        ...
    }
```


Will the graph construction crash?

zheng-da: i think it should be fine. let me modify the integration test to test this case.

classicsong (approved):
LGTM

PR #109, opened 2023-04-21
Reduce save model prints

classicsong:
*Issue #, if available:*
To many prints will cause CI error.

*Description of changes:*
Only rank 0 should print model saving prints.


zhjwy9343 (approved):
LGTM

Issue #111, opened 2023-04-21
[Config] need to rename "call_to_consider_early_stop" to an intuitive name


zhjwy9343:
The argument name "call_to_consider_early_stop" is not intuitive. After discussion, suggest to change it to "early_stop_burnin_rounds" and set default value to 0.

PR #112, opened 2023-04-21
[Fix] fix call_to_consider_early_stop argument

zhjwy9343:
*Issue #111 , if available:*

*Description of changes:*
Change all call_to_consider_early_stop and call-to-consider-early-stop to early_stop_burnin_rounds and early-stop-burnin-rounds.


training_scripts/gsgnn_np/arxiv_nc.yaml
```diff
@@ -16,8 +16,8 @@ gsf:
   input:
     restore_model_path: null
   output:
-    save_model_path: /data/ogbn-arxiv-1p/model/
-    save_embed_path: /data/ogbn-arxiv-1p/embed/
+    save_model_path: null
+    save_embed_path: null
```
thvasilo, line 20: Are these changes related to this PR? Or just a drive-by fix?

python/graphstorm/config/argument.py
```diff
@@ -759,16 +759,16 @@ def no_validation(self):
 
     ### control early stop ###
     @property
-    def call_to_consider_early_stop(self):
-        """ Burning period calls to start considering early stop
+    def early_stop_burnin_rounds(self):
+        """ Burnin rounds before to start early stop
```
thvasilo, line 763:
```suggestion
        """ Burn-in rounds before we start checking for the early stop condition.
```

thvasilo (approved):
Overall LGTM, might want to change the parameter description.

training_scripts/gsgnn_np/arxiv_nc.yaml
zhjwy9343: It is a drive-by fix.

python/graphstorm/config/argument.py
zhjwy9343: Good. I will change other doc strings similar to this line.

Issue #118, opened 2023-04-22
[Config] Remove argument `log_report_frequency`


isratnisa:
Remove argument `log_report_frequency`. It is not useful.

isratnisa:
SageMaker's `do_report` argument depends on `log_report_frequency`. So, GSF is keeping this parameter.

PR #119, opened 2023-04-22
[Config] Remove argument `log_report_frequency`

isratnisa:
Resolves Issue #118 

Removed argument `log_report_frequency`

### DIFF
--- a/python/graphstorm/tracker/sagemaker_tracker.py
+++ b/python/graphstorm/tracker/sagemaker_tracker.py
@@ -28,16 +28,6 @@ class GSSageMakerTaskTracker(GSTaskTrackerAbc):
             Task rank
     """
 
-    def _do_report(self, step):
-        """ Whether report the metric
-
-        Parameters
-        ----------
-        step: int
-            Current step
-        """
-        return step % self._report_frequency == 0
-
     def log_metric(self, metric_name, metric_value, step, force_report=False):
         """ log validation or test metric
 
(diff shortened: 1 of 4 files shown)


isratnisa:
@thvasilo I have removed the `_do_report` param from `sagemaker_tracker.py` file due to its dependency to `log_report_frequency`. Can you please check if that's alright?

python/graphstorm/tracker/sagemaker_tracker.py
```diff
@@ -52,7 +42,7 @@ def log_metric(self, metric_name, metric_value, step, force_report=False):
         force_report: bool
             If true, report the metric
         """
-        if force_report or self._do_report(step):
```
classicsong, line 55: do_report is useful when we are sending log to a system like mlflow or tensorboard.

classicsong: SageMaker Tracker also use this to decide whether to print the train step.

isratnisa: I see. We have to keep the argument for the `do_report` argument from SageMaker. For GSF, we can just make it an internal argument by disabling `group.add_argument("--log-report-frequency",...)`

zhjwy9343: What is our final decision for the "--log-report-frequency"?

classicsong: I would say, let's keep it for now.

isratnisa: OK. Closing this PR.

PR #123, opened 2023-04-23
Fix HDF5 array in graph construction

zheng-da:
*Issue #, if available:*

*Description of changes:*
HDF5 only support indexing with ascending order.


zhjwy9343 (approved):
There is a CI error. Other than that, LGTM

Issue #131, opened 2023-04-24
[Config] rename predict_*type to target_*type


zhjwy9343:
To be consistent, rename "predict_*type" to "target_*type"

PR #139, opened 2023-04-26
Rename predict_*type to target_*type

zhjwy9343:
*Issue #131 , if available:*

*Description of changes:*

Rename predict_*type to target_*type in both arguments and customer models.


examples/customized_models/HGT/acm_nc.yaml
```diff
@@ -3,15 +3,12 @@ version: 1.0
...
     model_encoder_type: rgcn
-    graph_name: acm
     backend: gloo
     ip_config: ip_list.txt
     part_config: /data/acm_nc/acm.json
     verbose: false
     mixed_precision: false
     no_validation: false
-    mp_opt_level: O2
-    train_nodes: 10
```
isratnisa, line 14: These two args are removed from the all the yaml files?

isratnisa (approved):
LGTM.

examples/customized_models/HGT/acm_nc.yaml
zhjwy9343: I think so. We don't have there two in our argument list.

PR #144, opened 2023-04-27
Compute BERT embeddings in the graph construction

zheng-da:
*Description of changes:*
Compute BERT embeddings.

### DIFF
--- a/python/graphstorm/gconstruct/transform.py
+++ b/python/graphstorm/gconstruct/transform.py
@@ -145,7 +145,10 @@ class ComputeBERT(FeatTransform):
         # We use the local GPU to compute BERT embeddings.
         if th.cuda.is_available():
             use_gpu = True
-            self.lm_model = self.lm_model.cuda()
+            gpu = int(os.environ['CUDA_VISIBLE_DEVICES']) \
+                    if 'CUDA_VISIBLE_DEVICES' in os.environ else 0
+            device = f"cuda:{gpu}"
+            self.lm_model = self.lm_model.to(device)
         else:
             use_gpu = False
         outputs = self.tokenizer(strs)
@@ -164,9 +167,9 @@ class ComputeBERT(FeatTransform):
             for tokens, att_masks, token_types in zip(tokens_list, att_masks_list,
                                                       token_types_list):
                 if use_gpu:
-                    outputs = self.lm_model(tokens.cuda(),
-                                            attention_mask=att_masks.cuda().long(),
-                                            token_type_ids=token_types.cuda().long())
+                    outputs = self.lm_model(tokens.to(device),
+                                            attention_mask=att_masks.to(device).long(),
+                                            token_type_ids=token_types.to(device).long())
                 else:
                     outputs = self.lm_model(tokens,
                                             attention_mask=att_masks.long(),
--- a/python/graphstorm/gconstruct/transform.py
+++ b/python/graphstorm/gconstruct/transform.py
@@ -62,14 +62,14 @@ class Tokenizer(FeatTransform):
         The name of the column that contains the text features
     feat_name : str
         The prefix of the tokenized data
-    tokenizer : HuggingFace Tokenizer
-        a tokenizer
+    bert_model : str
+        The name of the BERT model.
     max_seq_length : int
         The maximal length of the tokenization results.
     """
-    def __init__(self, col_name, feat_name, tokenizer, max_seq_length):
+    def __init__(self, col_name, feat_name, bert_model, max_seq_length):
         super(Tokenizer, self).__init__(col_name, feat_name)
-        self.tokenizer = tokenizer
+        self.tokenizer = BertTokenizer.from_pretrained(bert_model)
         self.max_seq_length = max_seq_length
 
     def __call__(self, strs):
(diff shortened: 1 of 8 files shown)


python/graphstorm/gconstruct/transform.py
```diff
@@ -101,6 +103,84 @@ def __call__(self, strs):
                 atten_mask_name: th.cat(att_masks, dim=0).numpy(),
                 token_type_id_name: th.cat(type_ids, dim=0).numpy()}
 
+class ComputeBERT(FeatTransform):
```
zhjwy9343, line 106: Because this class extends from FeatTransform, can we rename is to BERTTransform or similar. ComputeBERT is an action, not sounds like a noun.

zhjwy9343 (approved):
Besides the class name, others LGTM.

python/graphstorm/gconstruct/transform.py
```diff
@@ -101,6 +103,84 @@ def __call__(self, strs):
...
+            tokens_list = th.split(th.tensor(outputs['input_ids']), self.infer_batch_size)
+            att_masks_list = th.split(th.tensor(outputs['attention_mask']),
+                                      self.infer_batch_size)
+            token_types_list = th.split(th.tensor(outputs['token_type_ids']),
+                                        self.infer_batch_size)
+        else:
+            tokens_list = [th.tensor(outputs['input_ids'])]
+            att_masks_list = [th.tensor(outputs['attention_mask'])]
+            token_types_list = [th.tensor(outputs['token_type_ids'])]
+        with th.no_grad():
```
classicsong, line 166: We may also call model.eval().

python/graphstorm/gconstruct/utils.py
```diff
@@ -42,6 +44,14 @@ def worker_fn(task_queue, res_queue, user_parser):
     user_parser : callable
         The user-defined function to read and process the data files.
     """
+    # We need to set a GPU device for each worker process in case that
+    # some transformations (e.g., computing BERT embeddings) require GPU computation.
+    if th.cuda.is_available():
+        num_gpus = th.cuda.device_count()
+        gpu = worker_id % num_gpus
+        os.environ['CUDA_VISIBLE_DEVICES'] = str(gpu)
+        if worker_id >= num_gpus:
```
classicsong, line 53: In each worker, can we check whether worker_id > num_gpus? If so, the worker just return an empty result.

classicsong: This can help avoid potential OOM problem.

tests/unit-tests/test_construct_graph.py
```diff
@@ -188,6 +188,47 @@ def test_feat_ops():
     assert "attention_mask" in proc_res
     assert "token_type_ids" in proc_res
 
+    # Compute BERT embeddings.
+    feat_op3 = [
```
classicsong, line 192: Do you add test for GPU?

tests/unit-tests/test_construct_graph.py
```diff
@@ -188,6 +188,47 @@ def test_feat_ops():
...
+        },
+    ]
+    res3 = parse_feat_ops(feat_op3)
+    assert len(res3) == 1
+    assert res3[0].col_name == feat_op3[0]["feature_col"]
+    assert res3[0].feat_name == feat_op3[0]["feature_name"]
+    proc_res = process_features(data, res3)
+    assert "test4" in proc_res
+    assert len(proc_res['test4']) == 2
+    np.testing.assert_array_equal(proc_res['test4'][0], proc_res['test4'][1])
```
classicsong, line 209: What is proc_res['test4'][0] and proc_res['test4'][1]?

tests/unit-tests/test_construct_graph.py
```diff
@@ -188,6 +188,47 @@ def test_feat_ops():
...
+        },
+    ]
+    res4 = parse_feat_ops(feat_op4)
+    assert len(res4) == 1
+    assert res4[0].col_name == feat_op4[0]["feature_col"]
+    assert res4[0].feat_name == feat_op4[0]["feature_name"]
+    proc_res = process_features(data, res4)
+    assert "test4" in proc_res
+    assert len(proc_res['test4']) == 2
+    np.testing.assert_array_equal(proc_res['test4'][0], proc_res['test4'][1])
```
classicsong, line 230: Can you also compare the output of L209 with L230?

python/graphstorm/gconstruct/utils.py
zheng-da: I give a warning here. we should allow the number of processes to be larger than the number of GPUs. some of the feature operations run on CPU. currently, only bert computations run on GPUs.

zheng-da: this is not so good either. some feature transformations don't require GPUs. the message will be confusing.

tests/unit-tests/test_construct_graph.py
zheng-da: it's quite hard to test if the embeddings are computed by GPU.

python/graphstorm/gconstruct/utils.py
classicsong: OK.
Are we sharing the Bert model between workers using the same GPU or each worker needs to load the BERT model?

I am fine with current implementation. But we would better optimize it later.

tests/unit-tests/test_construct_graph.py
classicsong: We would better test whether the code works with multi-GPU and generate same (similar) embedding as CPU.
We can add a new unitest pipeline for multi-GPU unitests.

python/graphstorm/gconstruct/utils.py
zheng-da: the bert model is not shared. when computing bert embeddings, we should ask users to set the number of trainers correctly.

zheng-da: another option is to detect whether the feature transformation requires GPU. if it does, we set the number of trainer processes to the number of available GPUs automatically.

tests/unit-tests/test_construct_graph.py
zheng-da: i just realized that the code is tested on GPU. if the machine has GPU available, it uses GPU directly. Multi-GPU setting cannot only be tested in the integration test. The code requires to generate multiple input files to trigger multi-GPU tests. this is tested in the integration test.

classicsong: OK

python/graphstorm/gconstruct/utils.py
classicsong: I would prefer the later one.

classicsong (approved):
Overall LGTM.

tests/end2end-tests/data_process/movielens_test.sh
```diff
@@ -27,6 +27,17 @@ python3 -m graphstorm.gconstruct.construct_graph --conf_file $GS_HOME/tests/end2
...
+python3 -m graphstorm.gconstruct.construct_graph --conf_file $GS_HOME/tests/end2end-tests/data_process/movielens_bert.json --num_processes 1 --output_dir /tmp/movielens_bert --graph_name ml --add_reverse_edges
+
+error_and_exit $?
+
+echo "**************dataset: Test edge classification, RGCN layer: 1, node feat: BERT embeddings, inference: mini-batch"
+
+python3 -m graphstorm.run.gs_edge_classification --workspace $GS_HOME/training_scripts/gsgnn_ep --num_trainers $NUM_TRAINERS --num_servers 1 --num_samplers 0 --part_config /tmp/movielens_bert/ml.json --ip_config ip_list.txt --ssh_port 2222 --cf ml_ec.yaml --num-epochs 1 --node-feat-name user:feat movie:title
+
+error_and_exit $?
+
```
classicsong, line 40: We should rm /tmp/movielens_bert.

PR #147, opened 2023-04-27
[Config] Removing separate_eval

prateekdesai04:
*Issue #, if available:*

*Description of changes:*


thvasilo (approved):
LGTM. Wanted to note that I suggested to Prateek he remove the `separate_eval` property since we're not using it, instead of commenting/out leaving dead code. If in the future we decide to support this it should be easy to add the property back.

zhjwy9343 (approved):
LGTM

PR #148, opened 2023-04-29
Fix data split

zheng-da:
*Description of changes:*
If there is no training/val/test data, `node_split` or `edge_split` may return `None`. We need to handle this case.


zheng-da:
this should be fixed in DGL. https://github.com/dmlc/dgl/pull/5653

PR #151, opened 2023-04-29
Update Kill&Cleanup Script

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
Update the tool/kill_cleanup_disttrain.sh to include port number as an argument.

### DIFF
--- a/tools/kill_cleanup_disttrain.sh
+++ b/tools/kill_cleanup_disttrain.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 if [ "$#" -lt 2 ]; then
-  echo "Usage: $0 ip_list_file graph_name"
+  echo "Usage: $0 ip_list_file port_num graph_name"
   exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


tools/kill_cleanup_disttrain.sh
```diff
@@ -12,10 +12,16 @@ done < "$1"
 
 graph_name=$2
 
+if [ -z "$3"]; then
+    port_num=2222
```
classicsong, line 16: port_num=22?
This is the default port number for ssh. I believe most of people assume that by default ssh uses 22.

classicsong (approved):
One comment.

tools/kill_cleanup_disttrain.sh
zhjwy9343: This command should be executed within our Docker container. So the default should be 2222 as we designed.

If use 22, the command will fail and users have no clue what port number to use.

PR #154, opened 2023-05-02
Check the size for parquet input files

zheng-da:
*Description of changes:*
Parquet doesn't support large columns with more than 2B elements. Report errors if the input is too large.


classicsong (approved):
LGTM

Issue #162, opened 2023-05-05
[Bug] gconstruct cannot convert a 1D feature into 2D for GraphStorm


zhjwy9343:
Description:
If user creates the raw graph data, and have a node feature in 1D format, e.g., 1 col in Pandas dataframe as 1 feature, the gconstruct.py will not trigger any problem, and put the 1D tensor as a node feature.

Then during GraphStorm launch, gsf.py will check the shape of features, and failed with the following error track:

```
File "/graphstorm/python/graphstorm/gsf.py", line 110, in get_feat_size
    assert len(g.nodes[ntype].data[fname].shape) == 2, \
AssertionError: Input node features should be 2D tensors

```

PR #163, opened 2023-05-05
[Fix] Fix the gconstruction 1D feature bug

zhjwy9343:
*Issue #162, if available:*

*Description of changes:*
Add a shape check in the process_feature function and also add a test function to make sure it works.


python/graphstorm/gconstruct/transform.py
```diff
@@ -190,6 +190,13 @@ def process_features(data, ops):
         res = op(data[op.col_name])
         assert isinstance(res, dict)
         for key, val in res.items():
+            # Check if has 1D features. If yes, convert to 2D features
+            if len(val.shape) == 1:
+                if isinstance(val, HDF5Array):
+                    val = val.to_numpy().reshape(-1, 1)
```
zheng-da, line 196: it's better to print a message here saying that "We need to load the data from HDF5 to memory to convert its shape to XX".

The reason that we have HDF5Array is to avoid loading its data to memory during the graph construction to reduce memory consumption.

zheng-da (approved):

PR #164, opened 2023-05-06
Reduce memory consumption in the input layer.

zheng-da:
*Issue #, if available:*

*Description of changes:*
To reduce memory consumption, we avoid computing the input embeddings in advance and saving them in distributed memory.


python/graphstorm/model/gnn_encoder_base.py
```diff
@@ -140,7 +139,10 @@ def dist_inference(g, gnn_encoder, node_feats, batch_size, fanout,
                     assert len(g.ntypes) == 1
                     output_nodes = {g.ntypes[0]: output_nodes}
 
-                h = {k: x[k][input_nodes[k]].to(device) for k in input_nodes.keys()}
+                if i == 0:
+                    h = get_input_embeds(blocks[0], input_nodes)
```
classicsong, line 143: Will this cause duplicated input feature computation?
I believe so.

python/graphstorm/model/gnn_encoder_base.py
```diff
@@ -98,7 +98,6 @@ def dist_inference(g, gnn_encoder, node_feats, batch_size, fanout,
     dict of Tensor : the final GNN embeddings of all nodes.
     """
     device = gnn_encoder.device
-    x = node_feats
     with th.no_grad():
```
classicsong, line 101: The reason you want to compute input embedding within the for loop is that node_feats consumes many memory and we can not release it.

But if we do x = compute_node_input_embeddings(), x will automatically be released in L152.

python/graphstorm/model/gnn_encoder_base.py
zheng-da: i will. it trades off computation for memory. if the input layer is very simple, such as MLP, this works fine. if the input layer is BERT, we probably don't want to do it.

python/graphstorm/model/gnn_encoder_base.py
zheng-da: this approach requires to keep at least two embedding tensors of all nodes in memory.

classicsong: We always need two tensors (one for N-1 layer and one for N), right?

python/graphstorm/model/gnn_encoder_base.py
classicsong: Shall we implement two logics to cover both use cases?

zheng-da: probably not here. I'll make some modifications in the place where `dist_inference` is called.

python/graphstorm/model/embed.py
```diff
@@ -107,6 +107,19 @@ def get_sparse_params(self):
         # By default, there is no sparse_embeds
         return []
 
+    def require_cache_embed(self):
```
classicsong, line 110: Can you also change GSLMNodeEncoderInputLayer and GSPureLMNodeInputLayer with require_cache_embed() returning True.

zheng-da: they are changed.

classicsong (approved):
LGTM

classicsong (approved):
LGTM

Issue #49, opened 2023-03-30
Save embeddings based on the original node IDs


zheng-da:
When GNN embeddings are saved, they are saved based on the node IDs in the partitioned graph, which are different from the original node IDs of users' input data. We need to define a simple way for users to identify the right GNN embeddings from the original node IDs.

classicsong:
We also need to store prediction results according to node_id_mapping and edge_id_mapping.

PR #166, opened 2023-05-08
Shuffle node embeddings according to node_id_mapping when saving node embeddings

classicsong:
*Issue #, if available:*
#49 

*Description of changes:*
If node id mapping exists in the partition files, shuffle node embeddings according to the node_id_mapping when save node embeddings.


tests/unit-tests/test_utils.py
```diff
@@ -55,6 +66,134 @@ def helper_save_embedding(tmpdirname):
...
+# Will add test for nccl once we enable nccl
+@pytest.mark.parametrize("num_embs", [16, 17])
+@pytest.mark.parametrize("backend", ["gloo"])
+def test_save_embeddings_with_id_mapping(num_embs, backend):
+    import tempfile
+
+    # single embedding
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        emb, nid_mapping = gen_embedding_with_nid_mapping(num_embs)
+        print(nid_mapping)
```
zhjwy9343, line 124: Do we need to print the nid_mapping?

zhjwy9343 (approved):
LGTM, except for a few wording issues.

zhjwy9343 (approved):
LGTM, except for a few wording issues.

zhjwy9343:
The new argument is not added to argument.py and cause the failure of end-to-end test.

tests/unit-tests/test_utils.py
classicsong: NO.

python/graphstorm/model/utils.py
```diff
@@ -250,33 +289,103 @@ def save_embeddings(model_path, embeddings, local_rank, world_size):
...
     assert local_rank < world_size
-    def get_data_range(num_embs):
-        # Get corresponding data range
-        start = local_rank * (num_embs // world_size)
-        end = (local_rank + 1) * (num_embs // world_size)
-        end = num_embs if local_rank + 1 == world_size else end
-        return start, end
+    backend = th.distributed.get_backend()
+    device = th.device('cpu') if backend == "gloo" else device
+    def exchange_node_id_mapping(node_id_mapping, num_embs):
```
zheng-da, line 299: it's better to move the function out. it's not easy to read. it also makes the main function too long.

zheng-da (approved):

python/graphstorm/model/utils.py
classicsong: Updated

PR #186, opened 2023-05-22
[feature] Add local-joint negative sampler for link prediction

classicsong:
*Issue #, if available:*

*Description of changes:*
Add LocalJoint negative sampler. This can help reduce the possibility that a negative sample is remote, which helps reduce the cost of doing graph sampling remotely and pull node features from remote. Here is a profiling result:

Dataset: MAG-LSC with 38.6M training edges.
Environment: 4 * g5.48xlarge
GNN model setting: 2 layer RGCN with a fanout of 10,10; Training batch size: 1024
Training time:
                | epoch time
joint         |  17999.55s
localjoint |  7266.78s

The training speed is improved by 2.5X

Dataset: MAG ([example](https://github.com/awslabs/graphstorm/tree/main/examples/mag))
With localjoint, Link prediction on MRR can reach  0.2987 with 3 epochs.
Training time:
                |  End2end train time (3 epochs)
joint         |  2119 mins
localjoint |  1510 mins


python/graphstorm/dataloading/sampler.py
```diff
@@ -210,3 +210,43 @@ def _gen_neg_pair(pos_pair, utype, vtype):
             pos_neg_tuple = _gen_neg_pair(pos_pairs,
                 g.canonical_etypes[0][0], g.canonical_etypes[0][2])
         return pos_neg_tuple
+
+class JointLocalUniform(JointUniform):
+    '''Jointly corrupt a group of edges.
+    The main idea is to sample a set of nodes and use them to corrupt all edges in a mini-batch.
+    This algorithm won't change the sampling probability for each individual edge, but can
+    significantly reduce the number of nodes in a mini-batch.
+
```
isratnisa, line 219: It's hard to understand from the docstring how it is different than JointUniform or other sampler. Can you please clarify a little more?

isratnisa (changes requested):

isratnisa:
Can you please add model accuracy too in the description?

python/graphstorm/dataloading/sampler.py
```diff
@@ -210,3 +210,47 @@ def _gen_neg_pair(pos_pair, utype, vtype):
             pos_neg_tuple = _gen_neg_pair(pos_pairs,
                 g.canonical_etypes[0][0], g.canonical_etypes[0][2])
         return pos_neg_tuple
+
+class JointLocalUniform(JointUniform):
+    '''Jointly corrupt a group of edges.
+
+    The main idea is to sample a set of nodes and use them to corrupt all edges in a mini-batch.
+    This algorithm won't change the sampling probability for each individual edge, but can
```
zheng-da, line 218: i don't understand this statement. it changes the sampling probability for the negative nodes. why does it not change the sampling probability for each edge?

python/graphstorm/dataloading/sampler.py
```diff
@@ -210,3 +210,47 @@ def _gen_neg_pair(pos_pair, utype, vtype):
...
+    significantly reduce the number of nodes in a mini-batch.
+    The difference between JointUniform and JointLocalUniform is that JointUniform sample
+    negative nodes from the entire graph, but JointLocalUniform only sample negative nodes
+    from the local partition.
+
+    Parameters
+    ----------
+    k : int
+        The number of negative examples per edge.
```
zheng-da, line 227: add `per_trainer`

python/graphstorm/dataloading/sampler.py
```diff
@@ -210,3 +210,47 @@ def _gen_neg_pair(pos_pair, utype, vtype):
...
+
+    def _generate(self, g, eids, canonical_etype):
+        _, _, vtype = canonical_etype
+        shape = eids.shape
+        dtype = eids.dtype
+        ctx = eids.device
+        src, _ = g.find_edges(eids, etype=canonical_etype)
+        src = np.repeat(src.numpy(), self.k)
+
+        if vtype not in self._local_neg_nids:
```
zheng-da, line 242: how much memory consumption is this?

python/graphstorm/dataloading/sampler.py
```diff
@@ -210,3 +210,47 @@ def _gen_neg_pair(pos_pair, utype, vtype):
...
+            pb = g.get_partition_book()
+            if self._per_trainer:
+                neg_idx = node_split(th.full((g.num_nodes(vtype),), True, dtype=th.bool),
+                                     pb, ntype=vtype, force_even=False,
+                                     node_trainer_ids=g.nodes[vtype].data['trainer_id'])
+            else:
+                neg_idx = node_split(th.full((g.num_nodes(vtype),), True, dtype=th.bool),
```
zheng-da, line 249: do we need to store them?

python/graphstorm/dataloading/sampler.py
```diff
@@ -210,3 +210,47 @@ def _gen_neg_pair(pos_pair, utype, vtype):
...
+        ctx = eids.device
+        src, _ = g.find_edges(eids, etype=canonical_etype)
+        src = np.repeat(src.numpy(), self.k)
+
+        if vtype not in self._local_neg_nids:
+            pb = g.get_partition_book()
+            if self._per_trainer:
+                neg_idx = node_split(th.full((g.num_nodes(vtype),), True, dtype=th.bool),
+                                     pb, ntype=vtype, force_even=False,
+                                     node_trainer_ids=g.nodes[vtype].data['trainer_id'])
```
zheng-da, line 247: it samples negative nodes from the partition owned by the local process. it doesn't sample negative nodes from the partition owned by the local machine.

python/graphstorm/dataloading/sampler.py
classicsong: It does not change the probability of positive edges. Let me update the docstr.

python/graphstorm/dataloading/sampler.py
classicsong: You are right. I copy paste the code from local uniform

python/graphstorm/dataloading/sampler.py
classicsong: Yes. Otherwise we need to compute it every time we do negative sampling.

python/graphstorm/dataloading/sampler.py
zheng-da: O(N) over all processes?

python/graphstorm/dataloading/sampler.py
zheng-da: oh, both branches prepare negative sampling for the local partition. one is to sample from the local machine, the other is to sample from the local trainer. can we simplify this? do we really need to sample from the local trainer? sampling from the local machine should be enough.

python/graphstorm/dataloading/sampler.py
classicsong: Yes. You only need to store neg_idx.

python/graphstorm/dataloading/sampler.py
```diff
@@ -210,3 +216,52 @@ def _gen_neg_pair(pos_pair, utype, vtype):
...
+        shape = eids.shape
+        dtype = eids.dtype
+        ctx = eids.device
+        src, _ = g.find_edges(eids, etype=canonical_etype)
+        src = np.repeat(src.numpy(), self.k)
+
+        if vtype not in self._local_neg_nids:
+            pb = g.get_partition_book()
+            if self._per_trainer:
+                neg_idx = node_split(th.full((g.num_nodes(vtype),), True, dtype=th.bool),
```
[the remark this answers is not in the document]
classicsong: BTW, th.full((g.num_nodes(vtype),) will be released after node_split.

zheng-da (approved):

python/graphstorm/dataloading/sampler.py
classicsong: Done

isratnisa (approved):

PR #187, opened 2023-05-22
[Fix] revert --ssh-port to SSH_PORT

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
fix the error of --ssh-port in Dockerfile.local.


classicsong (approved):

prateekdesai04 (commented):
There is already an `apt-get update` statement in line 5.
One way to resolve this would be adding `docker build --no-cache -f $GSF_HOME"docker/Dockerfile.local" . -t $DOCKER_FULLNAME` in the **build_docker_oss4local.sh** file, this will build the image without caching and will solve the problem and we would not need the update commands in line 41 and 42.

zhjwy9343:
> There is already an `apt-get update` statement in line 5. One way to resolve this would be adding `docker build --no-cache -f $GSF_HOME"docker/Dockerfile.local" . -t $DOCKER_FULLNAME` in the **build_docker_oss4local.sh** file, this will build the image without caching and will solve the problem and we would not need the update commands in line 41 and 42.

Great! changed.

prateekdesai04 (approved):

PR #193, opened 2023-05-23
[Fix] Update setup adding fixed dependency versions

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
To have pip package to use required dependency, need to fix the versions of these dependencies.


classicsong (approved):
Can we add PyTorch>=1.12.0

classicsong (commented):
Can we use >xxx for most of cases?

Issue #63, opened 2023-04-04
[Feature Request] Weighted edge loss.


classicsong:
When doing link prediction training, different edges may have different weights representing their importance. Supporting weighted edge in link prediction loss is required.

We are going to add a new argument --lp-edge-weight-for-loss to specify the edge weight used in loss function.

Related #214

PR #201, opened 2023-05-25
Add edge weight in link prediction loss

classicsong:
*Issue #, if available:*
#63 

*Description of changes:*
Add LinkPredictWeightedDistMultDecoder and LinkPredictWeightedDotDecoder for weighted lp loss.
A weight is applied to positive losses generated by positive edges to re-scale the positive losses.
The negative loss will use the weight of 1. (No re-scaling)


python/graphstorm/config/argument.py
```diff
@@ -1176,6 +1176,50 @@ def lp_decoder_type(self):
...
 
+    @property
+    def lp_edge_weight_for_loss(self):
+        """ The edge data fields that stores the edge weights used
+            in computing link prediction loss
+
+            The edge_weight can be in following format:
+            1) [weight_name]: global weight name, if an edge has weight,
+            the corresponding weight name is <weight_name>
+            2) ["src0,rel0,dst0:weight0","src0,rel0,dst0:weight1",...]:
```
zheng-da, line 1187: maybe we should use two names. one is lp_edge_weight_per_edge and the other is lp_edge_weight_per_etype.

python/graphstorm/config/argument.py
```diff
@@ -1176,6 +1176,50 @@ def lp_decoder_type(self):
...
+                # global feat_name
+                return edge_weights[0]
+
+            # per edge type feature
+            weight_dict = {}
+            for weight_name in edge_weights:
+                weight_info = weight_name.split(":")
+                etype = tuple(weight_info[0].split(","))
+                if etype in weight_dict:
+                    assert False, \
```
zheng-da, line 1208: why don't you `assert etype in weight_dict`?

python/graphstorm/config/argument.py
```diff
@@ -1176,6 +1176,50 @@ def lp_decoder_type(self):
...
+                        f"You already specify the weight names of {etype}" \
+                        f"as {weight_dict[etype]}"
+
+                # TODO: if train_etype is None, we need to check if
+                # etype exists in g.
+                assert self.train_etype is None or etype in self.train_etype, \
+                    f"{etype} must in the training edge type list"
+                assert isinstance(weight_info[1], str), \
+                    f"Feature name of {etype} should be a string not {weight_info[1]}"
```
zheng-da, line 1217: `not` => instead of

python/graphstorm/config/argument.py
```diff
@@ -1176,6 +1176,50 @@ def lp_decoder_type(self):
...
+                weight_dict[etype] = [weight_info[1]]
```
zheng-da, line 1218: why does the value have to be a list?

python/graphstorm/dataloading/dataset.py
```diff
@@ -68,6 +68,40 @@ def prepare_batch_input(g, input_nodes,
...
+    """
+    feat = {}
+    for etypes, eid in input_edges.items():
+        feat_name = None if feat_field is None else \
+            [feat_field] if isinstance(feat_field, str) \
+            else feat_field[etypes] if etypes in feat_field else None
+
+        if feat_name is not None:
+            # concatenate multiple features together
+            feat[etypes] = th.cat([g.edges[etypes].data[fname][eid].to(dev) \
```
zheng-da, line 101: we may consider to not use concatenation here.

python/graphstorm/gsf.py
```diff
@@ -319,21 +324,38 @@ def create_builtin_lp_model(g, config, train_task):
...
             print("Using inner product objective for supervision")
-        decoder = LinkPredictDotDecoder(model.gnn_encoder.out_dims \
-                                            if model.gnn_encoder is not None \
-                                            else model.node_input_encoder.out_dims)
+        if config.lp_edge_weight_for_loss is None:
+            decoder = LinkPredictDotDecoder(model.gnn_encoder.out_dims \
+                                                if model.gnn_encoder is not None \
+                                                else model.node_input_encoder.out_dims)
+        else:
+            decoder = LinkPredictWeightedDotDecoder(model.gnn_encoder.out_dims \
```
zheng-da, line 332: do we need two classes `LinkPredictDotDecoder` and `LinkPredictWeightedDotDecoder`?

python/graphstorm/model/edge_decoder.py
```diff
@@ -639,3 +639,146 @@ def out_dims(self):
...
+                rel_embedding = rel_embedding.unsqueeze(dim=1)
+                src_type, _, dest_type = canonical_etype
+                u, v = g.edges(etype=canonical_etype)
+                src_emb = h[src_type][u]
+
+                dest_emb = h[dest_type][v]
+                rel_embedding = rel_embedding.repeat(1,dest_emb.shape[0]).T
+                scores_etype = calc_distmult_pos_score(src_emb, dest_emb, rel_embedding)
+
+                if self.training:
```
zheng-da, line 730: why is this needed here, but you don't need it in `LinkPredictWeightedDotDecoder`?

python/graphstorm/model/lp_gnn.py
```diff
@@ -78,6 +113,26 @@ def __init__(self, alpha_l2norm):
         super(GSgnnLinkPredictionModel, self).__init__()
         self.alpha_l2norm = alpha_l2norm
 
+    def prepare_pos_graph(self, pos_graph, data, device):
+        if isinstance(self.decoder, (LinkPredictWeightedDotDecoder,
```
zheng-da, line 117: is something under the if condition?

python/graphstorm/model/lp_gnn.py
```diff
@@ -78,6 +113,26 @@ def __init__(self, alpha_l2norm):
         super(GSgnnLinkPredictionModel, self).__init__()
         self.alpha_l2norm = alpha_l2norm
 
+    def prepare_pos_graph(self, pos_graph, data, device):
```
zheng-da, line 116: docstring.

python/graphstorm/trainer/lp_trainer.py
```diff
@@ -131,9 +131,13 @@ def fit(self, train_loader, num_epochs,
                 pos_graph = pos_graph.to(device)
                 neg_graph = neg_graph.to(device)
                 blocks = [blk.to(device) for blk in blocks]
-                input_feats = data.get_node_feats(input_nodes, device)
                 for _, nodes in input_nodes.items():
                     num_input_nodes += nodes.shape[0]
+                input_feats = data.get_node_feats(input_nodes, device)
+
+                # Preparing edge feature for pos_graph and neg_graph if any.
+                model.module.prepare_pos_graph(pos_graph, data, device)
```
zheng-da, line 139: it's a bit weird to have a model to do this. can we get the edge features in the trainer and pass them to the model?

python/graphstorm/config/argument.py
classicsong: For lp_edge_weight_per_etype do you mean, user only specify the weight of an edge type? Edges with the same edge type use the same weight?
We can support this later. I did not hear similar requirements before.

python/graphstorm/config/argument.py
classicsong: For extracting edge features from graph, I implemented a general function `prepare_batch_edge_input` which can be used by edge_weight_for_loss and edge_feature_for_decoder.
To follow `prepare_batch_edge_input`'s implementation, we return a list here.

python/graphstorm/dataloading/dataset.py
classicsong: I want to make it a general function which can handle single feature per edge or multiple features per edge.

python/graphstorm/gsf.py
classicsong: Yes, we do.
I don't want to make LinkPredictDotDecoder to complex to use. Basically LinkPredictWeightedDotDecoder inherits LinkPredictDotDecoder and rewrite the `forward` function.

python/graphstorm/model/edge_decoder.py
classicsong: We don't actually. Let me remove this.

python/graphstorm/trainer/lp_trainer.py
classicsong: In prepare_pos_graph, we are copying edge weights into pos_graph to compute edge loss. This is different from passing edge features for message passing. Pos_graph and neg_graph are not used in GNN message passing. 
My understanding of "get the edge features in the trainer and pass them to the model" is that it is how we pass edge features for message-passing.

python/graphstorm/config/argument.py
zheng-da: you are dealing with two use cases. I'm thinking if we should treat them separately.

python/graphstorm/config/argument.py
zheng-da: i think what is inconsistent is that line 1221 returns a string, but here it returns a list of one string for each edge type.

python/graphstorm/dataloading/dataset.py
zheng-da: my statement above is a general statement. there are cases we want to access each node data separately. but i don't think you need to address the problem in this PR.

python/graphstorm/trainer/lp_trainer.py
zheng-da: not necessarily. Trainers are responsible to prepare mini-batch data for model computation. Here it asks models to prepare the mini-batch data.

python/graphstorm/config/argument.py
classicsong: We only handle the first use case, i.e., lp_edge_weight_per_edge, in this PR. input option 2 just allows users to use different features (with different feature names) to store edge weights.

python/graphstorm/trainer/lp_trainer.py
zheng-da: the other question is how it is different from preparing edge features for message passing and for decoders? we should implement them in the same way.

python/graphstorm/config/argument.py
classicsong: It follows the same return type as L426 node_feat_name.
Returning a single string allows users to use one string to specify edge weights, whiling returning a dict of list of one string allows users to use different name to specify edge weights for different edge types.

Do you think we can only keep the `single string` option and force our user to use the same feature name for edge weights?

python/graphstorm/dataloading/dataset.py
classicsong: Agree.
That is a more complex use case.

python/graphstorm/model/lp_gnn.py
classicsong: Docstring is in L61 (GSgnnLinkPredictionModelInterface)

python/graphstorm/config/argument.py
classicsong: No. One edge weight tensor for an edge type. 
But in order to make `prepare_batch_edge_input` more generate, We use dict of list of strings instead of dict of strings here.

zheng-da (approved):

zheng-da (approved):

PR #207, opened 2023-05-27
Update utils.py to prevent json dump no file or directory error

Oceanusity:
Create the save_perf_results_path if it is not existed.

*Issue #, if available:*

*Description of changes:*
Create the save_perf_results_path if it is not existed to prevent json dump no file or directory error.


classicsong (approved):
LGTM

Issue #204, opened 2023-05-26
Clean up printed messages in graph construction pipeline and print out the final summary


zheng-da:
We should avoid printing out unnecessary messages but need to print out a final summary to inform users.

PR #209, opened 2023-05-29
Use logging to print messages.

zheng-da:
*Description of changes:*
We should use `logging` to print messages so that we can set the logging level to hide some messages.

Fix the issue: https://github.com/awslabs/graphstorm/issues/204

### DIFF
--- a/python/graphstorm/gconstruct/construct_graph.py
+++ b/python/graphstorm/gconstruct/construct_graph.py
@@ -413,32 +413,32 @@ def get_log_level(log_level):
         raise ValueError(f"Unknown logging level {log_level}. " + \
                 "The possible values are: debug, info, warning, error.")
 
-def print_graph_info(g):
+def print_graph_info(g, node_data, edge_data):
     logging.info("The graph has %d node types and %d edge types.",
                  len(g.ntypes), len(g.etypes))
     for ntype in g.ntypes:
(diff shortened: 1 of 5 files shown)


python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -430,7 +449,12 @@ def process_graph(args):
         edges = edges1
         sys_tracker.check('Add reverse edges')
     g = dgl.heterograph(edges, num_nodes_dict=num_nodes)
-    print(g)
+    logging.info("The graph has %d node types and %d edge types.",
+                 len(g.ntypes), len(g.etypes))
+    for ntype in g.ntypes:
+        logging.info("Node type %s has %d nodes.", ntype, g.number_of_nodes(ntype))
+    for etype in g.canonical_etypes:
+        logging.info("Edge type %s has %d edges.", str(etype), g.number_of_edges(etype))
```
classicsong, line 457: What about train/val/test info?

classicsong (approved):
One comment.
Overall LGTM

python/graphstorm/gconstruct/construct_graph.py
zheng-da: print out train/val/test split?

zheng-da: then we should print out features as well.

Issue #153, opened 2023-05-01
[Feature request] Support edge features


zheng-da:
Currently, GraphStorm doesn't pass the edge features.

PR #211, opened 2023-05-29
Edge feat decoder

classicsong:
*Issue #, if available:*
#153 

*Description of changes:*


python/graphstorm/gsf.py
```diff
@@ -233,6 +235,28 @@ def create_builtin_edge_model(g, config, train_task):
                                      num_classes,
                                      multilabel=config.multilabel,
                                      target_etype=target_etype)
+        elif decoder_type == "MLPEFeatEdgeDecoder":
```
zheng-da, line 238: why is it called `MLPEFeatEdgeDecoder`? MLP edge feature edge decoder?

python/graphstorm/model/edge_decoder.py
```diff
@@ -339,6 +354,106 @@ def out_dims(self):
...
+            nn_h = self.dropout(nn_h)
+            # [edge_feat] @ W -> h_dim
+            feat_h = th.matmul(efeat, self.feat_decoder)
+            feat_h = self.relu(feat_h)
+            feat_h = self.dropout(feat_h)
+            # [nn_h | feat_h] @ W -> h_dim
+            combine_h = th.cat([nn_h, feat_h], dim=1)
+            combine_h = th.matmul(combine_h, self.combine_decoder)
+            combine_h = self.relu(combine_h)
+            out = th.matmul(combine_h, self.decoder)
```
zheng-da, line 454: why do you not concatenate src node, dst node and edge features and then run MLP?

classicsong: Embeddings of src and dst are from NN network. However edge features are raw features, it would be better to project it into the same embeddings space of src and dst.

python/graphstorm/gsf.py
classicsong: MLP with edge feature edge decoder.

zheng-da (approved):

zheng-da (approved):

PR #212, opened 2023-05-29
[Preprocessing-Pipeline] Initial code drop for MetadataSchema class.

kylasa:
Initial code drop for the MetadataSchema class and its associated dependencies and unit test cases. 

*Issue #, if available:*

*Description of changes:*
MetadataSchema class is built as a wrapper around input datasets, metadata.json file. This class captures all the information in the input metadata.json file and builds appropriate data structures which are later used by the application. 


python/graphstorm/graphpartitioning/array_readwriter/__init__.py
```diff
@@ -0,0 +1,2 @@
+from . import csv, numpy_array, parquet
```
zheng-da, line 1: you need to add license in each file.

python/graphstorm/graphpartitioning/array_readwriter/csv.py
```diff
@@ -0,0 +1,61 @@
+import logging
+
+import pandas as pd
+import pyarrow
+import pyarrow.csv
+
+from .registry import register_array_parser
+
+@register_array_parser("csv")
```
zheng-da, line 9: what is this for?

python/graphstorm/graphpartitioning/array_readwriter/numpy_array.py
```diff
@@ -0,0 +1,43 @@
...
+    def read(self, path):
+        """ Read operation on a file.
+
+        Argument
+        --------
+        path: str
+            Path of a file from which to read.
+        """
+        logging.debug("Reading from %s using numpy format" % path)
+        arr = np.load(path, mmap_mode="r")
```
zheng-da, line 24: what is the numpy file format?

zheng-da: does it have to be written by `open_memmap`?

python/graphstorm/graphpartitioning/array_readwriter/numpy_array.py
```diff
@@ -0,0 +1,43 @@
...
+        --------
+        path: str
+            Path of a file to write data.
+        arr: numpy ndarray
+            Numpy ndarray to write to a file.
+        """
+        logging.debug("Writing to %s using numpy format" % path)
+        # np.save would load the entire memmap array up into CPU.  So we manually open
+        # an empty npy file with memmap mode and manually flush it instead.
+        new_arr = open_memmap(path, mode="w+", dtype=arr.dtype, shape=arr.shape)
```
zheng-da, line 41: do we need to close it explicitly?

python/graphstorm/graphpartitioning/array_readwriter/parquet.py
```diff
@@ -0,0 +1,82 @@
...
+        """ Read operation on a file.
+
+        Argument
+        --------
+        path: str
+            Path of a file from which to read.
+        """
+        logging.debug("Reading from %s using parquet format" % path)
+        metadata = pyarrow.parquet.read_metadata(path)
+        metadata = metadata.schema.to_arrow_schema().metadata
```
zheng-da, line 27: why do we need `metadata` while csv doesn't need it?

python/graphstorm/graphpartitioning/array_readwriter/parquet.py
```diff
@@ -0,0 +1,82 @@
...
+        # If not, it should be explictly specified in the file as metadata.
+        if metadata:
+            shape = metadata.get(b"shape", None)
+        else:
+            shape = None
+        table = pyarrow.parquet.read_table(path, memory_map=True)
+
+        data_types = table.schema.types
+        # Spark ML feature processing produces single-column parquet files where each row is a vector object
+        if len(data_types) == 1 and isinstance(data_types[0], pyarrow.ListType):
```
zheng-da, line 39: this is to store multi-dimension arrays?

python/graphstorm/graphpartitioning/array_readwriter/parquet.py
```diff
@@ -0,0 +1,82 @@
...
+            arr = np.array(table.to_pandas().iloc[:, 0].to_list())
+            logging.debug(
+                f"Parquet data under {path} converted from single vector per row to ndarray"
+            )
+        else:
```
zheng-da, line 44: this is to store single-column vector?

python/graphstorm/graphpartitioning/array_readwriter/parquet.py
```diff
@@ -0,0 +1,82 @@
...
+import numpy as np
+import pandas as pd
+import pyarrow
+import pyarrow.parquet
+
+from .registry import register_array_parser
+
+
+@register_array_parser("parquet")
+class ParquetArrayParser(object):
```
zheng-da, line 12: it seems that you are defining your own format to read data. Who prepared the data?

python/graphstorm/graphpartitioning/array_readwriter/parquet.py
```diff
@@ -0,0 +1,82 @@
...
+        shape = array.shape
+        if len(shape) > 2:
+            array = array.reshape(shape[0], -1)
+        if vector_rows:
+            table = pyarrow.table(
+                [pyarrow.array(array.tolist())], names=["vector"]
+            )
+            logging.debug("Writing to %s using single-vector rows..." % path)
+        else:
+            table = pyarrow.Table.from_pandas(pd.DataFrame(array))
```
zheng-da, line 78: how do you process multi-dimension arrays in this case?

python/graphstorm/graphpartitioning/array_readwriter/__init__.py
kylasa: Completed as suggested.

python/graphstorm/graphpartitioning/array_readwriter/csv.py
kylasa: This is a decorator for this class to register itself for csv files.

python/graphstorm/graphpartitioning/array_readwriter/parquet.py
kylasa: This is just a class decorator, for registering this class for read/write operations on parquet files. 

This class is borrowed from the existing pipeline's implementation and the original author of these set of classes is Minjie. That said, these set of class expect data to be read and returned to the invoker in numpy format. Similarly the numpy data is taken and written in the appropriate files, as defined by the user.

python/graphstorm/graphpartitioning/array_readwriter/numpy_array.py
kylasa: Not necessarily. But the expected use of all the ``write`` calls in this directory are meant to be for very small files. In the current implementation, these calls are ONLY used in the end-to-end tests. That is the only use case for these write function calls. 

If there is a need to support these function calls for general purpose usage, then this needs to be investigated further to explore the balance between speed and resource(memory) usage.

python/graphstorm/graphpartitioning/array_readwriter/numpy_array.py
kylasa: Not necessarily. We can do so manually, but upon exiting of the function the resources should be automatically released by the python environment. Again, as indicated earlier, the intended usage of the write function calls is only meant for end-to-end testing framework.

python/graphstorm/graphpartitioning/array_readwriter/parquet.py
kylasa: Here the metadata has a different meaning compared to csv files. The author expects that the dimensions/shape information can be extracted by reading the parquet file header.

python/graphstorm/graphpartitioning/array_readwriter/parquet.py
kylasa: The else part of this statement is meant for multi-dimensional arrays.

python/graphstorm/graphpartitioning/array_readwriter/parquet.py
kylasa: In the case of single-dimension, or vectors, array is converted to a list and converted to pyarrow.array. And for multi-dimensional arrays first array is read into a DataFrame(pandas) and converted to pyarrow.Table. .

kylasa:
As a general comment for this Pull Request. 
1. array_readwriter is borrowed, as it is, from the existing implementation of the current pipeline. 
2. First lintrunner is used to format the code using the styling guidelines enforced in dgl repository and on top of the pylint changes are applied as required by the graphstorm repo. 
3. the final code uses the style as required by the lintrunner/pylint tools. I notice some inconsistency in some lines of code which is the result of these tools. Such lines have not been manually modified.

.github/workflow_scripts/lint_check.sh
```diff
@@ -17,4 +17,7 @@ pylint --rcfile=./tests/lint/pylintrc ./python/graphstorm/tracker/
 pylint --rcfile=./tests/lint/pylintrc ./python/graphstorm/run/
 pylint --rcfile=./tests/lint/pylintrc ./python/graphstorm/utils.py
 
-pylint --rcfile=./tests/lint/pylintrc ./python/graphstorm/sagemaker/
\ No newline at end of file
+pylint --rcfile=./tests/lint/pylintrc ./python/graphstorm/sagemaker/
+
+pylint --rcfile=./tests/lint/pylintrc ./python/graphstorm/graphpartitioning/
+pylint --rcfile=./tests/lint/pylintrc ./python/graphstorm/graphpartitioning/array_readwriter/
```
zheng-da, line 23: do you need to add both?

kylasa: This is indeed not needed. line 23 is now deleted.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,799 @@
...
+    >>> # id_ntype_map will be {0: ``ntype0``, 1: ``ntype1``, ...}
+    >>> id_ntype_map = Metadata().id_ntype_map
+    """
+
+    _instance = None
+
+    def __new__(cls):
+        if cls._instance is None:
+            cls._instance = super(MetadataSchema, cls).__new__(cls)
+        return cls._instance
```
thvasilo, line 147: As discussed previously: I don't understand the need for a singleton design for this class.

What's wrong with having `my_metadata = MetadataSchema(config_file_path)`? A singleton object has _very_ specific use cases, this seems like over-engineering.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,799 @@
...
+        self._id_etype_map = None
+        self._etypes = None
+        self._ntype_features = None
+        self._etype_features = None
+        self._ntype_feature_files = None
+        self._etype_feature_files = None
+        self._etype_files = None
+
+    def _load_json(self, file_name, base_dir):
+        """Helpper function to read the metadata schema file of the input
```
thvasilo, line 171:
```suggestion
        """Helper function to read the metadata schema file of the input
```

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,799 @@
...
+        """
+        if file_name is None or not isinstance(file_name, str):
+            raise ValueError(f"File path is not valid. = {file_name}")
+
+        self._basedir = None
+        if base_dir is not None and os.path.isdir(base_dir):
+            if not os.path.isfile(os.path.join(base_dir, file_name)):
+                raise ValueError(
+                    f"Metadata file does not exist. {base_dir}, {file_name}"
+                )
```
thvasilo, line 195: Why raise a `ValueError` here instead of letting Python raise a more appropriate `FileNotFoundError`?

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,799 @@
...
+            self._metadata_path = os.path.join(
+                os.path.join(base_dir, file_name)
+            )
+            self._basedir = base_dir
+        else:
+            if not os.path.isfile(file_name):
+                raise ValueError(f"Metadata file does not exist. {file_name}")
```
thvasilo, line 202: Same here. Let Python handle error reporting.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,799 @@
...
+        --------
+        data_files: list of str
+            A set of file names, either in absolute or in relative formats.
+        src_type: str
+            Either node type or edge type from the input graph.
+        """
+        if not isinstance(data_files, list):
+            raise ValueError(
+                f"Incorrect value encountered when iterating over files for {src_type}."
+            )
```
thvasilo, line 224: Again, if `data_files` is not a list, Python will raise the appropriate error.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,799 @@
...
+                raise ValueError(f"Metadata file does not exist. {file_name}")
+            self._metadata_path = file_name
+
+        if self._basedir is None:
+            self._basedir = os.getcwd()
+
+        with open(self._metadata_path, "r", encoding="utf-8") as handle:
+            self._data = json.load(handle)
+
+    def _check_ifexists(self, data_files, src_type):
```
thvasilo, line 211: In general it's hard to see the value of this function, just let Python error out if a file doesn't exist.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,799 @@
...
+            if not os.path.isabs(filename):
+                if not os.path.isfile(os.path.join(self._basedir, filename)):
+                    raise FileNotFoundError(
+                        f"File: {os.path.join(self._basedir, filename)} does not exist"
+                    )
+            else:
+                if not os.path.isfile(filename):
+                    raise FileNotFoundError(f"File: {filename} does not exist.")
+
+    def init(self, file_path, base_dir, init_maps=True):
```
thvasilo, line 236: These should be the constructor of the class.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,799 @@
...
+            Path of the metadata schema file of the input dataset. It can be
+            either in absolute or relative formats.
+        base_dir: str
+            Absolute path of the directory in which the metadata schema file
+            is expected to be present. If not provided then current workding
+            directory is used instead to locate the metadata schema file.
+        init_maps: bool
+            Boolean flag to indicate whether to build all the necessary data
+            structures during initialization. Otherwise the user of this class
+            is exptected to invoke the appropriate initialization function.
```
thvasilo, line 252: In general what's the value of having this option? Isn't it expected that we should init all the maps every time?

```suggestion
        init_maps: bool
            Boolean flag to indicate whether to build all the necessary data
            structures during initialization. Otherwise the user of this class
            is expected to invoke the appropriate initialization function.
```

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,799 @@
...
+        """
+        self.init_global_nid_offsets()
+        self.init_global_eid_offsets()
+        self.init_ntype_maps()
+        self.init_etype_maps()
+        self.init_ntype_features()
+        self.init_etype_features()
+        self.init_etype_files()
+        self.init_ntype_feature_files()
+        self.init_etype_feature_files()
```
thvasilo, line 271: Why are all these functions public if the caller is not supposed to have access to them? The fact that they don't take arguments is an indication that they should just be called in the constructor.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,799 @@
...
+        self._ntype_id_map = None
+        self._id_ntype_map = None
+        self._etype_id_map = None
+        self._id_etype_map = None
+        self._etypes = None
+        self._ntype_features = None
+        self._etype_features = None
+        self._ntype_feature_files = None
+        self._etype_feature_files = None
+        self._etype_files = None
```
thvasilo, line 611: I'd be surprised if having a destructor adds value here. All these member vars will get released and gc'ed when the object goes out of scope right?

thvasilo (changes requested):
First pass review of the code, we can address comments and do a second iteration.

python/graphstorm/graphpartitioning/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+from .constants import STR_NUM_NODES_PER_TYPE
+from .constants import STR_EDGE_TYPE
+from .constants import STR_NUM_EDGES_PER_TYPE
+from .constants import STR_NODE_DATA
+from .constants import STR_EDGE_DATA
+from .constants import STR_DATA
+from .constants import STR_FORMAT
+from .constants import STR_FORMAT_DELIMITER
+from .constants import STR_NAME
+from .constants import STR_GRAPH_NAME
```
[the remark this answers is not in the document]
kylasa: Agreed. I raised this question in the grastorm-dev slack channel and @classicsong pointed me to the file python/graphstorm/config/config.py. I was following the guidelines used in this file. 

I changed the code per your comment, Hope someone will change the above mentioned file as well for consistency purposes.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,799 @@
...
+        * ``node_data``, if present, can contain keys whose count is less than
+                or equal to the node types present in the ``node_type`` list,
+                Each node type in this dictionary is expected to non-zero
+                features,
+        * ``edge_data``, if present, follows the same rules as described for
+                ``node_data`` in the earlier bullet point.
+
+    Examples:
+    ---------
+    Creating and initalizaing the singleton instance during the pipelines
```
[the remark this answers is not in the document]
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Following is the reasoning behind adapting singleton pattern for the use case in this context. 
1. Need to maintain only one instance of the metadata object throughout the application with its reference being stored in one single place. 
2. Build the metadata object once and use it everywhere it is needed, without having to build it from scratch multiple times or passing around reference to this object or storing copies of this objects reference in multiple places. One of my requirements is the need to NOT exceed more than 5 function parameters to any of the function in the entire pipeline. For this purpose singleton pattern could be an ideal pattern. All the user has to do is to get the object's reference either using class constructor call (as implemented currently) or use a class method to get the only instance of this class. 
3. Metadata information is used almost everywhere in the pipeline. All the user has to do is to get the reference to this object and use the object appropriately as needed. 
4. Information flow from this metadata class to the pipelines application logic is only in one direction and it is read-only access to the user who is using this object. 

`my_metadata` = MetadataSchema(config_file_path)`. The only issue with this code is that the object will be rebuilt wherever this piece of code is used. And all the associated logic will be executed every single time this line is used in the application. 

We have to agree on an option which adds most value to the application.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: The line of thought was that the function arguments are not valid in this case and hence the code.

Modified the code to let the system handle it appropriately.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: The main reason for this function is to ensure that the files are present, as per their suggested locations in the metadata file. This has to do with the user experience of the application. 

Imagine a use case where the pipeline is reading thousands of files. For instance, webgraph has about 700+ files for edges. And if the very last file is miss-spelled or could not be located for some reason. In this case in the current implementation the user has to wait for hours to run into a FileNotFoundError, an avoidable capex which is proactively being saved in this instance. 

Either a check is to made to ensure the presence of the file or let the system handle the case when it encountered such a thing. A decision was in the favor of proactively checking the presence of these files.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: This has to do with the way unit tests are designed. 

Within the scope of the pipeline, the use is expected to call the init method with the default value, init_maps=True. 
In the case of unit test cases, currently this flag is set to False. This way each unit test can control which part of the metadata file it will read and initialize only those data structures that are needed for that particular test. 

To enable this flexibility, some protected "init_xyz" methods had to be exposed to the user. The advantage is that the unit tests only build what part of the metadata file they are testing. But on the other side, some protected/private functions had to be exposed to the user. This can be avoided, by creating a child class for unit tests. If the child class is used, then it has its own complications with the singleton design of the base class. To share the same state of the member variables between the instances of the parent and child class instances, the constructor has to be something like 

`class MetadataSchema(object):
    _shared_state = {}
    def __new__(cls, *args, **kwargs):
        obj = super(MetadataSchema, cls).__new__(cls)
        obj.__dict__ = cls._shared_state
        return obj
`
which is a complicated piece of code and hard to maintain. 

That said, we have to agree on the usage and change the unit tests appropriate. 
If we want to avoid exposing some of the init methods and create/initialize the member variables in the constructor then we have to appropriate modify the unit test cases, which means that for every unit test we have to create a VALID metadata file which increases the code and setup time for each testcase.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: As described earlier, this was created so that unit tests could be performed with minimal setup.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Removed.

python/graphstorm/graphpartitioning/metadata_schema.py
thvasilo: Is the only reason this exists to make unit tests easier?

python/graphstorm/graphpartitioning/metadata_schema.py
thvasilo: > To enable this flexibility, some protected "init_xyz" methods had to be exposed to the user.

I'm not sure if I understand this. In Python a user can call any function whether it starts with a letter or an underscore. The notion of private/protected does not exist in the language the same way it does for say C++ or Java. The unit tests can just instantiate an object of the class and call functions that begin with and underscore as easily as they can call "public" functions.

> If the child class is used, then it has its own complications with the singleton design of the base class.

Again, there is no reason to use a singleton design here.

thvasilo: > for every unit test we have to create a VALID metadata file which increases the code and setup time for each testcase.

You can just create a static metadata file and check it in the code under tests/resources.

python/graphstorm/graphpartitioning/metadata_schema.py
thvasilo: > Need to maintain only one instance of the metadata object throughout the application with its reference being stored in one single place.

Why not instantiate an object and pass it as a function argument wherever it's needed then?

> One of my requirements is the need to NOT exceed more than 5 function parameters to any of the function in the entire pipeline.

As mentioned previously, the 5 argument rule is a code style _suggestion_ I don't think we should modify our design just to try to stick to this rule. Where it makes sense to have more than 5 arguments it's OK, as long as we try to keep it reasonable (15 arguments is definitely too many, 6-7 can be OK).

> Metadata information is used almost everywhere in the pipeline. All the user has to do is to get the reference to this object and use the object appropriately as needed.

The can do that without a singleton pattern though?

> The only issue with this code is that the object will be rebuilt wherever this piece of code is used.

The idea is that you initialize the object once then use it wherever necessary. There's no need to re-instantiate it.

kylasa: The way I understand your comments is that create an instance of the metadata schema object and pass it around where ever it is needed. 

Please note that **all the modules use MetadataSchema object, expect 1 or 2 modules.** 

What this means is that literally every function in this application will have metadata schema object as a parameter !!!, Doesn't this look odd ? Doesn't this call usage of some pattern (singleton, factory etc...) ?. 

If you still insist on creating metadata object and passing it around, I will do so. Please give it a thought and confirm. This change will trigger a lot of code changes on my end and consequently re-coding and testing time on it.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: As I described in my earlier comment, I think there is a need to use a pattern to avoid passing the same argument to all 95% of the functions, if not all. Can we do it without using a pattern, Yes we can. But the cost of doing so is passing the same argument to all the functions in the application. 

Regarding creating a static metadata file and checking it under tests/resources. ??
Yes we can do that, if there are a couple of files. Please note that there are various combinations to consider, Following are the boolean flags which control the creation of a metadata file. 
1. should the file be relative or absolute.
2. Does the edge files exist at the said location
3. Does the node feature files exist at the said location
4. Does the edge feature files exist at the said location
5. The type of delimiter to use in node features, edge features, edge files, 
6. Are these feature files absolute or relative. 

So as you can see, we can create 100's of metadata files and I don't this we can checkin all these files in the tests/resouces directory. It is easy to create them on the fly and throw them away.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Yes. This is the only reason they are present. To enable modular tests with minimal setup.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: We can definitely make them private. Once they are made so, either we can call all of them from the constructor or call them individually from the derived/child class to enable modular testing. As indicated earlier, if we use child/derived class it comes with its own complications which needs to include some funny code in the base class constructor.

python/graphstorm/graphpartitioning/metadata_schema.py
thvasilo: Only the constructor of each class that needs the metadata schema needs to take it as an argument, not every function.

```python
class MetadataSchema():
    def __init__(self, config_file):
        # Init here

class MyOtherClassThatNeedsMetadata():
    def __init__(self, metatadata_schema: MetadataSchema):
        self.metadata_schema = metatadata_schema
    def method_that_needs_metadata(self):
        # We don't need to pass metadata again to this or any class method
        do_something_with_metadata(self.metadata_schema)
```

So at most this would require the constructor of the class that needs the metadata to take it as an argument.

How is the singleton pattern more appropriate than this?

thvasilo: Resolving in favor of regular class object.

kylasa:
As per the discussion with @thvasilo on Friday 06/02 @ 3 PM, following changes are made in the latest commit.
1. MetadataSchema is now a regular class. User is expected to store a reference to this class and use it appropriately.
2. To enable modular unit testing, constructor takes in an argument, init_maps, Unit test cases pass this flag as False and control the construction of individual data structures so that small pieces of code is tested in a modular fashion. 
3. Unit tests are modified accordingly which now have a UTMetadataSchema class and control the construction of individual data structures in the base class. 
4. All the functions are sorted, alphabetically, in the MetadataSchema class.

tests/unit-tests/graphpartitioning/test_metadata.py
```diff
@@ -27,12 +27,37 @@
...
+        super()._init_etype_maps()
+
+    def init_etype_files(self):
+        super()._init_etype_files()
+
+    def init_ntype_feature_files(self):
+        super()._init_ntype_feature_files()
+
+    def init_etype_feature_files(self):
+        super()._init_etype_feature_files()
```
thvasilo, line 53: This is not really needed right? You can just do

```
my_meta = MetadataSchema(file_name, base_dir, init_maps=False)

my_meta._init_global_nid_offsets()
```

kylasa: I was hoping to put protected methods within a wrapper class and invoke methods on the wrapper class. I was told that functions with prefix "_" are only used within the class and NOT outside the class. 

Changed the code now and these protected methods are directly invoked where ever necessary, as indicated above.

thvasilo: In Python we generally we use `_name` to indicate a method that shouldn't be used externally, but in reality there's no mechanism preventing us from doing that.

In terms of unit tests, it's usually bad practice to [unit test private methods](https://softwareengineering.stackexchange.com/questions/380287/why-is-unit-testing-private-methods-considered-as-bad-practice). Wrapping the class and exposing the private methods is basically the same with an extra step.

If we wanted to follow all best practices here I'd suggest thinking carefully about which methods should be exposed to the caller ("public") and only testing those, thereby decoupling the implementation from the tests. But I think that would require quite a bit of redesigning the code here.

kylasa: I agree with your comments and reasoning. But the problem is once we decouple the code by truely exposing the only needed interface to the users then that would make unit-tests a much more involved proposition and it will become extremely difficult to test the said module effectively with meaningful test cases. Anyways we can argue all we want about this and settle on a agreeable middleground. 

I notice that you still did not approve this PR yet !

Please you still have concerns and would like to propose some changes to the code please let me know and I will get them done promptly.

thvasilo: I'm going to wait for @classicsong to take a look and then we can talk about approving.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+        src_type: str
+            Either node type or edge type from the input graph.
+        """
+        for filename in data_files:
+            if not os.path.isabs(filename):
+                if not os.path.isfile(os.path.join(self._basedir, filename)):
+                    raise FileNotFoundError(
+                        f"File: {os.path.join(self._basedir, filename)} "
+                        f"does not exist for {file_type}"
+                    )
```
zheng-da, line 187: you can just use assert to make the code shorter

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+            else:
+                if not os.path.isfile(filename):
+                    raise FileNotFoundError(
+                        f"File: {filename} does not exist for {file_type}."
+                    )
```
zheng-da, line 192: the same comment here

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+        assert (
+            len(self._data[STR_NUM_EDGES_PER_TYPE]) > 0
+        ), "Invalid number of edges"
+
+        for idx, num_nodes in enumerate(self._data[STR_NUM_NODES_PER_TYPE]):
+            if num_nodes > 0:
+                continue
+            raise ValueError(
+                f"No. of nodes at index: {idx} is not a valid value."
+            )
```
zheng-da, line 279: use assert.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+
+        # Check for each existing edge feature, corresponding edge type is valid
+        for key, _ in self._etype_feature_files.items():
+            etype = key[0]
+            if etype in self.etypes:
+                continue
+            raise ValueError(
+                f"Edge Type: {etype} is not present in the list of edge types"
+            )
```
zheng-da, line 318: user assert

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+            raise ValueError("Invalid value for No. of nodes per type")
+
+        prefix_sum = np.cumsum([0] + ntype_counts)
+        starts = prefix_sum[:-1]
+        ends = prefix_sum[1:]
+        ranges = zip(starts, ends)
+        self._global_nid_offsets = dict(zip(ntypes, ranges))
+
+    def _init_etype_maps(self):
+        """Initialization function to build unique id maps for node types of
```
zheng-da, line 384: edge type

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+        the input graph. This function creates a list of edge types present in
+        the graph, and unique-id <-> edge type and reverse maps.
+        """
+        self._etypes = self._data[STR_EDGE_TYPE]
+        self._etype_id_map = {
+            etype_name: idx for idx, etype_name in enumerate(self._etypes)
+        }
+        self._id_etype_map = dict(enumerate(self._etypes))
```
zheng-da, line 392: why do you need this? can't we use `self._etypes` ?

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+            etype1 is a edge type from the graph and
+            [feature1, feature2, ...] is a list of features associated with
+            the etype1 edge type.
+        """
+        self._etype_features = {}
+        edge_data = self._data.get(STR_EDGE_DATA, {})
+        for etype in self._data[STR_EDGE_TYPE]:
+            features = edge_data.get(etype, {})
+            if len(features) > 0:
+                self._etype_features[etype] = list(features.keys())
```
zheng-da, line 408: the way that the code is written is weird. isn't this
```
if etype in edge_data:
    ....
```

zheng-da: also, why do you need `self._etype_features`? isn't `edge_data` used for this purpose?

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+        try:
+            for etype, etype_info in self._data[STR_EDGES].items():
+                file_type = etype_info[STR_FORMAT][STR_NAME]
+                delimiter = etype_info[STR_FORMAT][STR_FORMAT_DELIMITER]
+                data_files = etype_info[STR_DATA]
+
+                self.__check_ifexists(data_files, etype)
+                self._etype_files[etype] = (file_type, delimiter, data_files)
+
+        except KeyError as exp:
```
zheng-da, line 469: why don't you use assert?

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+            logging.error(
+                "Error while reading files for edeges of the input graph."
+                "Please check the metadata file for correctness for loading "
+                "edge files."
+            )
+            raise exp
+
+        if len(self._etype_files) is not len(self._etypes):
```
zheng-da, line 477: use `==`

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+            raise ValueError(
+                f"In the metadata file there are some edges"
+                f" for which there are no corresponding edge files."
+                f" etypes = {self._etypes} are all the edge files. "
+                f" etypes, for which edge files are present: "
+                f" {self.etype_files.keys()}"
+            )
```
zheng-da, line 484: use assert

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+                    delimiter = feature_info[STR_FORMAT].get(
+                        STR_FORMAT_DELIMITER, ","
+                    )
+
+                self.__check_ifexists(data_files, f"{ntype}/{feature_name}")
+                self._ntype_feature_files[(ntype, feature_name)] = (
+                    file_type,
+                    delimiter,
+                    data_files,
+                )
```
zheng-da, line 545: all comments above are applied here as well.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+
+    ############################## Public Methods ##############################
```
zheng-da, line 547: why do you use methods instead of properties?

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+    def etype_id_map(self):
+        """Property to get the dictionary which stores (etype, edge_id) as
+        the key-value pairs.
+        """
+        return self._etype_id_map
+
+    @property
+    def etype_features(self):
+        """Property to get a list of feature names for each edge type."""
+        return self._etype_features
```
zheng-da, line 674: why do you define both properties and methods for this?

zheng-da:
overall comments: the code is overcomplicated. I don't think you need so many lines of code to define the MetadataSchema.

python/graphstorm/graphpartitioning/array_readwriter/__init__.py
```diff
@@ -0,0 +1,17 @@
+"""
```
zheng-da, line 1: as we discussed, the code doesn't belong to the library.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: The idea of creating a map between etype and etype_id and vice versa is to decouple the user to do so. This is the only one place which assigns ids to etypes. And if there is a logic change then there is only one place to change. This also serves to abstract away the logic from the application, which should be agnostic about how the ids are assigned to the etypes in the input graph.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: This has to do with the user experience. Following is the reason:
1. "edge_type", list of etypes in the graph, is the used to extract features from "edge_data". This is used to filter out any erroneous etypes which may show up in the "edge_data". If we read "etypes" and "features" from "edge_data" then we might run into erroneous cases where a certain etype is present in the "edge_data" but NOT present in the "edge_type" field. We are avoiding such a scenario and instead of throwing some wierd errors in such case we are gracefully continuing with the pipeline execution -- in this case by ignoring etypes/features which might be present in the "edge_data" but not present in the "edge_type".  This ensures that the user experience is enhanced. 

"self._etype_features" contains a list of features for each valid etype in the input dataset. Because of the above reason we cannot rely on the correctness of the input dataset.

By building self._etype_features, we are abstracting away the application to figure out which features exist for a given etype. So that way any changes to "edge_data" will result only in the changes in this file - as opposed to any changes in the application itself.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: The way I understand "assert" is to ensure smooth functioning of the code and validate conditions which are "expected" to be true at some point with the application logic. I think throwing AssertionError when actually a more meaningful/appropriate errors like FileNotFoundError or KeyError can be raised is very in-appropriate. 

In favor of saving time for the CI process, I changed all the error conditions to use assert statements throughout the file.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+        ), "Edge types with features does not match edge type counts."
+
+        # Check for each existing node feature, corresponding node type is valid
+        for key, _ in self._ntype_feature_files.items():
+            ntype = key[0]
+            if ntype in self.ntypes:
+                continue
+            raise ValueError(
+                f"Node Type: {ntype} is not present in the list of node types."
+            )
```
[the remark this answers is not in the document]
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+            etype1 is a string describing the edge type
+            and [a, b) is the range of global_edge_ids for the edge type
+            `etype1`.
+        """
+        etypes = self._data[STR_EDGE_TYPE]
+        etype_counts = self._data[STR_NUM_EDGES_PER_TYPE]
+        for count in etype_counts:
+            if isinstance(count, int) and count > 0:
+                continue
+            raise ValueError("Invalid value for No. of nodes per type")
```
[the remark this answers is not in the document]
kylasa: Done

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,730 @@
...
+            Edge type from the input graph.
+
+        Returns
+        -------
+        tuple: (str, str, list(str))
+            File type (a valid file type which are ``numpy`` or ``csv``),
+            valid delimiter and list of file paths in which edges are
+            stored for the given edge type.
+        """
+        return self._etype_files[etype]
```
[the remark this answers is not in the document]
kylasa: Python3 does not allow properties with parameters. This error is caught with pylint as "properties-with-parameters".

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Python does not allow properties with parameters.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Please note that property returns the entire data structure, where as the get method will lookup into the dictionary and returns value associated with that key. Its for ease of use for the application.

python/graphstorm/graphpartitioning/array_readwriter/__init__.py
kylasa: As I agreed with you earlier... moved this code to the test folder. Since it is expected to be used ONLY for testing purposes. The application needs only reading of the files, application never writes files to the disk. Just for reading files from the disk, we can use a much simplified function to read files from the disk.

kylasa:
@zheng-da, A couple of general comments on the way the code is structured. 
1. More appropriate errors conditions were used earlier like FileNotFoundError, KeyError etc... but per your request all such instances were changed to "assert" conditions. 
2. The general idea of this module is to abstract away the user of this class from the structure of the metadata file. That way we ensure that in future only this module will change if metadata.json were to be replaced by some other structured file. 
3. There are several validation steps happening when the metadata.json is being processed and converted to appropriate data structures. It is NOT guaranteed that the input metadata.json file is always valid. Explicit validation rules are enforced on this module once the metadata.json file is parsed. This should explain the code the way it is structured. 
4. pylint does now allow properties with parameters, which is the reason why "get" methods are in place for such methods. Following error is thrown by the pylint tool. "R0206: Cannot have defined parameters for properties (property-with-parameters)"


That said, I generally follow the following rules when I design a class. 
1. By design, I will make all the member variables "private", unless it has very specific reasons to be protected/public. In this case all the member variables should start with "__" a double underscore. But this design principle is not used in the graphstorm repo, for some reason. I checked the classes in the "gconstruct" folder, and in that folder all the class member variables are prefixed with "_". 
2. The same above principle I would like to apply to the member functions as well. In the current implementation, some member functions like "_init_etype_maps", "_init_etype_features" etc... are still marked as such. This means that use can invoke these functions from within the application. 
3. The only reason why these above functions exist in such state is to enable more modular unit testing. This way ensures that unit testing can be done on small pieces of code more easily. As opposed to testing very large pieces of code. But this is against the design principle I normally follow when writing a class. I guess this is expected and treated as normal based on our guidelines.

python/graphstorm/graphpartitioning/metadata_schema.py
zheng-da: how is `self._id_etype_map` different from `self._etypes`?

zheng-da: `self._id_etype_map[i] == self._etypes[i]`

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,706 @@
...
+            etype1: [feature1, feature2, ....] where
+            etype1 is a edge type from the graph and
+            [feature1, feature2, ...] is a list of features associated with
+            the etype1 edge type.
+        """
+        self._etype_features = {}
+        edge_data = self._data.get(STR_EDGE_DATA, {})
+        for etype in self._data[STR_EDGE_TYPE]:
+            features = edge_data.get(etype, {})
+            if len(features) > 0:
```
zheng-da, line 393: this is not a normal way of checking if `etype` exists. this is a common problem. i see you are doing it everywhere. i pointed out one place, you should change all places.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,706 @@
...
+                  are stored for this edge type.
+        """
+        self._etype_feature_files = {}
+        edge_data = self._data.get(STR_EDGE_DATA, {})
+        for etype, etype_info in edge_data.items():
+            for feature_name, feature_info in etype_info.items():
+                data_files = feature_info.get(STR_DATA, [])
+                file_type = None
+                delimiter = None
+                if len(data_files) > 0:
```
zheng-da, line 419: the same comment above

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,706 @@
...
+    def _init_ntype_maps(self):
+        """Initialization function to build unique id maps for node types of
+        the input graph. This function creates a list of node types present in
+        the graph, and unique-id <-> node type and reverse maps.
+        """
+        self._ntypes = self._data[STR_NODE_TYPE]
+        self._ntype_id_map = {
+            ntype_name: idx for idx, ntype_name in enumerate(self._ntypes)
+        }
+        self._id_ntype_map = dict(enumerate(self._ntypes))
```
zheng-da, line 471: the same comment here.

python/graphstorm/graphpartitioning/metadata_schema.py
zheng-da: since you already return them as a property, why cannot you just do something like
```
schema.etype_info[etype]
```

python/graphstorm/graphpartitioning/metadata_schema.py
zheng-da: i don't see how it is easier.

tests/unit-tests/graphpartitioning/test_metadata.py
```diff
@@ -0,0 +1,731 @@
...
+
+
+@pytest.mark.parametrize("ntypes", [[], [NODE1], [NODE1, NODE2]])
+def test_ntypes(ntypes):
+    """Unit Tests to test various scenarios when only nodes are in the input
+    graph dataset.
+    """
+    input_dict = {}
+    with tempfile.TemporaryDirectory() as root_dir:
+
```
zheng-da, line 66: remove the empty line.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,706 @@
...
+                valid node types are in the ``node_type`` list
+           * Edge type with edge features should be a valie edge type,
+                valid edge types are in the ``edge_type`` list
+        """
+        assert isinstance(self._data[STR_GRAPH_NAME], str), "Invalid graph name"
+        assert len(self._data[STR_GRAPH_NAME]) > 0, "Invalid graph name"
+
+        assert (
+            len(self._data[STR_NUM_NODES_PER_TYPE]) > 0
+        ), "Invalid number of nodes"
```
zheng-da, line 256: do we need the bracket? please follow the convention of other code.

zheng-da: it seems your own coding style is not consistent. sometimes you have bracket and sometimes you don't use bracket.

tests/unit-tests/graphpartitioning/test_metadata.py
```diff
@@ -0,0 +1,731 @@
...
+        obj = MetadataSchema(METADATA_NAME, root_dir, init_maps=False)
+        obj._init_etype_maps()
+
+        # Check the initialization
+        try:
+            obj._init_etype_files()
+
+            if create_files:
+                for idx, etype in enumerate(etypes):
+                    (file_type, delimiter, data_files) = obj.get_etype_info(
```
zheng-da, line 206: you don't need a bracket.

tests/unit-tests/graphpartitioning/test_metadata.py
```diff
@@ -0,0 +1,731 @@
...
+                        etype
+                    )
```
zheng-da, line 208: do you really need to break the line for the argument?

tests/unit-tests/graphpartitioning/test_metadata.py
```diff
@@ -0,0 +1,731 @@
...
+@pytest.mark.parametrize("delimiter", [" ", "/"])
+@pytest.mark.parametrize("create_files", [True, False])
+@pytest.mark.parametrize("prefix_dir", ["", "node_data"])
+@pytest.mark.parametrize("is_absolute", [True, False])
+def test_node_features(
+    ntypes, file_type, delimiter, create_files, prefix_dir, is_absolute
+):
+    """Unit tests when node features are present in the input graph."""
+
+    def run_test(root_dir, input_dict):
```
zheng-da, line 235: this is a very large function. can you define the function outside the test function?

tests/unit-tests/graphpartitioning/test_metadata.py
```diff
@@ -0,0 +1,731 @@
...
+
+        except AssertionError as exp:
+            if create_files:
+                raise exp
+            else:
+                pass
+        finally:
+            obj = None
+
+    def update_schema(input_dict, file_metadata):
```
zheng-da, line 275: the same here.

tests/unit-tests/graphpartitioning/test_metadata.py
```diff
@@ -0,0 +1,731 @@
...
+                    ), f"No. of edge feature files does not match."
+        except AssertionError as exp:
+            if create_files:
+                raise exp
+            else:
+                pass
+
+
+@pytest.mark.parametrize(
+    "metadata",
```
zheng-da, line 376: this is supercomplex.

zheng-da:
I agree with the principal you applied here. In terms of assert, I also agree in general that assert is to make sure that the input data should be right most of the time. however, you are using assert to check input arguments while using exceptions for other arguments. by following your principle, you should use exceptions for all arguments, not some of them. however, the code becomes super complex. I would prefer to just use assert for everything.

The problem with your code is that the coding style is inconsistent. you sometimes choose to do one way, but sometimes choose to do another way. it's not very clear to me why you don't do everything in one way. For example, even assert you sometimes use bracket, sometimes don't use bracket; sometimes you have error messages and sometimes you don't have error messages.

zheng-da:
I created a branch `dist_part`. please merge your PR to this dev branch first.

.github/workflow_scripts/lint_check.sh
```diff
@@ -17,4 +17,6 @@ pylint --rcfile=./tests/lint/pylintrc ./python/graphstorm/tracker/
 pylint --rcfile=./tests/lint/pylintrc ./python/graphstorm/run/
 pylint --rcfile=./tests/lint/pylintrc ./python/graphstorm/utils.py
 
-pylint --rcfile=./tests/lint/pylintrc ./python/graphstorm/sagemaker/
\ No newline at end of file
+pylint --rcfile=./tests/lint/pylintrc ./python/graphstorm/sagemaker/
+
+pylint --rcfile=./tests/lint/pylintrc ./python/graphstorm/graphpartitioning/
```
zheng-da, line 22: can you double check if pylint is working?

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,706 @@
...
+    >>> graph_ntypes = metadata_obj.ntypes
+    >>> print(graph_ntypes)
+    >>> ["ntype1", "ntype2"]
+
+    In the above code snipped ``ntypes`` property is invoked to get a list of
+    node types present in the input graph, whose return value is a list of
+    strings. The above code snippet assumes that the input graph has two
+    node types namely ``ntype1`` and ``ntype2``.
+    """
+
```
classicsong, line 144: Please provide descriptions of input args.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,706 @@
...
+    ############################## Private Methods #############################
+
+    def __check_ifexists(self, data_files, file_type):
+        """Helper function to verify the presence of a set of filenames.
+
+        Argument
+        --------
+        data_files: list of str
+            A set of file names, either in absolute or in relative formats.
+        src_type: str
```
classicsong, line 177: the second argument is file_type, not src_type.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,706 @@
...
+            Either node type or edge type from the input graph.
+        """
+        for file_name in data_files:
```
classicsong, line 180: Why not `for file_path in data_files:`

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,706 @@
...
+            file_path = file_name
+            if not os.path.isabs(file_path):
+                file_path = os.path.join(self._basedir, file_path)
+            assert os.path.isfile(
```
classicsong, line 184: We do not need to pass-in file_type. We can do 
```
assert self._check_ifexists(data_files), f"Files {data_files} does not exist for {file_type}."
```

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,706 @@
...
+            the input graph dataset
+        """
+        self._basedir = None
+        if base_dir is None:
+            self._basedir = os.getcwd()
+        else:
+            self._basedir = base_dir
+
+        self._metadata_path = os.path.join(
+            os.path.join(self._basedir, file_name)
```
classicsong, line 226: What if `file_name` is in absolute path format?

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,706 @@
...
+            etype1 is a string describing the edge type
+            and [a, b) is the range of global_edge_ids for the edge type
+            `etype1`.
+        """
+        etypes = self._data[STR_EDGE_TYPE]
+        etype_counts = self._data[STR_NUM_EDGES_PER_TYPE]
+        for count in etype_counts:
+            assert (
+                isinstance(count, int) and count > 0
+            ), "Invalid value for No. of nodes per type"
```
classicsong, line 330: This is etype_count, why the error is about No. of nodes per type?

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,706 @@
...
+            the etype1 edge type.
+        """
+        self._etype_features = {}
+        edge_data = self._data.get(STR_EDGE_DATA, {})
+        for etype in self._data[STR_EDGE_TYPE]:
+            features = edge_data.get(etype, {})
+            if len(features) > 0:
+                self._etype_features[etype] = list(features.keys())
+
+    def _init_etype_feature_files(
```
classicsong, line 396:
```suggestion
    def _init_etype_feature_files(self):
```

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,706 @@
...
+                    )
+
+                self.__check_ifexists(data_files, f"{etype}/{feature_name}")
+                self._etype_feature_files[(etype, feature_name)] = (
+                    file_type,
+                    delimiter,
+                    data_files,
+                )
+
+    def _init_etype_files(
```
classicsong, line 432:
```suggestion
    def _init_etype_files(self):
```

tests/unit-tests/graphpartitioning/array_readwriter/csv.py
```diff
@@ -0,0 +1,39 @@
...
+
+import pandas as pd
+import pyarrow
+import pyarrow.csv
+
+from .registry import register_array_parser
+
+
+@register_array_parser("csv")
+class CSVArrayParser(object):
```
classicsong, line 11: Any doc about this CSVArrayParser? Why we need this?

tests/unit-tests/graphpartitioning/graph_dataset.py
```diff
@@ -0,0 +1,370 @@
...
+
+import numpy as np
+import pyarrow as pa
+import torch
+
+import array_readwriter
+
+################### Constants used in pipelines unit tests ####################
+
+METADATA_NAME = "metadata.json"
```
classicsong, line 30: Shouldn't you import these constants from graphstorm.graphpartition.constants?

tests/unit-tests/graphpartitioning/test_metadata.py
```diff
@@ -0,0 +1,731 @@
...
+                        file_type == actual_file_type
+                    ), f"File type does not match."
+                    assert (
+                        delimiter == actual_delimiter
+                    ), f"Delimiters does not match."
+                    assert (
+                        len(data_files) == 5
+                    ), f"No. of node feature files does not match."
+
+        except AssertionError as exp:
```
classicsong, line 267: What is the purpose of this except?

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Removed for now. If needed will add later that will justify the need at the time.

python/graphstorm/graphpartitioning/metadata_schema.py
```diff
@@ -0,0 +1,706 @@
...
+            ntype1: [feature1, feature2, ...] where,
+            ntype1 is a node type from the graph and
+            [feature1, feature2, ....] is a list of features associated with
+            the ntype1 node type.
+        """
+        self._ntype_features = {}
+        node_data = self._data.get(STR_NODE_DATA, {})
+        for ntype in self._data[STR_NODE_TYPE]:
+            features = node_data.get(ntype, {})
+            if len(features) > 0:
```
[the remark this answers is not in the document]
kylasa: Addressed this comment as indicated earlier.

tests/unit-tests/graphpartitioning/test_metadata.py
```diff
@@ -0,0 +1,731 @@
...
+
+        assert (
+            obj.data == input_dict
+        ), f"Expected and Actual metadata objects does not match."
+
+    input_dict = {}
+    add_graph_name(input_dict)
+
+    with tempfile.TemporaryDirectory() as root_dir:
+
```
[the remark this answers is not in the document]
kylasa: This file has been re-organized from scratch... mostly to address @thvasilo comments. Now this is outdated.

tests/unit-tests/graphpartitioning/test_metadata.py
kylasa: Done.

tests/unit-tests/graphpartitioning/test_metadata.py
kylasa: Done by lintrunner.

tests/unit-tests/graphpartitioning/test_metadata.py
kylasa: Done by lintrunner.

tests/unit-tests/graphpartitioning/test_metadata.py
kylasa: This file has been re-organized from scratch... mostly to address @thvasilo comments. Now this is outdated.

tests/unit-tests/graphpartitioning/test_metadata.py
kylasa: This file has been re-organized from scratch... mostly to address @thvasilo comments. Now this is outdated.

tests/unit-tests/graphpartitioning/test_metadata.py
kylasa: Removed as this has already been tested by other test cases.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Corrected as indicated.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: I do not understand your comment. Here ``file_type`` is used to provide additional debugging information in the assertion error message. otherwise it might be hard to figure out why this error message is getting generated and from where. This function is invoked from three different places.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: The semantics is to provide a file_name, and a base_directory. If these arguments are incorrect or not as expected, then system will through/raise appropriate error. This behavior is the result of earlier comments.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Corrected. the typo.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

python/graphstorm/graphpartitioning/metadata_schema.py
kylasa: Done.

tests/unit-tests/graphpartitioning/array_readwriter/csv.py
kylasa: This is used to create test datasets on the fly. This is also a apart of integration testing for the end-to-end pipeline.

kylasa: The original code, as existing in the DGL repo today, was created by Minjie when the end-to-end testing pipeline was created around the last 3 months of last year.

tests/unit-tests/graphpartitioning/test_metadata.py
kylasa: This file has been re-organized from scratch... mostly to address @thvasilo comments. Now this is outdated.

kylasa:
As I indicated several times earlier, ''all the code styling/formatting" is done by the tool ``lintrunner`` and is configured as used in the DGL repo. On top of this tool, pylint is also executed as required by the graphstorm repo. 

They way lintrunner is treating python modules and python files is different. Majority of the comments from Da is about code styling (why a line is split into multiple lines, why brackets are used for function return parameters etc....) All these exist because of lintrunner styling/formatting. 

If this is NOT acceptable or creating inconsistencies in the files, then we need to agree on a set guidelines so that these can be resolved or completely do-away with the use of lintrunner. I will let you guys decide on this one.

kylasa:
About the use of assert statements and raising appropriate system errors as dictated by the system. Per @zheng-da, asserts are preferred, per @thvasilo earlier comments system will handle these errors appropriately. 

During the first round of changes, I agreed with @thvasilo and changed appropriately. Post @zheng-da comments, this same code snippets were changed to use assert statements. 

We have to avoid changing the same code multiple times to agree with multiple code reviewers preferences.

zheng-da (approved):

zheng-da:
This PR should be merged to the `dist_part` branch

thvasilo (approved):

thvasilo:
Approving for merge into a dev branch.

PR #219, opened 2023-05-31
Remove GSF dependency from SageMaker launch scripts.

thvasilo:
*Description of changes:*

We remove the import of the graphstorm module, allowing the scripts to be ran from a Python environment that doesn't have GraphStorm installed.
Small re-arrangement of the imports to match PEP8 guidelines.

### DIFF
--- a/sagemaker/launch/launch_infer.py
+++ b/sagemaker/launch/launch_infer.py
@@ -16,16 +16,20 @@
     Launch SageMaker training task
 """
 import os
-import sagemaker
 import argparse
+
 import boto3 # pylint: disable=import-error
-
-from graphstorm.config import SUPPORTED_TASKS
-
-from sagemaker.pytorch.estimator import PyTorch
 import sagemaker
+from sagemaker.pytorch.estimator import PyTorch
 
 INSTANCE_TYPE = "ml.g4dn.12xlarge"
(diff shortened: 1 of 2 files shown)


sagemaker/launch/launch_infer.py
```diff
@@ -16,16 +16,20 @@
...
-
-from graphstorm.config import SUPPORTED_TASKS
 
-from sagemaker.pytorch.estimator import PyTorch
+import boto3 # pylint: disable=import-error
 import sagemaker
+from sagemaker.pytorch.estimator import PyTorch
 
 INSTANCE_TYPE = "ml.g4dn.12xlarge"
+SUPPORTED_TASKS = {
```
classicsong, line 26: Is it possible to add a unit test to ensure SUPPORTED_TASKS is synced with graphstorm.config.SUPPORTED_TASKS ?

thvasilo: I'd be hesitant to add that because these scripts are supposed to be live outside the library, so we'd need to import them in some hacky way. 

I'd argue it's very unlikely this list will change often and if it does the SageMaker regression tests should catch that.

classicsong: That makes sense.

Issue #220, opened 2023-06-03
[Bug] Graph construction will fail if a worker returns a data object larger than 2GB


classicsong:
Graph construction work uses pickle to return data to master process (https://github.com/awslabs/graphstorm/blob/main/python/graphstorm/gconstruct/utils.py#L59-L61). If a worker returns a data object larger than 2GB it will crash the process and raise the following error:

```
  File "<string>", line 2, in put
  File "/usr/lib64/python3.7/multiprocessing/managers.py", line 818, in _callmethod
    conn.send((self._id, methodname, args, kwds))
  File "/usr/lib64/python3.7/multiprocessing/connection.py", line 206, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
  File "/usr/lib64/python3.7/multiprocessing/connection.py", line 393, in _send_bytes
    header = struct.pack("!i", n)
struct.error: 'i' format requires -2147483648 <= number <= 2147483647
```

Suggested solution: Convert data to be return into torch shared memory.

PR #222, opened 2023-06-05
Fix pickle max object can not exceed 2 GB issue.

classicsong:
*Issue #, if available:*
#220 

*Description of changes:*


python/graphstorm/gconstruct/utils.py
```diff
@@ -30,6 +30,112 @@
...
+    if th.is_tensor(data):
+        return data.share_memory_()
+    elif isinstance(data, np.ndarray):
+        # only handle data that can be converted to torch tensor
+        if data.dtype in [np.float64, np.float32, np.float16,
+                          np.complex64, np.complex128, np.int64,
+                          np.int32, np.int16, np.int8,
+                          np.uint8, np.bool]:
+            return th.tensor(data).share_memory_()
+        return data
```
zheng-da, line 54: what does this mean? the numpy array is returned directly? does it mean we give up in this case?

python/graphstorm/gconstruct/utils.py
```diff
@@ -30,6 +30,112 @@
...
+    elif isinstance(data, dict):
```
zheng-da, line 55: does it mean that this is a dict of tensors? i suppose it doesn't work if the value is a primitive, such as integer.

python/graphstorm/gconstruct/utils.py
classicsong: Yes. In this case, we do nothing.
This happens, when data is a numpy.object_.

python/graphstorm/gconstruct/utils.py
classicsong: this is dict of tensors or dict of other basic types like dict, list, etc.
It will not work with primitives.

python/graphstorm/gconstruct/utils.py
zheng-da: should we report an error for this case?

classicsong: In this case, we just fallback to use pickle to pack the object.
I believe the most memory consuming objects are np.arrays which are handed in L53

python/graphstorm/gconstruct/utils.py
```diff
@@ -30,6 +30,110 @@
...
+
+    Parameters
+    ----------
+    data: dict/tuple/list of tensors
+        Data returned by user_parser
+    """
+    if th.is_tensor(data):
+        data_size = data.element_size() * data.nelement()
+    elif isinstance(data, np.ndarray):
+        data_size = data.size * data.itemsize
```
zheng-da, line 117: if each element is an object, is this estimation still correct?

classicsong: No. But I think in most of the cases, it will be np arrays like features or node ids.

PR #225, opened 2023-06-06
[Fix] fix device refer before define error

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
Fix device referred before define error.

### DIFF
--- a/python/graphstorm/inference/np_infer.py
+++ b/python/graphstorm/inference/np_infer.py
@@ -95,6 +95,9 @@ class GSgnnNodePredictionInfer(GSInfer):
             label = res[1] if do_eval else None
         sys_tracker.check('compute embeddings')
 
+        device = th.device(f"cuda:{self.dev_id}") \
+                if self.dev_id >= 0 else th.device("cpu")
+
         # do evaluation first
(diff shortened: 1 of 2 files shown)


python/graphstorm/inference/np_infer.py
```diff
@@ -122,8 +125,6 @@ def infer(self, loader, save_embed_path, save_prediction_path=None,
                 ntype_emb = embs[ntype]
             embeddings = {ntype: ntype_emb}
 
-            device = th.device(f"cuda:{self.dev_id}") \
-                    if self.dev_id >= 0 else th.device("cpu")
             save_gsgnn_embeddings(save_embed_path,
                 embeddings, self.rank, th.distributed.get_world_size(),
                 device=device,
```
classicsong, line 130: Can you check if lp_infer.py and ep_infer.py have the same problem?

zhjwy9343: ep_infer.py has the same issue. Fixed.

lp_infer.py does not have the issue.

classicsong (approved):
LGTM

Issue #214, opened 2023-05-30
[Roadmap] V0.1.1 Release Plan


classicsong:
The V0.1.1 release will focus on optimizing user experiences on graph construction and model training/inference,  optimizing system efficiency. V0.1.1 will also release a GraphStorm Amazon SageMaker deployment solution.

- [Improvement] Optimizing user experiences on graph construction
  - [x] #41 
  - [x] #50 
  - [x] #242 
- [Improvement] Support using edge features in GNN training.
  - [x] #63 
  - [x] #153 
- [Improvement] Improving link prediction training and inference
  - [x] #208 
  - [x] #186 
  - [x] #206 
  - [x] #267 
- [Improvement] Improve GraphStorm Installation Experience
  - [x] Add pip package for standalone installation and documents
  - [x] #205 
- [Improvement] Documentation
  - [x] #213 
- [New Feature] Amazon SageMaker support
  - [x] #59

PR #229, opened 2023-06-08
[Feature] Add two phase node data transform support and an example of floating point Max-Min normalization

classicsong:
*Issue #, if available:*
#214 

*Description of changes:*
Feature transformations like categorical feature encoding, numerical feature encoding, etc require two phase of data transform. The first phase will collect global information such as the cardinality of categorical feature or the Maximum or minimum value of numerical features. The second phase will transform the data by using the global information collected in the first phase to transform the feature data. This PR provides a general design of Two Phase data transformation mechanism and provides an example, i.e., Numerical MAX_MIN transformation.


python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -181,13 +209,28 @@ def process_node_data(process_confs, arr_merger, remap_id, num_processes=1):
...
+        num_proc = num_processes if multiprocessing else 0
+
+        two_phase_feat_ops = []
+        for op in feat_ops:
+            if isinstance(op, TwoPhaseFeatTransform):
+                two_phase_feat_ops.append(op)
+        if len(two_phase_feat_ops) > 0:
+            user_pre_parser = partial(prepare_node_data, feat_ops=two_phase_feat_ops,
+                                      read_file=read_file)
+            start = time.time()
```
zheng-da, line 221: this is not a right place to start.

python/graphstorm/gconstruct/transform.py
```diff
@@ -52,6 +53,148 @@ def feat_name(self):
...
+    def collect_info(self, info):
+        """ Store global information for the second phase data processing
+
+        Parameters
+        ----------
+        info:
+            Information to be collected
+        """
+
+class NumericalTransform(TwoPhaseFeatTransform):
```
zheng-da, line 92: this name sounds too general. if this is doing min max, it should be named with MinMaxTransform

python/graphstorm/gconstruct/transform.py
```diff
@@ -52,6 +53,148 @@ def feat_name(self):
...
+
+        max_val = np.amax(max_vals, axis=0) if len(max_vals.shape) == 2 \
+            else np.array([np.amax(max_vals, axis=0)])
+        min_val = np.amin(min_vals, axis=0) if len(min_vals.shape) == 2 \
+            else np.array([np.amin(min_vals, axis=0)])
+
+        self._max_val = max_val
+        self._min_val = min_val
+
+class NumericalMinMaxTransform(NumericalTransform):
```
zheng-da, line 165: why this should be a separate class?

python/graphstorm/gconstruct/transform.py
```diff
@@ -52,6 +53,148 @@ def feat_name(self):
...
+        """
+        assert isinstance(feats, (np.ndarray, HDF5Array)), \
+            "Feature of NumericalTransform must be numpy array or HDF5Array"
+
+        assert not np.any(self._max_val == self._min_val), \
+            f"At least one element of Max Val {self._max_val} " \
+            f"and Min Val {self._min_val} is equal. This will cause divide by zero error"
+
+        if isinstance(feats, HDF5Array):
+            feats = feats.to_numpy()
```
zheng-da, line 190: this will make all features stored in memory. it can have scalability issues.

python/graphstorm/gconstruct/utils.py
```diff
@@ -182,6 +182,26 @@ def worker_fn(worker_id, task_queue, res_queue, user_parser):
...
+
+    Parameters
+    ----------
+    phase_one_info: dict
+        A dict mapping file index to node/edge features info corresponding to ops.
+    ops: dict of FeatTransform
+        The operations run on the node/edge features of the node/edge files.
+    """
+    feat_info = {}
+    for i, finfo in phase_one_info.items():
```
zheng-da, line 196: what is `i`? is it a sequence number?

python/graphstorm/gconstruct/transform.py
```diff
@@ -272,11 +415,45 @@ def parse_feat_ops(confs):
                                                 int(conf['max_seq_length'])),
                                       conf['bert_model'],
                                       infer_batch_size=infer_batch_size)
+            elif conf['name'] == 'float_max_min':
```
zheng-da, line 418: how about call it `max_min_norm`? you are actually doing normalization.

python/graphstorm/gconstruct/construct_graph.py
classicsong: This is for pre-processing.

python/graphstorm/gconstruct/transform.py
classicsong: It is a template class for numerical transform ops. It only collects max/min info. Any numerical transform ops which needs the max/min info can inherit from NumericalTransform.
NumericalMinMaxTransform is for max/min normalization.

python/graphstorm/gconstruct/transform.py
```diff
@@ -52,6 +53,148 @@ def feat_name(self):
...
+
+        Parameters
+        ----------
+        feats: np.array
+            Data to be processed
+        """
+        assert isinstance(feats, (np.ndarray, HDF5Array)), \
+            "Feature of NumericalTransform must be numpy array or HDF5Array"
+        if isinstance(feats, HDF5Array):
+            feats = feats.to_numpy()
```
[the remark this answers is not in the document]
classicsong: That's True. Let me add to HINT here.

python/graphstorm/gconstruct/transform.py
classicsong: There can be other Numerical transform ops.

python/graphstorm/gconstruct/utils.py
classicsong: i is fileno. See https://github.com/awslabs/graphstorm/blob/main/python/graphstorm/gconstruct/utils.py#L243-L244 and https://github.com/awslabs/graphstorm/blob/main/python/graphstorm/gconstruct/construct_graph.py#L198

python/graphstorm/gconstruct/transform.py
classicsong: max_min_norm is better

python/graphstorm/gconstruct/transform.py
zheng-da: i'm not sure if this kind of abstraction is needed.

zheng-da: what it does is to collect min and max. it makes more sense to just have `NumericalMinMaxTransform`

zheng-da: do you have any other examples that this base class can be used for other operations?

zheng-da (approved):

PR #232, opened 2023-06-09
Support storing node/edge feature as float16.

classicsong:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/python/graphstorm/gconstruct/construct_graph.py
+++ b/python/graphstorm/gconstruct/construct_graph.py
@@ -43,7 +43,7 @@ FEATURE_STORE_PRECISION_FP32 = "fp32"
 FEATURE_STORE_PRECISION_FP16 = "fp16"
 
 def parse_node_data(in_file, feat_ops, label_ops, node_id_col,
-                    read_file, feat_dtype=np.flaot32):
+                    read_file, feat_dtype=np.float32):
     """ Parse node data.
 
     The function parses a node file that contains node IDs, features and labels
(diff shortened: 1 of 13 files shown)


python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -39,7 +39,11 @@
 from .id_map import NoopMap, IdMap, map_node_ids
 from .utils import multiprocessing_data_read, ExtMemArrayMerger, partition_graph
 
-def parse_node_data(in_file, feat_ops, label_ops, node_id_col, read_file):
+FEATURE_STORE_PRECISION_FP32 = "fp32"
+FEATURE_STORE_PRECISION_FP16 = "fp16"
+
+def parse_node_data(in_file, feat_ops, label_ops, node_id_col,
+                    read_file, feat_dtype=np.float32):
```
zheng-da, line 46: i would prefer that you create a feature operator that converts the data type. in this way, we don't need to pass anything argument to any of the functions.

zheng-da: in this way, users can define output data type in the JSON file to define the output data type for each feature.

python/graphstorm/gconstruct/file_io.py
```diff
@@ -244,13 +245,27 @@ def __getitem__(self, idx):
...
         """ Return Numpy array.
         """
-        return self._arr[:]
+        res = self._arr[:]
+        if self._out_dtype is not None:
+            res = res.astype(self._out_dtype)
+        return res
+
+    def astype(self, dtype):
+        """ Set the output dtype.
```
zheng-da, line 265: specify that `dtype` is numpy dtype.

zheng-da:
Can you update the README file to show out_dtype is supported.

zheng-da:
originally, I was thinking of another Transform that convert the data type if `out_dtype` is specified in a feature block. But I guess this is also fine. The only problem is that all Transform operation now needs to support an out_dtype.

classicsong:
> Can you update the README file to show out_dtype is supported.

Sure

classicsong:
> originally, I was thinking of adding another Transform that converts the data type if `out_dtype` is specified in a feature block. But I guess this is also fine. The only problem is that all Transform operation now needs to support an out_dtype.

It it not mandatory for all Transform to support out_dtype. For example, Tokenizer ignores out_dtype.

zheng-da (approved):

PR #235, opened 2023-06-09
[Bug Fix] Fix EntityRegression with --return-proba

classicsong:
*Issue #, if available:*
Node regression tasks with --return-proba flag will crash.

*Description of changes:*
Add predict_proba in EntityRegression


zhjwy9343 (approved):
LGTM

PR #236, opened 2023-06-09
change the way of saving ID map.

zheng-da:
*Description of changes:*
We cannot convert the ID map to numpy arrays first. Numpy arrays consume a large amount of memory to store a large string array.


zhjwy9343 (approved):
LGTM

Issue #239, opened 2023-06-11
`out_dtype` doesn't work if there is no `transform`


zheng-da:
For example, the block below doesn't output float16 tensors.
```
                {
                        "node_type":    "campaign",
                        "format":       {"name": "hdf5"},
                        "files":        "nfeats-campaign/campaign-embedding/*.hdf5",
                        "features":     [
                                {
                                        "feature_col":  "my_tensor",
                                        "feature_name": "feat",
                                        "out_dtype":    "float16"
                                }
                        ]
                },
```

PR #240, opened 2023-06-12
Fix bug of some transform ops do not return float16 data if out_dtype is set to float16

classicsong:
*Issue #, if available:*
#239 

*Description of changes:*


zheng-da (approved):
LGTM.
I think the current way of supporting out dtype may cause bugs. we have to modify many places to convert data type. if we creat an additional `ConvertDTypeTransform` whenever users specify `out_dtype`, we may be able to avoid modifications everywhere.

PR #245, opened 2023-06-13
[Bug Fix] Use torch multiprocess in graph construction.

classicsong:
*Issue #, if available:*
Python multiprocess can not handle torch shared mem object.

*Description of changes:*
Use torch multiprocess instead.

### DIFF
--- a/python/graphstorm/gconstruct/utils.py
+++ b/python/graphstorm/gconstruct/utils.py
@@ -167,7 +167,7 @@ def worker_fn(worker_id, task_queue, res_queue, user_parser):
         while True:
             # If the queue is empty, it will raise the Empty exception.
             i, in_file = task_queue.get_nowait()
-            logging.debug(f"{worker_id} Processing {in_file}")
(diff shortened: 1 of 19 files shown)


python/graphstorm/gconstruct/utils.py
```diff
@@ -167,6 +167,7 @@ def worker_fn(worker_id, task_queue, res_queue, user_parser):
         while True:
             # If the queue is empty, it will raise the Empty exception.
             i, in_file = task_queue.get_nowait()
+            logging.debug("%d Processing %s", worker_id, in_file)
```
zhjwy9343, line 170: Can we control the different log level to avoid output more information?

zhjwy9343 (approved):
LGTM

python/graphstorm/gconstruct/utils.py
classicsong: logging.debug will not print until you use --logging-level debug.

PR #253, opened 2023-06-14
Fix the case when dtype of predict and label does not match in regression tasks.

classicsong:
*Issue #, if available:*
When dtype of predict and label does not match, compute_mse and compute_rmse will crash

*Description of changes:*


zhjwy9343 (approved):
LGTM

Issue #248, opened 2023-06-13
[Bug] Best epoch does not work


zhjwy9343:
- The best epoch output is always 0 in Edge classification/regression and Link Prediction.
- There is no best epoch output for Node classification/regression.

Without this indicator, it is hard to find the best epoch+iteration for the inference. Need to save logs and search for the best evaluation value back to find the best epoch+iteration.

classicsong:
We have some 'Best Iteration' in our printing logs. Do you mean we need to save a meta-file indicating the best path to the best model?

PR #255, opened 2023-06-14
[Bug Fix] Fix best epoch issue

zhjwy9343:
*Issue #248 , if available:*

*Description of changes:*
Return both best iteration and best model path to replace best epoch.

### DIFF
--- a/python/graphstorm/trainer/ep_trainer.py
+++ b/python/graphstorm/trainer/ep_trainer.py
@@ -106,7 +106,6 @@ class GSgnnEdgePredictionTrainer(GSgnnTrainer):
 
         # training loop
         dur = []
-        best_epoch = 0
         num_input_nodes = 0
         forward_time = 0
         back_time = 0
@@ -230,7 +229,8 @@ class GSgnnEdgePredictionTrainer(GSgnnTrainer):
             output = {'best_test_score': self.evaluator.best_test_score,
                        'best_val_score': self.evaluator.best_val_score,
                        'peak_mem_alloc_MB': th.cuda.max_memory_allocated(device) / 1024 / 1024,
-                       'best_epoch': best_epoch}
+                       'best validation iteration': self.evaluator.best_iter_num[self.evaluator.metric[0]],
+                       'best model path': self.get_best_model_path()}
             self.log_params(output)
 
             if save_perf_results_path is not None:
--- a/python/graphstorm/trainer/lp_trainer.py
+++ b/python/graphstorm/trainer/lp_trainer.py
@@ -107,7 +107,6 @@ class GSgnnLinkPredictionTrainer(GSgnnTrainer):
 
         # training loop
         dur = []
-        best_epoch = 0
         num_input_nodes = 0
         total_steps = 0
         early_stop = False # used when early stop is True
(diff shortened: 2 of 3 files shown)


python/graphstorm/trainer/ep_trainer.py
```diff
@@ -230,7 +229,10 @@ def fit(self, train_loader, num_epochs,
             output = {'best_test_score': self.evaluator.best_test_score,
                        'best_val_score': self.evaluator.best_val_score,
                        'peak_mem_alloc_MB': th.cuda.max_memory_allocated(device) / 1024 / 1024,
-                       'best_epoch': best_epoch}
+                       'best validation iteration': \
+                           self.evaluator.best_iter_num[self.evaluator.metric[0]],
+                       'best model path': \
+                           self.get_best_model_path() if save_model_path is not None else None}
```
classicsong, line 235:
```suggestion
                           self.get_best_model_path() if save_model_path is not None else "No model is saved, please set save_model_path"}
```

python/graphstorm/trainer/ep_trainer.py
```diff
@@ -230,7 +229,10 @@ def fit(self, train_loader, num_epochs,
             output = {'best_test_score': self.evaluator.best_test_score,
                        'best_val_score': self.evaluator.best_val_score,
                        'peak_mem_alloc_MB': th.cuda.max_memory_allocated(device) / 1024 / 1024,
-                       'best_epoch': best_epoch}
+                       'best validation iteration': \
+                           self.evaluator.best_iter_num[self.evaluator.metric[0]],
```
classicsong, line 233: You need to check whether no-validation is set.

python/graphstorm/trainer/lp_trainer.py
```diff
@@ -227,7 +226,10 @@ def fit(self, train_loader, num_epochs,
             output = {'best_test_mrr': self.evaluator.best_test_score,
                        'best_val_mrr':self.evaluator.best_val_score,
                        'peak_mem_alloc_MB': th.cuda.max_memory_allocated(device) / 1024 / 1024,
-                       'best_epoch': best_epoch}
+                       'best validation iteration': \
+                           self.evaluator.best_iter_num[self.evaluator.metric[0]],
```
classicsong, line 230: You need to check whether no-validation is set.

python/graphstorm/trainer/np_trainer.py
```diff
@@ -217,7 +217,11 @@ def fit(self, train_loader, num_epochs,
         if self.rank == 0 and self.evaluator is not None:
             output = {'best_test_score': self.evaluator.best_test_score,
                        'best_val_score': self.evaluator.best_val_score,
-                       'peak_mem_alloc_MB': th.cuda.max_memory_allocated(device) / 1024 / 1024}
+                       'peak_mem_alloc_MB': th.cuda.max_memory_allocated(device) / 1024 / 1024,
+                       'best validation iteration': \
+                           self.evaluator.best_iter_num[self.evaluator.metric[0]],
```
classicsong, line 222: You need to check whether no-validation is set.

python/graphstorm/trainer/ep_trainer.py
zhjwy9343: The check has be done by setting up evaluator.

If no-validation is set, there will no evaluator defined. So, this output will be disabled automatically.

python/graphstorm/trainer/lp_trainer.py
zhjwy9343: ditto

python/graphstorm/trainer/np_trainer.py
zhjwy9343: ditto

Issue #249, opened 2023-06-13
Link prediction training fails when there's no test set provided


thvasilo:
During a link prediction training task, where a train and validation set are provided, not providing a test set leads to an error.

Expected behavior: when a user provides either a validation or test set, the evaluation step should only perform model eval with one of those datasets.

Actual behavior: the evaluation step assumes both the validation and test set exist, leading to an error during evaluation.

To reproduce: Run an LP training job on a graph that only includes a train and one of validation/test sets.

Stacktrace:

```
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ /opt/ml/code/graphstorm/python/graphstorm/run/gsgnn_lp/gsgnn_lp.py:158 in    │
│ <module>                                                                     │
│                                                                              │
│   155 │                                                                      │
│   156 │   args = parser.parse_args()                                         │
│   157 │   print(args)                                                        │
│ ❱ 158 │   main(args)                                                         │
│   159                                                                        │
│                                                                              │
│ /opt/ml/code/graphstorm/python/graphstorm/run/gsgnn_lp/gsgnn_lp.py:124 in    │
│ main                                                                         │
│                                                                              │
"│   121 │   │   save_model_path = os.path.join(config.save_embed_path, ""model"" │"
│   122 │   else:                                                              │
│   123 │   │   save_model_path = None                                         │
"│ ❱ 124 │   trainer.fit(train_loader=dataloader, val_loader=val_dataloader,    │"
"│   125 │   │   │   │   test_loader=test_dataloader, num_epochs=config.num_epo │"
"│   126 │   │   │   │   save_model_path=save_model_path,                       │"
"│   127 │   │   │   │   use_mini_batch_infer=config.use_mini_batch_infer,      │"
│                                                                              │
│ /opt/ml/code/graphstorm/python/graphstorm/trainer/lp_trainer.py:194 in fit   │
│                                                                              │
│   191 │   │   │                                                              │
│   192 │   │   │   val_score = None                                           │
│   193 │   │   │   if self.evaluator is not None and self.evaluator.do_eval(t │
"│ ❱ 194 │   │   │   │   val_score = self.eval(model.module, data,              │"
"│   195 │   │   │   │   │   │   │   │   │     val_loader, test_loader, total_s │"
│   196 │   │   │   │   │   │   │   │   │     edge_mask_for_gnn_embeddings)    │
│   197                                                                        │
│                                                                              │
[...]

PR #258, opened 2023-06-15
[Bug fix] Fix SageMaker LP no Prediction Error

zhjwy9343:
*Issue #249 , if available:*

*Description of changes:*
Fix the no prediction error in SageMaker Link Prediction Inference task.


thvasilo (approved):
The changes LGTM. Do we need to apply this restriction at a level lower than the SageMaker scripts? In other words, is a user able to request predictions when using the local mode? Are there any issues there?

PR #267, opened 2023-06-16
Update memory profiling support for all training tasks

classicsong:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/python/graphstorm/utils.py
+++ b/python/graphstorm/utils.py
@@ -180,9 +180,30 @@ class SysTracker:
     def __init__(self):
         self._checkpoints = []
         self._rank = -1
+        self._verbose = False
 
     # This is to create only one instance.
     _instance = None
+
+    def init(self, verbose=False, rank=None):
+        """ Initialize the profiler.
+
+            Set tracking level.
+            TODO: Only have verbose=True or False.
+            Need to support different logging level later.
+
+        Parameter
+        ---------
+        verbose: bool
+            Verbose for print out more info
+        rank: int
+            Current trainer rank
+        """
+        self._verbose = verbose
+        if rank is None:
+            self._rank = get_rank()
+        else:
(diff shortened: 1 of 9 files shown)


python/graphstorm/utils.py
```diff
@@ -311,5 +311,89 @@ def save_profile(self):
...
+                    _mem_byte2gb(mem_stat.used),
+                    _mem_byte2gb(mem_stat.available))
+        if name not in self._checkpoints:
+            self._checkpoints[name] = [mem_info]
+        else:
+            self._checkpoints[name].append(mem_info)
+        if name not in self._runtime:
+            self._runtime[name] = [mem_info]
+        else:
+            self._runtime[name].append(mem_info)
```
zheng-da, line 365: why do you store mem_info in runtime?

python/graphstorm/utils.py
```diff
@@ -213,8 +237,8 @@ def check(self, name):
         if len(self._checkpoints) >= 2 and self._rank == 0:
             checkpoint1 = self._checkpoints[-2]
             checkpoint2 = self._checkpoints[-1]
-            logging.debug("{}: elapsed time: {:.3f}, mem (curr: {:.3f}, peak: {:.3f}, \
-                    shared: {:.3f}, global curr: {:.3f}, global shared: {:.3f}) GB".format(
+            logging.debug("{}: elapsed time: {:.3f}, mem (curr: {:.3f}, shared: {:.3f}, \
+                    peak: {:.3f}, global curr: {:.3f}, global shared: {:.3f}) GB".format(
```
zheng-da, line 241: I think the original is correct. Are you sure this is correct?

classicsong: The original one is correct. Let me revert the change.

zheng-da (approved):

PR #270, opened 2023-06-19
Support feature normalization ops on edge features.

classicsong:
*Issue #, if available:*
#259 

*Description of changes:*


python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -34,20 +34,18 @@
...
-                        GlobalProcessFeatTransform)
+from .transform import do_multiprocess_transform
 from .id_map import NoopMap, IdMap, map_node_ids
 from .utils import (multiprocessing_data_read,
                     update_two_phase_feat_ops, ExtMemArrayMerger,
                     partition_graph)
 
 def prepare_node_data(in_file, feat_ops, read_file):
-    """ Parse node data.
+    """ Prepare node data for feature transform
```
zheng-da, line 44: transform => transformation

python/graphstorm/gconstruct/utils.py
```diff
@@ -378,8 +378,15 @@ def __call__(self, arrs, name):
         if len(arrs) > 1:
             return _merge_arrs(arrs, tensor_path)
         else:
+            # To get the output dtype by accessing the
+            # first element of the arrays (numpy array or HDFArray)
+            # Note: We use arrs[0][0] instead of arrs[0] because
+            #       arrs[0][0] is a transformed data with out_dtype
+            #       while arrs[0] can be a HDFArray and has not
+            #       been cast to out_dtype.
+            dtype = arrs[0][0].dtype
```
zheng-da, line 387: how many places do you need to change it like this?

zheng-da: maybe write a wrapper function?

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -101,6 +99,32 @@ def parse_node_data(in_file, feat_ops, label_ops, node_id_col, read_file):
     node_ids = data[node_id_col] if node_id_col in data else None
     return (node_ids, feat_data)
 
+def prepare_edge_data(in_file, feat_ops, read_file):
+    """ Parse node data.
```
zheng-da, line 103: we need to change the docstring. how is this different from `parse_edge_data`?

zheng-da (approved):

python/graphstorm/gconstruct/utils.py
classicsong: only 2

python/graphstorm/gconstruct/construct_graph.py
classicsong: Updated

python/graphstorm/gconstruct/utils.py
classicsong: I agree, we need a wrapper function. This part of the code is tricky.

PR #272, opened 2023-06-19
Add function 'to_undirected' to graph construction

CongWeilin:
*Issue #, if available:*

Inspired by [ogbn-mag's dgl code](https://github.com/dmlc/dgl/tree/1.0.x/examples/pytorch/ogb/ogbn-mag), we might want to add support to construct undirected graph. For example, Instead of reversing `(paper, cites, paper)` into a new relation like `(paper, rev-cites, paper)`, sometime we just want to make these into undirected edges.

*Description of changes:*

(1) Add a argparser to see if user want to construct an undirected graph

(2) If add_reverse_edges=True but to_undirected=True, we only consider changing edges that has save source and destination node type. 

(3) If add_reverse_edges=False but to_undirected=True, we change all nodes to undirected.


python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -567,9 +567,22 @@ def process_graph(args):
...
+    elif args.to_undirected:
+        edges1 = {}
+        for etype in edges:
+            e = edges[etype]
+            assert isinstance(e, tuple) and len(e) == 2
+            assert isinstance(etype, tuple) and len(etype) == 3
+            edges1[etype] = e
+            edges1[etype[2], etype[1], etype[0]] = (e[1], e[0])
+        edges = edges1
+        sys_tracker.check('To undirected')
```
zheng-da, line 585: is this necessary?

CongWeilin: It depends. Please ignore this pull require if is not necessary.

PR #273, opened 2023-06-20
[Feature] Categorical feature transform allows some input values are None

classicsong:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/python/graphstorm/gconstruct/transform.py
+++ b/python/graphstorm/gconstruct/transform.py
@@ -164,15 +164,21 @@ class CategoricalTransform(TwoPhaseFeatTransform):
             # It will load all data into main memory.
             feats = feats.to_numpy()
 
+        feats[feats == None] = "MAGIC_NONE"
         if self._separator is None:
-            return {self.feat_name: np.unique(feats)}
+            category_keys = np.unique(feats)
         else:
             assert feats.dtype.type is np.str_, \
                     "We can only convert strings to multiple categorical values with separaters."
             vals = []
             for feat in feats:
                 vals.extend(feat.split(self._separator))
-            return {self.feat_name: np.unique(vals)}
+            category_keys = np.unique(vals)
+
+        category_keys = set(category_keys.flatten())
+        if "MAGIC_NONE" in category_keys:
+            category_keys.remove("MAGIC_NONE")
+        return {self.feat_name: np.array(list(category_keys))}
 
     def update_info(self, info):
         """ Store global information for the second phase data processing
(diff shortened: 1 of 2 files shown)


python/graphstorm/gconstruct/transform.py
```diff
@@ -164,15 +164,21 @@ def pre_process(self, feats):
             # It will load all data into main memory.
             feats = feats.to_numpy()
 
+        feats[feats == None] = "MAGIC_NONE"
```
zheng-da, line 167: what does this do?

python/graphstorm/gconstruct/transform.py
```diff
@@ -164,15 +164,21 @@ def pre_process(self, feats):
...
         else:
             assert feats.dtype.type is np.str_, \
                     "We can only convert strings to multiple categorical values with separaters."
             vals = []
             for feat in feats:
                 vals.extend(feat.split(self._separator))
-            return {self.feat_name: np.unique(vals)}
+            category_keys = np.unique(vals)
+
+        category_keys = set(category_keys.flatten())
```
zheng-da, line 178: you don't need to convert it into a set. you only do test once.

python/graphstorm/gconstruct/transform.py
zheng-da: can `None` be the value of the elements in the numpy array?

tests/unit-tests/gconstruct/test_transform.py
```diff
@@ -190,6 +190,8 @@ def test_categorize_transform():
     }
     transform = CategoricalTransform("test1", "test", transform_conf=transform_conf)
     str_ids = [str(i) for i in np.random.randint(0, 10, 1000)]
+    str_ids[0] = None
+    str_ids[-1] = None # allow None data
```
zheng-da, line 194: how do you get this value from the input file?

python/graphstorm/gconstruct/transform.py
classicsong: None is a special object. We can not do np.unique on an array with None value.
Here I try to convert None into MAGIC_NONE, so that we can do np.unique.

python/graphstorm/gconstruct/transform.py
classicsong: I want to remove the key MAGIC_NONE from category_keys.

python/graphstorm/gconstruct/transform.py
classicsong: We don't need this. I have updated the implementation.

python/graphstorm/gconstruct/transform.py
```diff
@@ -164,6 +164,7 @@ def pre_process(self, feats):
             # It will load all data into main memory.
             feats = feats.to_numpy()
 
+        feats = feats[feats is not None]
```
zheng-da, line 167: the idea here is to have `None` as element values, but we don't create a value for `None` in the dict?

classicsong: It should be feats = feats[feats != None]

zheng-da (approved):

zheng-da:
it looks good to me. but do we really need to support `None` in the input data?

zheng-da:
does it represent missing categorical values?

classicsong:
> does it represent missing categorical values?

Yes. If for some rows, there is no categorical value, i.e., empty, we treat it as all zeros in one-hot encoding.

PR #274, opened 2023-06-20
Handle cases when data types of input features do not match the corresponding transform ops

classicsong:
*Issue #, if available:*

*Description of changes:*


python/graphstorm/gconstruct/id_map.py
```diff
@@ -82,7 +82,13 @@ def __init__(self, ids):
...
         if isinstance(ids, HDF5Array):
             ids = ids.to_numpy()
-        self._ids = {id1: i for i, id1 in enumerate(ids)}
+
+        if isinstance(ids[0], int) or np.issubdtype(ids.dtype, np.integer):
+            # node_ids are integer ids
+            self._ids = {id1: i for i, id1 in enumerate(ids)}
+        else:
+            # cast everything else into string
+            self._ids = {str(id1): i for i, id1 in enumerate(ids)}
```
zheng-da, line 91: why do we need to do this?

python/graphstorm/gconstruct/utils.py
```diff
@@ -304,7 +304,9 @@ def _merge_arrs(arrs, tensor_path):
     """
     assert isinstance(arrs, list)
     shape = _get_tot_shape(arrs)
-    dtype = arrs[0].dtype
+    # To get the output dtype by accessing the
+    # first element of the arrays (numpy array or HDFArray)
+    dtype = arrs[0][0].dtype
```
zheng-da, line 309: why do you need to change this?

python/graphstorm/gconstruct/transform.py
```diff
@@ -261,10 +261,15 @@ def pre_process(self, feats):
...
-            "or integers"
+        if feats.dtype not in [np.float64, np.float32, np.float16, np.int64, \
+                           np.int32, np.int16, np.int8]:
+            logging.warning("The feature %s has to be floating points or integers,"
+                            "but get %s. Try to cast it into float32",
+                            self.feat_name, feats.dtype)
+            try:
+                feats = feats.astype(np.float32)
+            except: # pylint: disable=bare-except
+                raise ValueError(f"The feature {self.feat_name} has to be integers or floats.")
```
zheng-da, line 272: are you really converting float64 to float32? it seems to me that this conversion only works if the data is not float or integer.

python/graphstorm/gconstruct/transform.py
```diff
@@ -577,7 +594,7 @@ def parse_feat_ops(confs):
         assert 'feature_col' in feat, \
                 "'feature_col' must be defined in a feature field."
         feat_name = feat['feature_name'] if 'feature_name' in feat else feat['feature_col']
-        out_dtype = _get_output_dtype(feat['out_dtype']) if 'out_dtype' in feat else None
+        out_dtype = _get_output_dtype(feat['out_dtype']) if 'out_dtype' in feat else np.float32
```
zheng-da, line 597: this is to convert everything to float32 by default?
what is my input features are integer? for example, one-hot encoding.

PR #280, opened 2023-06-22
Handle the case when input node id dtype is not integer or string

classicsong:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/python/graphstorm/gconstruct/id_map.py
+++ b/python/graphstorm/gconstruct/id_map.py
@@ -82,7 +82,15 @@ class IdMap:
         # the following operations.
         if isinstance(ids, HDF5Array):
             ids = ids.to_numpy()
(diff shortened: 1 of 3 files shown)


zheng-da (approved):

python/graphstorm/gconstruct/id_map.py
```diff
@@ -100,18 +108,20 @@ def map_id(self, ids):
         tuple of tensors : the tensor of new IDs, the location of the IDs in the input ID tensor.
         """
         for id_ in self._ids:
```
zheng-da, line 110: this loop is to test whether the dtype of keys matches the input array?

classicsong: Yes.

PR #281, opened 2023-06-22
Set default dtype of transformed feature to float32

classicsong:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/python/graphstorm/gconstruct/transform.py
+++ b/python/graphstorm/gconstruct/transform.py
@@ -147,6 +147,9 @@ class GlobalProcessFeatTransform(FeatTransform):
             np.array: processed feature
         """
 
+    def call(self, feats):
+        raise NotImplementedError
+
 class TwoPhaseFeatTransform(FeatTransform):
     """ The base class for two phasefeature transformation.
 
@@ -175,6 +178,9 @@ class TwoPhaseFeatTransform(FeatTransform):
         info:
             Information to be collected
         """
+
+    def call(self, feats):
+        raise NotImplementedError
 
 class CategoricalTransform(TwoPhaseFeatTransform):
     """ Convert the data into categorical values.
@@ -460,7 +466,7 @@ class Tokenizer(FeatTransform):
         self.tokenizer = BertTokenizer.from_pretrained(bert_model)
         self.max_seq_length = max_seq_length
 
-    def __call__(self, strs):
+    def call(self, strs):
         """ Tokenization function.
 
         Parameters
@@ -547,7 +553,7 @@ class Text2BERT(FeatTransform):
                 lm_model = lm_model.to(self.device)
             self.lm_model = lm_model
 
-    def call(self, strs):
+    def call(self, feats):
         """ Compute BERT embeddings of the strings..
 
         Parameters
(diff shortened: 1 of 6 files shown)


python/graphstorm/gconstruct/transform.py
```diff
@@ -577,7 +577,11 @@ def parse_feat_ops(confs):
         assert 'feature_col' in feat, \
                 "'feature_col' must be defined in a feature field."
         feat_name = feat['feature_name'] if 'feature_name' in feat else feat['feature_col']
-        out_dtype = _get_output_dtype(feat['out_dtype']) if 'out_dtype' in feat else None
+
+        # By default the out_dtype is set to float32
+        # FeatTransform sub-classes can ignore out_dtype
+        # or redefine it in its init function.
+        out_dtype = _get_output_dtype(feat['out_dtype']) if 'out_dtype' in feat else np.float32
```
zheng-da, line 584: what about one-hot encoding? we decide to use int8 to store the data.

python/graphstorm/gconstruct/utils.py
```diff
@@ -304,7 +304,13 @@ def _merge_arrs(arrs, tensor_path):
     """
     assert isinstance(arrs, list)
     shape = _get_tot_shape(arrs)
-    dtype = arrs[0].dtype
+    # To get the output dtype by accessing the
+    # first element of the arrays (numpy array or HDFArray)
+    # Note: We use arrs[0][0] instead of arrs[0] because
+    #       arrs[0][0] is a transformed data with out_dtype
+    #       while arrs[0] can be a HDFArray and has not
+    #       been cast to out_dtype.
```
zheng-da, line 312: does it mean that HDFArray has different dtype from the data type in array?

python/graphstorm/gconstruct/transform.py
classicsong: The One-hot encoding transformer can ignore this out_dtype and has its own strategy.

classicsong: The conversion is done inside each transformation.

python/graphstorm/gconstruct/utils.py
classicsong: HDFArray's dtype it is input dtype.

python/graphstorm/gconstruct/transform.py
classicsong: CategoricalTransform just ignore out_dtype.

python/graphstorm/gconstruct/transform.py
```diff
@@ -372,7 +431,7 @@ def after_merge_transform(self, feats):
         feats = np.clip(feats, -1 + self._epsilon, 1 - self._epsilon)
         feats = erfinv(feats)
 
-        feats = _feat_astype(feats, self._out_dtype)
+        self.as_out_dtype(feats)
```
zheng-da, line 434: do you need to call it here?

python/graphstorm/gconstruct/utils.py
zheng-da: i don't understand. what is input dtype?

classicsong: dtype of HDFArray._arr

python/graphstorm/gconstruct/transform.py
classicsong: For RankGaussTransform, yes. 
Its feats is transformed in after_merge_transform(), which is after __call__() call.

zheng-da (approved):

PR #282, opened 2023-06-22
Use cached BERT model in CI-docker image to do testing instead of downloading it

classicsong:
*Issue #, if available:*

*Description of changes:*


prateekdesai04 (approved):
LGTM
Only question - what if there are updates, we keep checking and re-building the image?

classicsong:
> LGTM Only question - what if there are updates, we keep checking and re-building the image?

Can we fix the cache?

PR #298, opened 2023-06-28
Edge classification/regression inference tasks can work with graphs without test masks.

classicsong:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/python/graphstorm/dataloading/dataset.py
+++ b/python/graphstorm/dataloading/dataset.py
@@ -433,6 +433,7 @@ class GSgnnEdgeInferData(GSgnnEdgeData):
         """
         pb = g.get_partition_book()
         test_idxs = {}
+        infer_idxs = {}
         # If eval_etypes is None, we use all edge types.
         if self.eval_etypes is None:
             self._eval_etypes = g.canonical_etypes
@@ -446,14 +447,27 @@ class GSgnnEdgeInferData(GSgnnEdgeData):
                 if dist_sum(len(test_idx)) > 0:
                     test_idxs[canonical_etype] = test_idx
             else:
-                print(f"WARNING: {canonical_etype} does not contains " \
-                      "test_mask, skip testing {canonical_etype}")
+                print(f"NOTE: {canonical_etype} does not contains " \
+                      "test_mask, skip testing {canonical_etype}. \n" \
+                      "We will do inference on the whole edge list.")
+                infer_idx = dgl.distributed.edge_split(
+                    th.full((g.num_edges(canonical_etype),), True, dtype=th.bool),
+                    pb, etype=canonical_etype, force_even=True)
+                infer_idxs[canonical_etype] = infer_idx
+
         self._test_idxs = test_idxs
+        self._infer_idxs = infer_idxs
 
     @property
     def eval_etypes(self):
         """edge type for evaluation"""
         return self._eval_etypes
+
+    @property
+    def infer_idxs(self):
+        """ Set of edges to do inference.
+        """
+        return self._infer_idxs
 
 #### Node classification/regression Task Data ####
 class GSgnnNodeData(GSgnnData):  # pylint: disable=abstract-method
--- a/tools/partition_graph.py
+++ b/tools/partition_graph.py
@@ -61,6 +61,9 @@ if __name__ == '__main__':
     argparser.add_argument("--generate-new-edge-split", type=lambda x:(str(x).lower() in ['true', '1']),
                            default=False, help="If we are splitting the data from scatch we should "
                            + "not do it by default.")
+    argparser.add_argument("--no-split", type=lambda x:(str(x).lower() in ['true', '1']),
+                           default=False, help="Remove the train/valid/test mask."
+                           "Used when testing inference.")
     argparser.add_argument('--train-pct', type=float, default=0.8,
                            help='The pct of train nodes/edges. Should be > 0 and < 1, and only work in '
                                 + 'generating new split')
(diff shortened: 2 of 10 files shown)


python/graphstorm/dataloading/dataset.py
```diff
@@ -446,15 +447,28 @@ def prepare_data(self, g):
                 if dist_sum(len(test_idx)) > 0:
                     test_idxs[canonical_etype] = test_idx
             else:
-                print(f"WARNING: {canonical_etype} does not contains " \
-                      "test_mask, skip testing {canonical_etype}")
+                print(f"NOTE: {canonical_etype} does not contains " \
+                      "test_mask, skip testing {canonical_etype}. \n" \
+                      "We will do inference on the whole edge list.")
+                infer_idx = dgl.distributed.edge_split(
+                    th.full((g.num_edges(canonical_etype),), True, dtype=th.bool),
```
zheng-da, line 454: we should modify DGL to split all edges.

python/graphstorm/dataloading/dataset.py
```diff
@@ -446,15 +447,28 @@ def prepare_data(self, g):
...
         self._test_idxs = test_idxs
+        self._infer_idxs = infer_idxs
 
     @property
     def eval_etypes(self):
         """edge type for evaluation"""
         return self._eval_etypes
 
+    @property
+    def infer_idxs(self):
```
zheng-da, line 467: i don't understand the logic here. how is `infer_idxs` different from `test_idxs`? it seems to me that if test edges are not provided, we compute scores over all edges?

python/graphstorm/dataloading/dataset.py
classicsong: You mean DGL provides an API like edge_split_all?

python/graphstorm/dataloading/dataset.py
classicsong: You are right. If there is no test edges, we compute scores for all edges.

classicsong: But for infer edges, no label is needed.

python/graphstorm/dataloading/dataset.py
zheng-da: maybe something like `edge_split(dgl.ALL, ...)`

python/graphstorm/dataloading/dataset.py
zheng-da: what does infer edges mean? do you mean to compute the scores on all edges? this is link prediction. what do you mean by labels?

classicsong: Edge idxs for inference. This dataset can also be used by edge classificaiton/regression.

zheng-da: i see. i checked `GSgnnNodeInferData`. it doesn't have this method. why is this required for `GSgnnEdgeInferData`?

classicsong: If we want to support node inference without test_mask, we also need to add infer_idxs in GSgnnNodeInferData.
The reason I add infer_idxs instead of re-using test_idx is because, I want to separate this case with the case that has test_mask and will output test scores.

classicsong: I will add inference without test mask support for GSgnnNodeInferData in another PR.

classicsong: @zheng-da Any concerns?

tools/partition_graph.py
```diff
@@ -263,6 +266,25 @@
             raise Exception('There is no predicted edge type to split. Please set the '
                             +'target_etype argument ......')
 
+    if args.no_split:
+        if pred_ntypes is not None:
+            for ntype in pred_ntypes:
+                if 'train_mask' in g.nodes[ntype].data:
+                    del g.nodes[ntype].data['train_mask']
```
zheng-da, line 273: why do you need to delete train_mask? does the train mask exist in the original graph data?

classicsong: The answer is yes for some datasets. So if no_split is set to True, we need to remove these masks.

python/graphstorm/dataloading/dataset.py
```diff
@@ -436,31 +441,52 @@ def prepare_data(self, g):
...
                 test_idx = dgl.distributed.edge_split(
                     g.edges[canonical_etype].data['test_mask'],
                     pb, etype=canonical_etype, force_even=True)
                 # If there are test data globally, we should add them to the dict.
                 if test_idx is not None and dist_sum(len(test_idx)) > 0:
                     test_idxs[canonical_etype] = test_idx
-                elif test_idx is None:
-                    print(f"WARNING: {canonical_etype} does not contains " \
-                            "test data, skip testing {canonical_etype}")
+            elif 'infer_mask' in g.edges[canonical_etype].data:
```
zheng-da, line 458: we are going to support a new type of mask?

classicsong: I think this can provide more flexibility for users to specify a subset of edges to do inference.

python/graphstorm/inference/ep_infer.py
```diff
@@ -70,6 +70,11 @@ def infer(self, loader, save_embed_path, save_prediction_path=None,
             Whether to return all the predictions or the maximum prediction.
         """
         do_eval = self.evaluator is not None
+        if do_eval:
```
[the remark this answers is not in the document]
classicsong: Added.
But we still need to add an assert here. If we only do assert in edge_mini_batch_predict, we may waste time on doing do_full_graph_inference.

zheng-da (approved):

PR #299, opened 2023-06-29
Fix a bug in custom data split.

zheng-da:
*Description of changes:*
The current data split doesn't work if the input node data are stored in multiple files.
https://github.com/awslabs/graphstorm/issues/297

### DIFF
--- a/python/graphstorm/gconstruct/transform.py
+++ b/python/graphstorm/gconstruct/transform.py
@@ -738,7 +738,16 @@ def parse_feat_ops(confs):
             else:
                 raise ValueError('Unknown operation: {}'.format(conf['name']))
         ops.append(transform)
-    return ops
+
+    two_phase_feat_ops = []
+    after_merge_feat_ops = {}
+    for op in ops:
+        if isinstance(op, TwoPhaseFeatTransform):
+            two_phase_feat_ops.append(op)
+        if isinstance(op, GlobalProcessFeatTransform):
+            after_merge_feat_ops[op.feat_name] = op
+
+    return ops, two_phase_feat_ops, after_merge_feat_ops
 
 def preprocess_features(data, ops):
     """ Pre-process the data with the specified operations.
--- a/python/graphstorm/gconstruct/transform.py
+++ b/python/graphstorm/gconstruct/transform.py
@@ -922,15 +922,7 @@ class CustomLabelProcessor:
         -------
         dict of Tensors : it contains the labels as well as training/val/test splits.
         """
-        if self.col_name in data:
-            label = data[self.col_name]
-            num_samples = len(label)
-        else:
-            assert len(data) > 0, "The edge data is empty."
-            label = None
-            for val in data.values():
-                num_samples = len(val)
-                break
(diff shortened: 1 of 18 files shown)


zhjwy9343 (approved):
LGTM

python/graphstorm/gconstruct/transform.py
```diff
@@ -853,13 +855,14 @@ class CustomLabelProcessor:
...
-    def __init__(self, col_name, label_name, task_type,
+    def __init__(self, col_name, label_name, id_col, task_type,
                  train_idx=None, val_idx=None, test_idx=None):
+        self._id_col = id_col
         self._col_name = col_name
         self._label_name = label_name
-        self._train_idx = train_idx
-        self._val_idx = val_idx
-        self._test_idx = test_idx
+        self._train_idx = set(train_idx.tolist()) if train_idx is not None else None
```
classicsong, line 863: This is a list of strings?

python/graphstorm/gconstruct/transform.py
```diff
@@ -1113,8 +1113,9 @@ def parse_label_ops(confs, is_node):
         val_idx = read_index_json(custom_split['valid']) if 'valid' in custom_split else None
         test_idx = read_index_json(custom_split['test']) if 'test' in custom_split else None
         label_col = label_conf['label_col'] if 'label_col' in label_conf else None
-        return [CustomLabelProcessor(label_col, label_col, task_type,
-                                     train_idx, val_idx, test_idx)]
+        assert "node_id_col" in confs, "Custom data split only works for nodes."
```
classicsong, line 1116: We need to update the README to introduce "node_id_col".

python/graphstorm/gconstruct/transform.py
```diff
@@ -874,28 +877,30 @@ def label_name(self):
         """
         return self._label_name
 
-    def data_split(self, num_samples):
+    def data_split(self, ids):
```
classicsong, line 880: Can we add a unit test for CustomLabelProcessor?
We only have end2end test now.

python/graphstorm/gconstruct/transform.py
zheng-da: these are node IDs. they can be strings or integers.

python/graphstorm/gconstruct/transform.py
zheng-da: what do you mean? we have node_id_col all the time

python/graphstorm/gconstruct/transform.py
zheng-da: we have unit test for custom data split.

zheng-da: https://github.com/awslabs/graphstorm/blob/main/tests/unit-tests/gconstruct/test_construct_graph.py#L599-L697

zheng-da: we have many unit tests for custom data split.

classicsong (approved):

Issue #265, opened 2023-06-16
[Feature request] Add CPU enablement


DominikaJedynak:
It would be valuable to be able to run GraphStorm in both GPU and CPU environments.

bartekkuncer:
@classicsong as https://github.com/awslabs/graphstorm/pull/300 have been merged, maybe it is the right time to close this issue?

PR #300, opened 2023-06-29
[New feature] CPU enablement

DominikaJedynak:
Addresses https://github.com/awslabs/graphstorm/issues/265

This PR adds the possibility to run GraphStorm on machines with CPU only, giving users more flexibility in choosing their environment settings.
Solution for CPU (similarly like for GPU) uses torch DistributedDataParallel.


classicsong:
@DominikaJedynak
Thank you for your contribution. :)
Do you have a timeline for this PR?

DominikaJedynak:
> Do you have a timeline for this PR?

Hi @classicsong 
I don't have a specific timeline but I am currently working on it and testing it out in a distributed setting.

zheng-da:
@DominikaJedynak do you have a plan to finish the PR in a week?

DominikaJedynak:
@zheng-da Yes, I have. I am in contact with @classicsong regarding that.

zhjwy9343:
@DominikaJedynak This is James. I have made a few modifications of your PR to fix these pylint errors.

Now it can pass the basic CI tests, but failed at the end-to-end tests. Could you help to fix these GPU related issues?

I would be happy to help to facilitate this new feature to be included.

python/graphstorm/run/launch.py
```diff
@@ -419,6 +419,8 @@ def construct_dgl_server_env_vars(
 
     """
     server_env_vars_template = (
+        #"NVIDIA_VISIBLE_DEVICES={NVIDIA_VISIBLE_DEVICES} " temporary flags for testing
```
DominikaJedynak, line 422: Should I remove those flags or might their be useful for later eg. for testing?

zheng-da:
did we run the tests in CI on CPU instances?

examples/customized_models/HGT/hgt_nc.py
```diff
@@ -195,7 +195,10 @@ def forward(self, blocks, node_feats, edge_feats, labels, input_nodes):
             if self.adapt_ws[ntype] is None:
                 n_id = self.node_dict[ntype]
                 emb_id = self.ntype_id_map[n_id]
-                embeding = self.ntype_embed(torch.Tensor([emb_id]).long().to('cuda'))
+                emb_tensor = torch.Tensor([emb_id]).long()
+                if torch.cuda.is_available():
```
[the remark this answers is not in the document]
DominikaJedynak: Another approach that I can think of is to check whether self.dev_id >= 0 but according to my measurement it takes similar time as cuda.is_available() (tenth of a millisecond) but is more prone to bugs I think, as it relies on implementation.

DominikaJedynak:
> did we run the tests in CI on CPU instances?

I was talking with @zhjwy9343 that I was hoping to run exactly the same tests on CI CPU-only machines as are now for GPU, but I am not aware if anything was added. For now, I ran some tests by myself on EC2 instances, I can prepare the results.

classicsong:
> > did we run the tests in CI on CPU instances?
> 
> I was talking with @zhjwy9343 that I was hoping to run exactly the same tests on CI CPU-only machines as are now for GPU, but I am not aware if anything was added. For now, I ran some tests by myself on EC2 instances, I can prepare the results.

We will help setup a CPU CI pipeline.

python/graphstorm/run/launch.py
zhjwy9343: I would suggest to remove them and find other way for testing.

python/graphstorm/run/gsgnn_np/gsgnn_np.py
```diff
@@ -86,7 +86,10 @@ def main(config_args):
     if trainer.rank == 0:
         tracker.log_params(config.__dict__)
     trainer.setup_task_tracker(tracker)
-    device = 'cuda:%d' % trainer.dev_id
+    if th.cuda.is_available():
```
classicsong, line 89: We can have a setup_device() function in graphstorm.utils.py as:

```
def setup_device(local_rank):
    """ Setup device

    Parameters
    -----------
    local_rank: int
        Locak rank, same as gpu id.
    """
    if th.cuda.is_available():
        device = 'cuda:%d' % local_rank
        th.cuda.set_device(device)
    else:
        device="cpu"

    return device
```

classicsong: Here is an example: https://github.com/awslabs/graphstorm/pull/337/files#diff-fb641736e08124342fd53ef3ba61dd132f6e25851fe261250d536573a23fc8e7

zhjwy9343:
@DominikaJedynak Could you also help to fix this log output?

`[Rank 0] dist_inference: Layer [0]finishes [{iter_l}] iterations.`

Should the {iter_l} be a number?

python/graphstorm/run/gsgnn_np/gsgnn_np.py
DominikaJedynak: I can change it like that if you prefer so. But do we want to change it here if it is already done in another incoming PR?

classicsong: Let's do it in this PR. This can make codes simpler.
That PR is a draft.

python/graphstorm/trainer/lp_trainer.py
```diff
@@ -210,11 +212,14 @@ def fit(self, train_loader, num_epochs,
...
-        print("Peak Mem alloc: {:.4f} MB".format(th.cuda.max_memory_allocated(device) / 1024 /1024))
+        print("Peak GPU Mem alloc: {:.4f} MB".format(th.cuda.max_memory_allocated(device) /
+                                                     1024 / 1024))
+        print("RAM memory used: {:.4f} MB".format(psutil.virtual_memory().used / 1024 / 1024))
         if self.rank == 0 and self.evaluator is not None:
             output = {'best_test_mrr': self.evaluator.best_test_score,
                        'best_val_mrr':self.evaluator.best_val_score,
-                       'peak_mem_alloc_MB': th.cuda.max_memory_allocated(device) / 1024 / 1024,
+                       'peak_GPU_mem_alloc_MB': th.cuda.max_memory_allocated(device) / 1024 / 1024,
+                       'RAM_mem_used': psutil.virtual_memory().used / 1024 / 1024,
```
classicsong, line 222: Is this message useful?
Can we get peak memory usage during training?

examples/customized_models/HGT/hgt_nc.py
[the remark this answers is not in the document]
DominikaJedynak: Changed

.github/workflow_scripts/e2e_cpu_check.sh
```diff
@@ -6,7 +6,6 @@ set -ex
 sh ./tests/end2end-tests/setup.sh
 sh ./tests/end2end-tests/create_data.sh
 sh ./tests/end2end-tests/custom-gnn/run_test.sh
-bash ./tests/end2end-tests/graphstorm-nc/test.sh
-bash ./tests/end2end-tests/graphstorm-lp/test.sh
-bash ./tests/end2end-tests/graphstorm-ec/test.sh
```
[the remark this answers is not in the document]
DominikaJedynak: @zheng-da was there some particular reason for that?

examples/customized_models/HGT/hgt_nc.py
```diff
@@ -227,7 +231,10 @@ def predict(self, blocks, node_feats, _, input_nodes, return_proba):
             if self.adapt_ws[ntype] is None:
                 n_id = self.node_dict[ntype]
                 emb_id = self.ntype_id_map[n_id]
-                embeding = self.ntype_embed(torch.Tensor([emb_id]).long().to('cuda'))
+                emb_tensor = torch.Tensor([emb_id]).long()
```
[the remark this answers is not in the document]
DominikaJedynak: Changed as well

python/graphstorm/trainer/lp_trainer.py
DominikaJedynak: I thought that printing used memory would be useful, but I could change it to eg. `resource.getrusage(resource.RUSAGE_SELF).ru_maxrss / 1024` 
from https://docs.python.org/3/library/resource.html#resource-usage which should return peak memory. Do you think this would fit better?

classicsong (approved):
In general LGTM.

python/graphstorm/trainer/ep_trainer.py
```diff
@@ -213,11 +214,14 @@ def fit(self, train_loader, num_epochs,
                 break
 
         rt_profiler.save_profile()
-        print("Peak Mem alloc: {:.4f} MB".format(th.cuda.max_memory_allocated(device) / 1024 /1024))
+        print("Peak GPU Mem alloc: {:.4f} MB".format(th.cuda.max_memory_allocated(device) /
```
classicsong, line 217: We need to check whether cuda device is avaliable before we collect GPU Mem alloc info.

DominikaJedynak: I have changed it here so that only one info is printed depending on if cuda is used or not. But for the log_params I left both parameters because max GPU memory will be just 0 in case of calculations on CPU and all the runs will have consistent  parameters.

zhjwy9343 (approved):
LGTM

PR #302, opened 2023-06-30
Optimize OMP setting

classicsong:
*Issue #, if available:*
Distributed KVServer always use OMP=1, set to a larger number to leverage multi-core CPU.

*Description of changes:*

### DIFF
--- a/python/graphstorm/run/launch.py
+++ b/python/graphstorm/run/launch.py
@@ -932,8 +932,8 @@ def check_input_arguments(args):
         # where the launch script runs.
         # The total number of CPU cores consumed
         # by trainers will be half of total cpu cores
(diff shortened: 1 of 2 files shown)


python/graphstorm/run/launch.py
```diff
@@ -913,9 +912,6 @@ def check_input_arguments(args):
     assert (
         args.num_servers is not None and args.num_servers > 0
     ), "--num-servers must be a positive number."
-    assert (
-        args.num_server_threads > 0
-    ), "--num-server-threads must be a positive number."
```
zheng-da, line 918: why is this assert not always true?

zheng-da (approved):

PR #305, opened 2023-07-03
Updating CUDA  docker image tag to 11.3.1

prateekdesai04:
*Description of changes:*

11.3.0 does not seem to be supported anymore.
https://gitlab.com/nvidia/container-images/cuda/blob/master/doc/supported-tags.md


zhjwy9343 (approved):
LGTM. Let's make this minor version change first. And later we need to test the cuda 12.* docker image to avoid this issue for a couple of years.

PR #311, opened 2023-07-04
Add gradient clipping by norm to lp_trainer

GrayKing:
Add gradient clipping to ensure the stability of training process. 


python/graphstorm/trainer/lp_trainer.py
```diff
@@ -55,7 +55,8 @@ def fit(self, train_loader, num_epochs,
             save_model_frequency=None,
             save_perf_results_path=None,
             edge_mask_for_gnn_embeddings='train_mask',
-            freeze_input_layer_epochs=0):
+            freeze_input_layer_epochs=0,
+            max_grad_norm=None):
```
zheng-da, line 59: please update the docstring. also, can you show an example of how this argument is used? basically, what are the typical values of this argument?

PR #315, opened 2023-07-06
Remove models as we have already uploaded the model

classicsong:
*Issue #, if available:*

*Description of changes:*


zhjwy9343 (commented):
For Inference, we need to have the same function. Because the inference task in nature is also a SageMaker Training Job, and need to upload many things to S3.

classicsong:
> For Inference, we need to have the same function. Because the inference task in nature is also a SageMaker Training Job, and need to upload many things to S3.

Updated

zhjwy9343 (approved):
LGTM. 

I tested it with the MAG-large graph. It works. SageMaker just upload its metrics, no more model artifacts uploaded.

PR #317, opened 2023-07-10
add checks to verify the correctness of the input arguments

zheng-da:
*Description of changes:*
We need to verify the correctness of the input arguments as early as possible. This is important when we deal with large datasets.
https://github.com/awslabs/graphstorm/issues/20

### DIFF
--- a/python/graphstorm/config/argument.py
+++ b/python/graphstorm/config/argument.py
@@ -259,9 +259,9 @@ class GSConfig:
             _ = self.remove_target_edge_type
 
         # LM module
-        _ = self.lm_infer_batch_size
-        _ = self.freeze_lm_encoder_epochs
-        _ = self.node_lm_configs
+        if self.node_lm_configs:
+            _ = self.lm_infer_batch_size
+            _ = self.freeze_lm_encoder_epochs
 
         # Model architecture
         _ = self.model_encoder_type
--- a/python/graphstorm/config/argument.py
+++ b/python/graphstorm/config/argument.py
@@ -233,8 +233,6 @@ class GSConfig:
         _ = self.eval_frequency
         _ = self.no_validation
         _ = self.save_prediction_path
-        _ = self.eval_negative_sampler
-        _ = self.num_negative_edges_eval
         _ = self.eval_etype
         if self.task_type is not None:
             _ = self.eval_metric
@@ -310,6 +308,8 @@ class GSConfig:
             _ = self.lp_edge_weight_for_loss
             _ = self.lp_loss_func
             _ = self.num_negative_edges
+            _ = self.eval_negative_sampler
+            _ = self.num_negative_edges_eval
 
     def _turn_off_gradient_checkpoint(self, reason):
         """Turn off `gradient_checkpoint` flags in `node_lm_configs`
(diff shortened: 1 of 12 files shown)


python/graphstorm/config/argument.py
```diff
@@ -203,6 +203,113 @@ def override_arguments(self, cmd_args):
...
+        # Data
+        _ = self.node_feat_name
+        _ = self.decoder_edge_feat
+
+        # Evaluation
+        _ = self.eval_fanout
+        _ = self.use_mini_batch_infer
+        _ = self.eval_batch_size
+        _ = self.eval_frequency
+        _ = self.no_validation
```
classicsong, line 234: We only need to check these evaluation related arguments when
`if self.no_validation is False`

python/graphstorm/config/argument.py
```diff
@@ -203,6 +203,113 @@ def override_arguments(self, cmd_args):
...
+            _ = self.early_stop_rounds
+            _ = self.early_stop_strategy
+            _ = self.use_early_stop
+            _ = self.wd_l2norm
+            _ = self.train_negative_sampler
+            _ = self.train_etype
+            _ = self.remove_target_edge_type
+
+        # LM module
+        _ = self.lm_infer_batch_size
```
classicsong, line 264: We only need to check lm related module when 
`if self.node_lm_configs is not None`

python/graphstorm/config/argument.py
```diff
@@ -1393,8 +1501,6 @@ def exclude_training_targets(self):
     def gamma(self):
         """ Gamma for DistMult
         """
-        assert self.lp_decoder_type == BUILTIN_LP_DISTMULT_DECODER, \
```
classicsong, line 1396: We do not use gamma with dot-product decoder

tests/unit-tests/test_config.py
```diff
@@ -159,201 +159,6 @@ def test_load_basic_info():
         check_failure(config, "ip_config")
         check_failure(config, "part_config")
 
-def create_gnn_config(tmp_path, file_name):
```
classicsong, line 162: Why remove these tests?

zheng-da: these are duplicated code.

python/graphstorm/config/argument.py
zheng-da: even in the inference script?

python/graphstorm/config/argument.py
zheng-da: we need to make a choice on how to verify the value when training/inference scripts start. either we don't verify it, or we only verify it when the lp decoder is distmult, or we don't enforce the check in the argument.

python/graphstorm/config/argument.py
classicsong: Inference also accepts no_validation.

python/graphstorm/config/argument.py
classicsong: My opinion is we can verify it when the lp decoder is distmult.

zheng-da: that is very specific verification. we can do this for an argument, but my concern is that this will explode the difficulty of verifying arguments in the future.

classicsong (approved):

PR #320, opened 2023-07-11
Reduce memory consumption in graph construction pipeline.

zheng-da:
*Description of changes:*
This PR uses a wrapper for numpy memmapped arrays and release the numpy array after it is created. This can reduce memory consumption at least from the perspective of virtual memory.

### DIFF
--- a/python/graphstorm/gconstruct/utils.py
+++ b/python/graphstorm/gconstruct/utils.py
@@ -302,6 +302,29 @@ def _get_arrs_out_dtype(arrs):
     """
     return arrs[0][0].dtype
 
+class ExtNumpyWrapper:
+    def __init__(self, arr_path, shape, dtype):
+        self._arr_path = arr_path
+        self._shape = shape
+        self._dtype = dtype
+        self._arr = None
+
+    @property
+    def dtype(self):
+        return self._dtype
+
+    @property
+    def shape(self):
+        return self._shape
+
+    def __getitem__(self, idx):
+        if self._arr is None:
+            self._arr = np.memmap(self._arr_path, self._dtype, mode="r", shape=self._shape)
+        return self._arr[idx]
+
+    def cleanup(self):
+        self._arr = None
+
 def _merge_arrs(arrs, tensor_path):
     """ Merge the arrays.
 
@@ -329,7 +352,7 @@ def _merge_arrs(arrs, tensor_path):
         for arr in arrs:
             out_arr[row_idx:(row_idx + arr.shape[0])] = arr[:]
             row_idx += arr.shape[0]
-        return out_arr
+        return ExtNumpyWrapper(tensor_path, out_arr.shape, out_arr.dtype)
     elif isinstance(arrs[0], HDF5Array):
         arrs = [arr.to_numpy() for arr in arrs]
         return np.concatenate(arrs)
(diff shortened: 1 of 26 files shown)


python/graphstorm/gconstruct/utils.py
```diff
@@ -302,6 +302,224 @@ def _get_arrs_out_dtype(arrs):
...
+    @property
+    def shape(self):
+        """ The shape of the array.
+        """
+        return self._shape
+
+    def astype(self, dtype):
+        """ Return an array with converted data type.
+        """
+        arr = copy.copy(self)
```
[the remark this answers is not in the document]
classicsong: or self._dtype = dtype?

python/graphstorm/gconstruct/utils.py
```diff
@@ -329,8 +547,8 @@ def _merge_arrs(arrs, tensor_path):
         for arr in arrs:
             out_arr[row_idx:(row_idx + arr.shape[0])] = arr[:]
             row_idx += arr.shape[0]
-        return out_arr
-    elif isinstance(arrs[0], HDF5Array):
+        return ExtNumpyWrapper(tensor_path, out_arr.shape, out_arr.dtype)
```
classicsong, line 550: Why not pass out_arr to ExtNumpyWrapper?

python/graphstorm/gconstruct/utils.py
```diff
@@ -329,8 +547,8 @@ def _merge_arrs(arrs, tensor_path):
...
+    elif isinstance(arrs[0], ExtMemArrayWrapper):
         arrs = [arr.to_numpy() for arr in arrs]
         return np.concatenate(arrs)
```
classicsong, line 553: Does np.concatenate support concatenating memmaped arrays?

classicsong (commented):
I don't understand why this PR can save memory consumption.

python/graphstorm/gconstruct/utils.py
```diff
@@ -302,6 +302,224 @@ def _get_arrs_out_dtype(arrs):
...
+            idx = idx.numpy()
+        # If the idx are sorted.
+        if np.all(idx[1:] - idx[:-1] > 0):
+            arr = self._arr[idx]
+        else:
+            # There are two cases here: 1) there are duplicated IDs,
+            # 2) the IDs are not sorted. Unique can return unique
+            # IDs in the ascending order that meets the requirement of
+            # HDF5 indexing.
+            uniq_ids, reverse_idx = np.unique(idx, return_inverse=True)
```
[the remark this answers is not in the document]
zheng-da: it probably will consume a lot of memory. we should fix it in other PRs.

python/graphstorm/gconstruct/utils.py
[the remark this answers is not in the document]
zheng-da: what is the semantic of `astype`? I think this method shouldn't change the existing array.

python/graphstorm/gconstruct/utils.py
zheng-da: we want to release the original numpy array and open a new one when it is used.

python/graphstorm/gconstruct/utils.py
zheng-da: we have called to_numpy. it should return normal numpy arrays.

python/graphstorm/gconstruct/utils.py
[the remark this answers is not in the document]
classicsong: Return a new array with the input dtype.
My question is will copy.copy make a copy of self._arr?

python/graphstorm/gconstruct/utils.py
classicsong: I c

classicsong (approved):
LGTM

PR #322, opened 2023-07-11
Fix Device or resource busy: '/opt/ml/model' bug when running SageMaker training and inference

classicsong:
*Issue #, if available:*
Commit 32c2baa995e331db1bfff574ddd0de2dae20816e tries to remove the trained model or generated node embeddings under /opt/ml/model/ and /opt/ml/checkpoints/ to avoid uploading the trained model or node embeddings using SageMaker default model saving mechanism which is slow. However, this may trigger the "Device or resource busy: '/opt/ml/model'" error.

*Description of changes:*
In this PR, we change the training/inference output path to /opt/ml/gsgnn_model/ and /opt/ml/infer_output respectively to avoid the above problem.


prateekdesai04 (approved):
LGTM

PR #326, opened 2023-07-16
Fix JSON input format.

zheng-da:
*Description of changes:*
A field of a JSON block may contain a list. We need to convert it to numpy array.


python/graphstorm/gconstruct/file_io.py
```diff
@@ -131,9 +141,13 @@ def read_data_json(data_file, data_fields):
         for key in data_fields:
             assert key in record, \
                     f"The data field {key} does not exist in the record {record} of {data_file}."
-            data[key].append(record[key])
+            record1 = np.array(record[key]) if isinstance(record[key], list) else record[key]
+            data[key].append(record1)
     for key in data:
-        data[key] = np.array(data[key])
+        if isinstance(data[key][0], np.ndarray):
```
classicsong, line 147: Is it possible that Process 0 has max_len 10 while Process 1 has max_len 12. for example.

python/graphstorm/gconstruct/file_io.py
```diff
@@ -102,6 +102,16 @@ def write_data_csv(data, data_file, delimiter=','):
     data_frame = pd.DataFrame(data)
     data_frame.to_csv(data_file, index=True, sep=delimiter)
 
+def _pad_stack(arrs):
+    max_len = max(len(arr) for arr in arrs)
+    dtype = arrs[0].dtype
+    for i, arr in enumerate(arrs):
```
classicsong, line 108: Will this implementation more efficient?
```
new_arrs = np.zeros((len(arr), max_len), dtype=dtype)
for i, arr in enumerate(arrs):
   new_arrs[i][[:len(arr)] = arr
```
So we do not need to allocate many small numpy arrays.

python/graphstorm/gconstruct/file_io.py
zheng-da: this is a concern. if this happens, we should return an error.

zheng-da: I'll submit another PR to handle errors in the worker processes.

classicsong (approved):

Issue #283, opened 2023-06-22
Modify the dataset/data loaders to support semi-supervised node-level tasks


wangz10:
Make use of un-labeled nodes in the graph to improve performances of node-level tasks.

PR #327, opened 2023-07-17
Support semi-supervised node-level tasks

wangz10:
*Issue #, if available:* #283

*Description of changes:*
- Implemented a new data loader `GSgnnNodeSemiSupDataLoader` to enable semi-supervised training for node-level tasks. 
- Implemented semi-supervised training for GLEM
- semi-supervised training for GLEM performs better when LM and GNN are pre-trained, hence added `num_pretrain_epochs` to specify number of pretraining epochs for GLEM


python/graphstorm/config/argument.py
```diff
@@ -1085,6 +1086,17 @@ def label_field(self):
             "Must provide the feature name of labels through label_field"
         return self._label_field
 
+    @property
+    def semi_supervised(self):
```
zheng-da, line 1090: do you think if this naming is a bit confusing? we also say that GCN model training is semi-supervised.

python/graphstorm/model/node_glem.py
```diff
@@ -161,7 +166,8 @@ def toggle(self, part='lm'):
             raise ValueError(f"Unknown model part: {part}")
 
     def forward(self, blocks, node_feats, edge_feats, labels, input_nodes, use_gnn=True,
-                no_pl=False):
+                no_pl=False, blocks_u=None, node_feats_u=None, edge_feats_u=None,
+                input_nodes_u=None):
```
zheng-da, line 170: can you explain the new input arguments?

python/graphstorm/run/gsgnn_np/gsgnn_np.py
```diff
@@ -87,10 +89,19 @@ def main(config_args):
         tracker.log_params(config.__dict__)
     trainer.setup_task_tracker(tracker)
     device = 'cuda:%d' % trainer.dev_id
-    dataloader = GSgnnNodeDataLoader(train_data, train_data.train_idxs, fanout=config.fanout,
-                                     batch_size=config.batch_size, device=device, train_task=True)
     val_dataloader = None
     test_dataloader = None
+    if config.semi_supervised:
+        # treat val and test sets as unlabeled node sets
+        unlabeled_idxs = combine_idxs(train_data.val_idxs, train_data.test_idxs)
```
zheng-da, line 96: why not use the other unlabeled nodes? in many datasets, there are many unlabeled nodes that are not in the training, validation and test datasets.

python/graphstorm/trainer/glem_np_trainer.py
```diff
@@ -116,7 +117,7 @@ def fit(self, train_loader, num_epochs,
                 self._model.lm.unfreeze_input_encoder()
                 no_pl = False
 
-            use_gnn = not self._model.em_order_gnn_first
+            use_gnn = self._model.em_order_gnn_first
```
zheng-da, line 120: why do you change the order?

python/graphstorm/model/node_glem.py
```diff
@@ -175,12 +181,23 @@ def forward(self, blocks, node_feats, edge_feats, labels, input_nodes, use_gnn=T
         no_pl : bool
             If True, do not calculate pseudo likelihood, use MLE loss only
         """
-        if use_gnn:
-            total_loss = self.forward_gnn(blocks, node_feats, edge_feats, labels, input_nodes,
-                                          no_pl=no_pl)
+        if blocks_u is None or no_pl:
```
zheng-da, line 184: do you need to test both?

python/graphstorm/trainer/glem_np_trainer.py
```diff
@@ -160,25 +161,42 @@ def _fit_one_epoch(self, use_gnn, model, g, data, train_loader,
...
+                # supervised setting, no unlabeled data
+                input_nodes, input_feats, blocks, lbl = _prepare_batch(*batch)
+                input_nodes_u = input_feats_u = blocks_u = None
+            profiler.record('train_sample')
+            total_steps += 1
+            batch_tic = time.time()
             # Run forward function to compute loss:
-            loss = model(blocks, input_feats, None, lbl, input_nodes, use_gnn=use_gnn, no_pl=no_pl)
+            loss = model(blocks, input_feats, None, lbl, input_nodes, use_gnn=use_gnn,
+                         no_pl=no_pl or (epoch < model.module.num_pretrain_epochs),
```
zheng-da, line 197: should this argument be associated with the model? not with the trainer?

python/graphstorm/trainer/glem_np_trainer.py
wangz10: The previous implementation was wrong: conflicts with what the flag suggests.

python/graphstorm/run/gsgnn_np/gsgnn_np.py
wangz10: I don't think there is any unlabeled nodes in public graphs (arxiv, products, MovieLens). All target nodes . Do you think we should use all nodes not in `train_idxs` instead?

python/graphstorm/model/node_glem.py
wangz10: will do

python/graphstorm/trainer/glem_np_trainer.py
wangz10: It can be associated with either. But this is specific to GLEM, currently GLEM trainer doesn't have configurable arguments. So I think it's clearer if all options for training GLEM are grouped to the model like this: 

```yml
  lmgnn:
    lm_train_nodes: 10
    lm_infer_batch_size: 32
    freeze_lm_encoder_epochs: 0
    model_encoder_type: rgcn
    fanout: "4"
    num_layers: 1
    hidden_size: 128
    use_mini_batch_infer: true
    training_method:
      name: glem
      kwargs:
        em_order_gnn_first: false
        inference_using_gnn: true
        pl_weight: 0.5
        num_pretrain_epochs: 1
```

python/graphstorm/config/argument.py
wangz10: Yes, what name do you suggest?

python/graphstorm/model/node_glem.py
wangz10: Yes. `no_pl` is used to force ignoring unlabeled data for pretraining LM or GNN. This is required even when unlabeled data are provided.

python/graphstorm/run/gsgnn_np/gsgnn_np.py
zheng-da: yes, there are many datasets with nodes that are not included in training/val/test sets.

python/graphstorm/trainer/glem_np_trainer.py
zheng-da: i think so. what i mean is whether we should attach `num_pretrain_epochs` to the model object or the trainer object.

python/graphstorm/trainer/glem_np_trainer.py
```diff
@@ -45,6 +46,7 @@ def __init__(self, model, rank, topk_model_to_save=1):
         assert isinstance(model, GSgnnNodeModelInterface) and isinstance(model, GLEM), \
                 "The input model is not a GLEM node model. Please implement GLEM."
         self.early_stop = False # used when early stop is True
+        self.num_pretrain_epochs = model.num_pretrain_epochs
```
zheng-da, line 49: i wonder why `num_pretrain_epochs` is not an argument of `GLEMNodePredictionTrainer`?

zheng-da (approved):

PR #331, opened 2023-07-19
Enable standalone mode.

zheng-da:
*Description of changes:*
This PR allows users to run the training/inference scripts without using the launch script. this will make it much easier for people to develop and debug models in GraphStorm.


classicsong (approved):
LGTM
We should add a new section in wiki talking about the standalone mode.

PR #332, opened 2023-07-19
Fix wrong graph partition config filename bug

classicsong:
*Issue #, if available:*
There is a logic error in [python/graphstorm/sagemaker/utils.py:L251](https://github.com/awslabs/graphstorm/pull/332/files#diff-3bd5dedc90be6c0508ab83d0a94ed14e7dfd5857db7eeb9eb32ea4aed46fdeb5). The graph_config will always be "metadata.json", which actually is no longer used. The actual graph_config name should be f"{graph_name}.json"

*Description of changes:*


zhjwy9343 (approved):
LGTM

PR #337, opened 2023-07-21
[Draft] Enable nccl backend for pure language model tuning tasks.

classicsong:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/examples/customized_models/HGT/hgt_nc.py
+++ b/examples/customized_models/HGT/hgt_nc.py
@@ -16,6 +16,7 @@ from graphstorm.dataloading import GSgnnNodeTrainData, GSgnnNodeInferData
 from graphstorm.dataloading import GSgnnNodeDataLoader
 from graphstorm.eval import GSgnnAccEvaluator
 from graphstorm.tracker import GSSageMakerTaskTracker
+from graphstorm.utils import setup_device
 
 from dgl.nn.functional import edge_softmax
 
@@ -259,6 +260,7 @@ class HGT(gsmodel.GSgnnNodeModelBase):
 def main(args):
     gs.initialize(ip_config=args.ip_config, backend="gloo")
     config = GSConfig(args)
+    device = setup_device(config.local_rank)
 
     # Process node_feat_field to define GraphStorm dataset
     node_feat_fields = {}
@@ -305,8 +307,7 @@ def main(args):
 
     # Create a trainer for the node classification task.
     trainer = GSgnnNodePredictionTrainer(model, gs.get_rank(), topk_model_to_save=config.topk_model_to_save)
-    trainer.setup_cuda(dev_id=gs.get_rank())
-    device = 'cuda:%d' % trainer.dev_id
+    trainer.setup_device(device=device)
 
     # Define the GraphStorm train dataloader
     dataloader = GSgnnNodeDataLoader(train_data, train_data.train_idxs, fanout=config.fanout,
@@ -346,7 +347,7 @@ def main(args):
     # After training, get the best model from the trainer.
     best_model_path = trainer.get_best_model_path()
     model.restore_model(best_model_path)
-    
+
     # Create a dataset for inference.
     infer_data = GSgnnNodeInferData(config.graph_name, config.part_config,
                                     eval_ntypes=config.target_ntype,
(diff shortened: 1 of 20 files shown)


python/graphstorm/eval/utils.py
```diff
@@ -233,7 +233,9 @@ def calc_ranking(pos_score, neg_score):
     _, indices = th.sort(scores, dim=1, descending=True)
     indices = th.nonzero(indices == 0)
     rankings = indices[:, 1].view(-1) + 1
-    rankings = rankings.cpu().detach()
+    rankings = rankings.detach()
+    if th.distributed.get_backend() == "gloo":
+        rankings = rankings.cpu() # Save GPU memory
```
isratnisa, line 238: Does this mean for NCCL backed we want to store `rankings` on GPU? Is that necessary?

isratnisa (commented):
1. If I understood correctly, when NCCL backend is used, we want to calculate score and ranking on GPU, otherwise on CPU. Previously, we would copy related tensors to CPU regardless the backed. Should we be concerned about the extra GPU memory consumption? 
2. This PR is more about memory efficiency with NCCL/GLOO backend rather than enabling/disabling NCCL. If so, we should change the PR title?

python/graphstorm/eval/utils.py
classicsong: We need to do all2all using nccl, so we have to save rankings on GPU. This may be a limitation of nccl.

python/graphstorm/eval/utils.py
```diff
@@ -315,11 +318,10 @@ def broadcast_data(rank, world_size, data_tensor):
         dtype=data_tensor.dtype,
         device=device) for size in data_size]
     data_tensors = [data_tensor for _ in data_size]
-    if th.distributed.get_backend() == "gloo":
+    if get_backend() == "gloo":
         alltoallv_cpu(rank, world_size, gather_list, data_tensors)
-    else: #th.distributed.get_backend() == "nccl"
+    else: #get_backend() == "nccl"
         alltoallv_nccl(rank, world_size, gather_list, data_tensors)
```
isratnisa, line 325: We don't need to define alltoallv_nccl. We can directly use Torch's alltoall for nccl backend. You can take a look at my distDGL's PR https://github.com/dmlc/dgl/pull/5929/files

python/graphstorm/model/edge_decoder.py
```diff
@@ -602,8 +603,15 @@ def calc_test_scores(self, emb, pos_neg_tuple, neg_sample_type, device):
             assert len(neg_score.shape) == 2
             neg_scores.append(neg_score)
 
-        neg_scores = th.cat(neg_scores, dim=-1).detach().cpu()
-        pos_scores = pos_scores.detach().cpu()
+        neg_scores = th.cat(neg_scores, dim=-1).detach()
+        # gloo with cpu will consume less GPU memory
+        neg_scores = neg_scores.cpu() \
+            if is_distributed() and get_backend() == "gloo" \
```
isratnisa, line 609: Since we are checking is_distributed inside get_backend, we can remove is_distributed.

python/graphstorm/eval/utils.py
isratnisa: By default, its always stored on CPU right?

python/graphstorm/eval/utils.py
classicsong: This is helpful.

python/graphstorm/eval/utils.py
classicsong: The rankings are stored in GPU if the backend is nccl. If the backend is gloo, the rankings are stored in CPU.

python/graphstorm/model/edge_decoder.py
classicsong: In get_backend, it is an assertion. we can not move is_distributed.

zheng-da (approved):

python/graphstorm/eval/utils.py
```diff
@@ -315,11 +318,10 @@ def broadcast_data(rank, world_size, data_tensor):
...
         device=device) for size in data_size]
     data_tensors = [data_tensor for _ in data_size]
-    if th.distributed.get_backend() == "gloo":
+    if get_backend() == "gloo":
         alltoallv_cpu(rank, world_size, gather_list, data_tensors)
-    else: #th.distributed.get_backend() == "nccl"
-        alltoallv_nccl(rank, world_size, gather_list, data_tensors)
-
+    else: #get_backend() == "nccl"
+        th.distributed.all_to_all(gather_list, data_tensors)
```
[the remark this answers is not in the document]
classicsong: Once cpu supports alltoallv, we can wrap this up with CPU together.

PR #338, opened 2023-07-21
Add multi-target node training support

CongWeilin:
*Issue #, if available:*

*Description of changes:*

Previous GraphStorm implementation assume only a single target node type is used to compute loss and gradient. In this version, we the loss and gradient can be computed on multiple node types.

However, for evaluation, we still only evaluate the model performance on a single node type (the node type with the largest number of nodes for now). To better support multiple target node type evaluation, we might need to re-design our evaluator to better store the best score and its associated model.


python/graphstorm/model/node_gnn.py
```diff
@@ -110,15 +110,15 @@ def forward(self, blocks, node_feats, _, labels, input_nodes=None):
...
-        logits = self.decoder(emb)
-        pred_loss = self.loss_func(logits, labels)
-
+        target_ntypes = list(labels.keys())
+        # compute loss for each node type and aggregate per node type loss
+        pred_loss = 0
+        for target_ntype in target_ntypes:
+            assert target_ntype in encode_embs
+            emb = encode_embs[target_ntype]
+            ntype_labels = labels[target_ntype]
```
zheng-da, line 119: can you check if target_ntype also exists in labels as well?

python/graphstorm/model/node_gnn.py
```diff
@@ -110,15 +110,15 @@ def forward(self, blocks, node_feats, _, labels, input_nodes=None):
...
-        emb = encode_embs[target_ntype]
-        labels = labels[target_ntype]
-        logits = self.decoder(emb)
-        pred_loss = self.loss_func(logits, labels)
-
+        target_ntypes = list(labels.keys())
+        # compute loss for each node type and aggregate per node type loss
+        pred_loss = 0
+        for target_ntype in target_ntypes:
+            assert target_ntype in encode_embs
```
zheng-da, line 117: can you print errror messages if the assert fails?

python/graphstorm/model/node_gnn.py
CongWeilin: added to new commit

python/graphstorm/model/node_gnn.py
CongWeilin: added to new commit

python/graphstorm/model/node_gnn.py
```diff
@@ -169,15 +171,15 @@ def node_mini_batch_gnn_predict(model, loader, return_proba=True, return_label=F
...
 
     if return_label:
         assert data.labels is not None, \
             "Return label is required, but the label field is not provided whem" \
             "initlaizing the inference dataset."
 
-    embs = []
-    labels = []
+    embs = {ntype: [] for ntype in g.ntypes}
+    labels = {ntype: [] for ntype in g.ntypes}
```
zheng-da, line 182: is it a problem that the dict has all node types?

zheng-da: instead of `target_ntypes`

zheng-da:
can you add an integration test to trigger the case?

python/graphstorm/model/node_gnn.py
CongWeilin: I updated this in the latest commit 582bb64e0703f6d5490529443bac56913777115f by 

- first create an empty `embs` as dictionary
- adding per node type `emb` to `embs` only if it is returned by the data loader

Since data loader only return node types in `target_ntypes`, it should be fine right now in this latest commit.

python/graphstorm/model/node_gnn.py
```diff
@@ -187,18 +189,46 @@ def node_mini_batch_gnn_predict(model, loader, return_proba=True, return_label=F
...
+                if ntype in labels:
+                    labels[ntype].append(ntype_labels.cpu())
+                else:
+                    labels[ntype] = [ntype_labels.cpu()]
+            if isinstance(pred, dict):
+                for ntype, ntype_pred in pred.items():
+                    if ntype in preds:
+                        preds[ntype].append(ntype_pred.cpu())
+                    else:
+                        preds[ntype] = [ntype_pred.cpu()]
```
zheng-da, line 202: i see this is replicated everywhere. do you mind creating a function for this operation.

def _add(list_dict, ntype, tensor):
    if ntype in list_dict:
        list_dict[ntype].append(tensor.cpu())
    else:
        list_dict[ntype] = [tensor.cpu()]

python/graphstorm/model/node_gnn.py
```diff
@@ -187,18 +189,46 @@ def node_mini_batch_gnn_predict(model, loader, return_proba=True, return_label=F
...
             blocks = [block.to(device) for block in blocks]
             pred, emb = model.predict(blocks, input_feats, None, input_nodes, return_proba)
-            preds.append(pred.cpu())
-            embs.append(emb.cpu())
+            for ntype, ntype_labels in data.get_labels(seeds).items():
+                if ntype in labels:
+                    labels[ntype].append(ntype_labels.cpu())
+                else:
+                    labels[ntype] = [ntype_labels.cpu()]
+            if isinstance(pred, dict):
```
zheng-da, line 197: is `pred` always a dict?

python/graphstorm/model/node_gnn.py
```diff
@@ -187,18 +189,46 @@ def node_mini_batch_gnn_predict(model, loader, return_proba=True, return_label=F
...
+                    else:
+                        preds[ntype] = [ntype_pred.cpu()]
+            else: # in case model (e.g., llm encoder) only output a tensor without ntype
+                assert len(labels) == 1
+                ntype = list(labels.keys())[0]
+                if ntype in preds:
+                    preds[ntype].append(pred.cpu())
+                else:
+                    preds[ntype] = [pred.cpu()]
+            if isinstance(pred, dict):
```
zheng-da, line 210: i think this if condition should be merged to the previous one.

tests/unit-tests/data_utils.py
```diff
@@ -115,6 +115,95 @@ def generate_dummy_hetero_graph(size='tiny', gen_mask=True):
...
+        hetero_graph.nodes[target_ntype[1]].data['val_mask'] = node_val_mask2
+        hetero_graph.nodes[target_ntype[1]].data['test_mask'] = node_test_mask2
+
+        hetero_graph.edges[target_etype[0]].data['train_mask'] = edge_train_mask
+        hetero_graph.edges[target_etype[0]].data['val_mask'] = edge_val_mask
+        hetero_graph.edges[target_etype[0]].data['test_mask'] = edge_test_mask
+
+        hetero_graph.edges[target_etype[1]].data['train_mask'] = edge_train_mask2
+        hetero_graph.edges[target_etype[1]].data['val_mask'] = edge_val_mask2
+        hetero_graph.edges[target_etype[1]].data['test_mask'] = edge_test_mask2
```
zheng-da, line 203: why target_ntype graphs have target_etype?

zheng-da: can we extend the existing generate_dummy_hetero_graph function?

CongWeilin: > why target_ntype graphs have target_etype?

I was following the original `generate_dummy_hetero_graph` function to make it close to the original function as much as possible

> can we extend the existing generate_dummy_hetero_graph function?

We cannot do this because in another unit-test [here](https://github.com/awslabs/graphstorm/blob/de978745ffad7557ca9f1effde836a8c060da27d/tests/unit-tests/test_dataloading.py#L279) we check whether another node type can fetch labels.

python/graphstorm/trainer/np_trainer.py
```diff
@@ -278,6 +278,17 @@ def eval(self, model, val_loader, test_loader, use_mini_batch_infer, total_steps
                 test_label = None
             sys_tracker.check('after_test_score')
         sys_tracker.check('predict')
+
+        # select node type with largest number of nodes as target evaluation node type
```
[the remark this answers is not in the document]
CongWeilin: sure, l will work on this. Shall we limit `len(eval_node_type)==1` for this PR? 

We might want to add other supports in the future PRs. For example, we might need to consider different target node types have different evaluation metric in the future.

python/graphstorm/model/node_gnn.py
CongWeilin: Yes. Our function `node_mini_batch_gnn_predict` is only used in evaluation, and its output is set to a dictionary by default in this PR. 

For single task node case, then it is a dictionary with length as 1.

zheng-da: i try to understand if we need this if condition. if it's always a dict, why do we need to check its type.

python/graphstorm/trainer/np_trainer.py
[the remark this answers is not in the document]
zheng-da: sure. we can assume that there is only one eval node type.

python/graphstorm/model/node_gnn.py
CongWeilin: Oh, I see your point. 

(1) If the prediction is the output of `gnn_encoder` or `decoder`, it is aways a dictionary
(2) However, there is a special case then prediction is the output of `llm_coder` (e.g., one of the test case in `end2end-tests/graphstorm-nc/mgpu_test.sh`), then the output is just a tensor. 

Maybe we keep this `isinstance` check for now for safe. We can remove it after we make sure all modules return is dictionary.

python/graphstorm/model/node_gnn.py
CongWeilin: Done. Please refer to please refer to function `append_to_dict(from_dict, to_dict):`

python/graphstorm/model/node_gnn.py
```diff
@@ -187,18 +197,24 @@ def node_mini_batch_gnn_predict(model, loader, return_proba=True, return_label=F
...
+                append_to_dict(pred, preds)
+                append_to_dict(emb, embs)
+            else: # in case model (e.g., llm encoder) only output a tensor without ntype
+                assert len(labels) == 1
+                ntype = list(labels.keys())[0]
+                append_to_dict({ntype: pred}, preds)
+                append_to_dict({ntype: emb}, embs)
 
-            if return_label:
```
zheng-da, line 193: it seems you don't check `return_label` any more.

zheng-da (approved):

PR #339, opened 2023-07-21
CI Infra for CPU Support

prateekdesai04:
*Description of changes:*
Building CI infrastructure for CPU support.
Uses the same infrastructure as done by lint check.

### DIFF
--- a/.github/workflow_scripts/e2e_cpu_check.sh
+++ b/.github/workflow_scripts/e2e_cpu_check.sh
@@ -5,7 +5,6 @@ set -ex
 
 sh ./tests/end2end-tests/setup.sh
 sh ./tests/end2end-tests/create_data.sh
-sh ./tests/end2end-tests/tools/test_mem_est.sh
 sh ./tests/end2end-tests/custom-gnn/run_test.sh
 bash ./tests/end2end-tests/graphstorm-nc/test.sh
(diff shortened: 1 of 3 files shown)


.github/workflow_scripts/e2e_cpu_check.sh
```diff
@@ -0,0 +1,13 @@
+# Move to parent directory
+cd ../../
+
+set -ex
+
+sh ./tests/end2end-tests/setup.sh
+sh ./tests/end2end-tests/create_data.sh
+sh ./tests/end2end-tests/tools/test_mem_est.sh
```
classicsong, line 8: This is not needed.

classicsong (approved):
Overall LGTM

.github/workflow_scripts/e2e_cpu_check.sh
prateekdesai04: we would need _create_data.sh_ for partitioning

classicsong: I mean we do not need `sh ./tests/end2end-tests/tools/test_mem_est.sh`

prateekdesai04: Done

PR #340, opened 2023-07-22
Handle errors in the worker processes in the graph construction pipeline

zheng-da:
*Description of changes:*
If a worker process fails, the entire graph construction pipeline will hang. If a worker fails, we should fail the entire graph construction pipeline immediately.

### DIFF
--- a/python/graphstorm/gconstruct/utils.py
+++ b/python/graphstorm/gconstruct/utils.py
@@ -183,7 +183,7 @@ def worker_fn(worker_id, task_queue, res_queue, user_parser):
             gc.collect()
     except queue.Empty:
         pass
-    except Exception as e:
+    except Exception as e:  # pylint: disable=broad-exception-caught
         print(e, flush=True)
         res_queue.put((i, None))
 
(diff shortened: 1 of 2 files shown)


python/graphstorm/gconstruct/utils.py
```diff
@@ -182,6 +183,9 @@ def worker_fn(worker_id, task_queue, res_queue, user_parser):
             gc.collect()
     except queue.Empty:
         pass
+    except Exception as e:  # pylint: disable=broad-exception-caught
+        print(e, flush=True)
```
classicsong, line 187: use logging.error?

python/graphstorm/gconstruct/utils.py
```diff
@@ -246,6 +250,8 @@ def multiprocessing_data_read(in_files, num_processes, user_parser):
         return_dict = {}
         while len(return_dict) < num_files:
             file_idx, vals= res_queue.get()
+            if vals is None:
+                raise RuntimeError("One of the worker processes fails. Stop data processing.")
```
classicsong, line 254: It seems file_idx set in L188 is not used.
BTW, why not raise the RuntimeError in L188?

zheng-da: the main idea is that we need to capture all exceptions in the worker processes and stop the master process whenever an error happens in the worker process.

classicsong (approved):
LGTM

PR #341, opened 2023-07-22
Handle empty edges in graph construction pipelines

zheng-da:
*Description of changes:*
If the end point nodes of all edges don't exist, it will return empty edge list. we need to handle empty edge list.


python/graphstorm/gconstruct/id_map.py
```diff
@@ -184,26 +186,28 @@ def map_node_ids(src_ids, dst_ids, edge_type, node_id_map, skip_nonexist_edges):
...
-        bool_mask[orig_locs] = False
+        if len(orig_locs) > 0:
+            bool_mask[orig_locs] = False
         if skip_nonexist_edges:
             logging.warning("source nodes of %s do not exist: %s",
                             src_type, str(src_ids[bool_mask]))
         else:
             raise ValueError(f"source nodes of {src_type} do not exist: {src_ids[bool_mask]}")
-        dst_ids = dst_ids[orig_locs]
+        dst_ids = dst_ids[orig_locs] if len(orig_locs) > 0 else np.array([], dtype=dst_ids.dtype)
```
classicsong, line 196: If src_ids is empty, we make dst_ids also empty here?

python/graphstorm/gconstruct/id_map.py
```diff
@@ -184,26 +186,28 @@ def map_node_ids(src_ids, dst_ids, edge_type, node_id_map, skip_nonexist_edges):
     # If some of the source nodes don't exist in the node set.
     if len(orig_locs) != len(src_ids):
         bool_mask = np.ones(len(src_ids), dtype=bool)
-        bool_mask[orig_locs] = False
+        if len(orig_locs) > 0:
```
classicsong, line 189: Can you add some doc string about how map_node_ids handle conner cases like none of src_ids exists in the map.

python/graphstorm/gconstruct/id_map.py
zheng-da: yes, src_ids and dst_ids should have the same length.

classicsong (approved):
LGTM

PR #348, opened 2023-07-26
[Feature] Add SageMaker CPU docker build code

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
Add SageMaker CPU docker build file and change build shell script to allow users specify CPU.


docker/sagemaker/Dockerfile.sm_cpu
```diff
@@ -0,0 +1,40 @@
+# Docker file for building a docker image for running GraphStorm code on Amazon SageMaker
+# Note: Distributed graph partition will use another docker image which will come soon.
+ARG PYTHON_VERSION=3.9
+
+FROM 763104351884.dkr.ecr.us-east-1.amazonaws.com/pytorch-training:1.13.1-cpu-py39-ubuntu20.04-sagemaker
```
thvasilo, line 5: If the only thing that is changing between the CPU  and GPU Dockerfiles is the source image we should use a build argument to avoid code duplication:

```suggestion
ARG PYTHON_VERSION=3.9
ARG DEVICE=gpu

FROM 763104351884.dkr.ecr.us-east-1.amazonaws.com/pytorch-training:1.13.1-${DEVICE}-py39-ubuntu20.04-sagemaker
```

Then in the build script we can have

```
if [[ $IMAGE_TYPE = "GPU" ]]; then
    docker build --build-arg="DEVICE=gpu" -f $GSF_HOME"docker/sagemaker/Dockerfile.sm" . -t $DOCKER_FULLNAME
elif [[ $IMAGE_TYPE = "CPU" ]]; then
    docker build --build-arg="DEVICE=cpu" -f $GSF_HOME"docker/sagemaker/Dockerfile.sm" . -t $DOCKER_FULLNAME 
else [...]
```

thvasilo (changes requested):
Would rather we use a build argument vs. duplicating the Dockerfile.

docker/sagemaker/Dockerfile.sm_cpu
zhjwy9343: > If the only thing that is changing between the CPU and GPU Dockerfiles is the source image we should use a build argument to avoid code duplication:
> 
> Then in the build script we can have
> 
> ```
> if [[ $IMAGE_TYPE = "GPU" ]]; then
>     docker build --build-arg="DEVICE=gpu" -f $GSF_HOME"docker/sagemaker/Dockerfile.sm" . -t $DOCKER_FULLNAME
> elif [[ $IMAGE_TYPE = "CPU" ]]; then
>     docker build --build-arg="DEVICE=cpu" -f $GSF_HOME"docker/sagemaker/Dockerfile.sm" . -t $DOCKER_FULLNAME 
> else [...]
> ```
> 
> Or if we use lowercase for the IMAGE_TYPE we can just do:
> 
> ```
> docker build --build-arg="DEVICE=${IMAGE_TYPE}" -f $GSF_HOME"docker/sagemaker/Dockerfile.sm" . -t $DOCKER_FULLNAME
> ```
> 
> and avoid the if clause.
> 
> See https://docs.docker.com/build/guide/build-args/

Another dependency is DGL. So, I need to test DGL installation to check if DGL-cu116 also works for CPU-only SageMaker instance.

zhjwy9343: Need to use a cpu supported DGL, any idea of how to use argument to set different DGL versions like the below two options:
# DGL CPU
RUN pip3 install dgl==1.0.3 -f https://data.dgl.ai/wheels-internal/repo.html
# DGL GPU
RUN pip3 install dgl==1.0.4+cu117 -f https://data.dgl.ai/wheels/cu117/repo.html

thvasilo (approved):
LGTM with the build stage depending on the provided build argument.

PR #350, opened 2023-07-26
[SageMaker] Allow passing of SageMaker Estimator arguments during job…

### LINKED ISSUE #309: [Bug] SageMaker-based training and inference does not support --volume-size argument
The SageMaker training and inference instances have 30GB disk volume by default. 

But, for large graphs, e.g., MAG-large, one partition will be much larger than 30GB. And will cause "lack of disk space" error when download file from S3.

Therefore, need to set the SageMaker --volume-size argument and pass it in the Training and Inference launch scripts.

thvasilo:
… launch

*Issue #, if available:*
Fixes #309 

*Description of changes:*
Allow forwarding of SageMaker Estimator arguments during job launch.

To achieve this we create a common parser for all launch scripts, that includes a new argument `sm-estimator-parameters`. End users can provide a space-separated list of SageMaker Estimator arguments here, in the format
`arg_name1=arg_val1 arg_name2=arg_val2`.

We use space as the entry separator, and the values themselves must not contain extra spaces. 

So a list of strings needs to be passed as `subnets=['subnet-123456','subnet-567799']` and not `subnets=['subnet-123456', 'subnet-567799']`.
We note this detail in the README. 

We parse the values as Python literals allowing them to interpreted with the expected type (str, int, list(str), bool) 

Tested by running SageMaker jobs where we pass e.g. `--sm-estimator-parameters "max_run=3600 volume_size=100 encrypt_inter_container_traffic=True subnets=['subnet-123456','subnet-567799']"`

### DIFF
--- a/sagemaker/launch/launch_gconstruct.py
+++ b/sagemaker/launch/launch_gconstruct.py
@@ -23,11 +23,11 @@ from sagemaker.processing import (ScriptProcessor,
                                   ProcessingOutput)
 import sagemaker
 
-from common_parser import get_common_parser
+from common_parser import get_common_parser, parse_estimator_kwargs
 
 INSTANCE_TYPE = "ml.m5.12xlarge"
 
-def run_job(input_args, image, unknowargs):
+def run_job(input_args, image, unknownargs):
     """ Run job using SageMaker ScriptProcessor
 
     Parameters
@@ -36,7 +36,7 @@ def run_job(input_args, image, unknowargs):
         Input arguments
     image: str
         ECR image uri
-    unknowargs: dict
+    unknownargs: dict
         GraphStorm graph construction parameters
     """
     sm_task_name = input_args.task_name # SageMaker task name
@@ -48,7 +48,6 @@ def run_job(input_args, image, unknowargs):
     output_graph_s3 = input_args.output_graph_s3
     graph_name = input_args.graph_name # Inference graph name
     graph_config_file = input_args.graph_config_file # graph config file
-    volume_size_in_gb = input_args.volume_size_in_gb # instance disk size
 
     boto_session = boto3.session.Session(region_name=region)
     sagemaker_client = boto_session.client(service_name="sagemaker", region_name=region)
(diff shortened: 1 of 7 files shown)


sagemaker/launch/launch_gconstruct.py
```diff
@@ -95,44 +95,38 @@ def run_job(input_args, image, unknowargs):
...
 
 if __name__ == "__main__":
-    arg_parser = parse_args()
+    arg_parser = get_gconstruct_parser()
     args, unknownargs = arg_parser.parse_known_args()
     print(args)
 
     gconstruct_image = args.image_url
 
+    if not args.instance_type:
```
classicsong, line 129: In common_parser L23 we force users to provide an instance_type. So this if statement will always be False.

sagemaker/launch/launch_infer.py
```diff
@@ -71,7 +71,7 @@ def run_job(input_args, image, unknowargs):
 
     container_image_uri = image
 
-    prefix = "script-mode-container"
+    prefix = f"graphstorm-train-{sm_task_name}"
```
classicsong, line 74:
```suggestion
    prefix = f"graphstorm-infer-{sm_task_name}"
```

sagemaker/launch/launch_infer.py
```diff
@@ -125,61 +125,58 @@ def run_job(input_args, image, unknowargs):
...
     return parser
 
 if __name__ == "__main__":
-    arg_parser = parse_args()
+    arg_parser = get_inference_parser()
     args, unknownargs = arg_parser.parse_known_args()
     print(args)
 
     infer_image = args.image_url
+    if not args.instance_type:
```
classicsong, line 178: Same here.

sagemaker/launch/launch_infer.py
```diff
@@ -125,61 +125,58 @@ def run_job(input_args, image, unknowargs):
               {"Key":"GraphStorm_Task", "Value":"Inference"}],
```
classicsong, line 125: missing **estimator_kwargs as in launch_partition.py

sagemaker/launch/launch_train.py
```diff
@@ -107,57 +112,49 @@ def run_job(input_args, image, unknowargs):
...
 
 if __name__ == "__main__":
-    arg_parser = parse_args()
+    arg_parser = get_train_parser()
     args, unknownargs = arg_parser.parse_known_args()
     print(args)
 
     train_image = args.image_url
 
+    if not args.instance_type:
```
classicsong, line 157: Same here.

sagemaker/launch/launch_gconstruct.py
thvasilo: Right, I'll remove the `required` from the common parser to allow different default instances for each task.

thvasilo:
@classicsong addressed all PR comments. Let me know if there's anything else you'd like to see.

classicsong (approved):
LGTM

Issue #87, opened 2023-04-12
Support NCCL distributed backend in GraphStorm


classicsong:
Currently, GraphStorm's evaluation code can only run with gloo. We need to support NCCL in the future.

Here are codes can break NCCL:
https://github.com/awslabs/graphstorm/blob/bdddfde5556069c7267d61b7e2aa37be32197efd/python/graphstorm/dataloading/utils.py#L40-L43

PR #383, opened 2023-08-21
[Performance] NCCL workaround

isratnisa:
This PR provides a workaround of Issue https://github.com/awslabs/graphstorm/issues/87 

NCCL backend is not supported for utilizing learnable embeddings on nodes. It has a dependency on distDGL (PR https://github.com/dmlc/dgl/pull/5929). This PR removes the dependency by suggesting gloo backed for those cases.


python/graphstorm/model/embed.py
```diff
@@ -179,6 +179,19 @@ def __init__(self,
...
+
+        # NCCL backend is not supported for utilizing learnable embeddings on nodes. It has a
+        # dependency on distDGL (PR https://github.com/dmlc/dgl/pull/5929).
+        # TODO (Israt): Add NCCL support to distDGL.
+        if th.distributed.is_initialized() and th.distributed.get_backend() == "nccl":
+            if self.use_node_embeddings:
+                raise NotImplementedError('NCCL backend is not supported for utilizing \
+                    node embeddings. Please use gloo backend.')
+            for ntype in g.ntypes:
+                if not feat_size[ntype]:
```
classicsong, line 191:
```suggestion
                if feat_size[ntype] == 0:
```
This is more easy to understand

python/graphstorm/model/embed.py
```diff
@@ -179,6 +179,19 @@ def __init__(self,
         self.dropout = nn.Dropout(dropout)
         self.use_node_embeddings = use_node_embeddings
         self.activation = activation
+
+        # NCCL backend is not supported for utilizing learnable embeddings on nodes. It has a
+        # dependency on distDGL (PR https://github.com/dmlc/dgl/pull/5929).
+        # TODO (Israt): Add NCCL support to distDGL.
+        if th.distributed.is_initialized() and th.distributed.get_backend() == "nccl":
```
classicsong, line 186: Please use: https://github.com/awslabs/graphstorm/blob/b199d281e0de6e8bf25509507d788fb57a7a1f3f/python/graphstorm/utils.py#L43-L45

classicsong: We should also add a get_backend in graphstrom.utils.

python/graphstorm/utils.py
```diff
@@ -45,21 +45,26 @@ def is_distributed():
     """
     return th.distributed.is_initialized()
 
+
+def get_backend():
+    """ Get the backend of a process group.
+    """
+    return th.distributed.get_backend()
```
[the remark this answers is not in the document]
classicsong: We can add an assertion here as:
```
assert is_distributed(), "get_backend() is valid only when is_distributed() is True."
```

classicsong (approved):
LGTM

PR #386, opened 2023-08-22
Allow using different LM model tokenizer for built-in datasets.

classicsong:
*Issue #, if available:*

*Description of changes:*


zhjwy9343 (changes requested):
In our YAML files, there are several arguments using bert_. Should we change them to lm_ too?
- bert_infer_bs
- bert_models
- bert_tune_lr
In the gconstruct JSON, we also use the following bert_ arguments, should we change them? 
- bert_model

classicsong:
> In our YAML files, there are several arguments using bert_. Should we change them to lm_ too?
> 
> * bert_infer_bs
> * bert_models
> * bert_tune_lr
>   In the gconstruct JSON, we also use the following bert_ arguments, should we change them?
> * bert_model

Let's do it in another PR.
This is a breaking change. May be it is better to and more arguments.

zhjwy9343 (approved):
LGTM

PR #390, opened 2023-08-25
Using logging to print messages

zheng-da:
*Description of changes:*
Using logging to avoid printing too many messages. Users can change the logging level to print out more information for debugging.


python/graphstorm/config/argument.py
```diff
@@ -457,6 +459,15 @@ def verbose(self):
 
         return False
 
+    @property
+    def logging_level(self):
```
classicsong, line 463: Shall we add a --logging-level to allow user set logging level through arugment.

python/graphstorm/dataloading/dataset.py
```diff
@@ -640,8 +642,8 @@ def prepare_data(self, g):
                 if dist_sum(len(test_idx)) > 0:
                     test_idxs[ntype] = test_idx
 
-        print('part {}, train: {}, val: {}, test: {}'.format(get_rank(), num_train,
-                                                             num_val, num_test))
+        logging.debug('part %d, train: %d, val: %d, test: %d',
```
classicsong, line 645: I think the size of train/val/test is commonly used by users as it helps them to understand how large the dataset is.
I would suggest we use logging.info()

python/graphstorm/dataloading/dataset.py
```diff
@@ -378,8 +380,8 @@ def prepare_data(self, g):
                 # If there are test data globally, we should add them to the dict.
                 if dist_sum(len(test_idx)) > 0:
                     test_idxs[canonical_etype] = test_idx
-        print('part {}, train: {}, val: {}, test: {}'.format(get_rank(), num_train,
-                                                             num_val, num_test))
+        logging.debug('part %d, train: %d, val: %d, test: %d',
```
classicsong, line 383: I think the size of train/val/test is commonly used by users as it helps them to understand how large the dataset is.
I would suggest we use logging.info()

python/graphstorm/model/gnn.py
```diff
@@ -582,8 +583,8 @@ def save_model(self, model_path):
                            get_rank(),
                            get_world_size())
         if get_rank() == 0:
-            print('successfully save the model to ' + model_path)
-            print('Time on save model {}'.format(time.time() - start_save_t))
+            logging.debug('successfully save the model to %s', model_path)
```
classicsong, line 586: Shall we use logging.info?
This tells users where does the model save to.

python/graphstorm/model/lm_embed.py
```diff
@@ -45,7 +47,7 @@ def update_bert_cache(g, lm_models_info, lm_models, lm_emb_cache, lm_infer_batch
         lm_model.eval()
         for ntype in lm_ntypes:
             if get_rank() == 0:
-                print('compute bert embedding on node {}'.format(ntype))
+                logging.debug('compute bert embedding on node %d', ntype)
```
classicsong, line 50: I would sugguest we use logging.info here. Otherwise, the system will stuck in computing bert embeddings without printing any information.

python/graphstorm/utils.py
```diff
@@ -365,7 +392,7 @@ def save_profile(self):
             profile_path = os.path.join(self._profile_path, f"{self._rank}.csv")
             data_frame = pd.DataFrame(runtime)
             data_frame.to_csv(profile_path, float_format='%.3f', index=False)
-            print(f"save profiling in {profile_path}")
+            logging.debug("save profiling in %s", profile_path)
```
classicsong, line 395: Let's make it logging.info, so users can directly know where the profile data are stored.

python/graphstorm/config/argument.py
zheng-da: i did

classicsong (approved):
LGTM

PR #401, opened 2023-08-30
Update GSgnnNodeModelInterface interface

classicsong:
*Issue #, if available:*

*Description of changes:*
Both predictions and embed_dict are either tensors or dictionaries of tensors


python/graphstorm/model/node_gnn.py
```diff
@@ -200,13 +203,22 @@ def append_to_dict(from_dict, to_dict):
...
+            # pred can be a Tensor or a dict of Tensor
+            # emb can be a Tensor or a dict of Tensor
             if isinstance(pred, dict):
                 append_to_dict(pred, preds)
+            else:
+                assert len(seeds) == 1, \
+                    f"Expect prediction results of multiple node types {label.keys()}" \
+                    f"But only get results of one node type"
+                ntype = list(seeds.keys())[0]
+                append_to_dict({ntype: pred}, preds)
```
zheng-da, line 216: is this possible that `emb` is a dict while `pred` is a tensor?

classicsong: We just assume, the return value of model.predict can be a tensor or a dict and we turn them into dicts.

zheng-da:
do we need tests to trigger the code?

classicsong:
> do we need tests to trigger the code?

Added

zheng-da (approved):

PR #402, opened 2023-08-30
[SageMaker][Docker] Clone more recent DGL code, remove cache files, use Buildkit.

thvasilo:
*Issue #, if available:*

*Description of changes:*

* We clone a more recent version of DGL on to the image, that includes changes needed to run partitioning on SageMaker.
* We clean up temporary cache files after pip/apt installations, as listed in [Dockerfile best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get). This reduces image size by ~300MB.
* Enable [buildkit](https://docs.docker.com/build/buildkit/) if available for builds. This will avoid pulling both CPU and GPU images among other optimizations.


classicsong (approved):
LGTM

PR #406, opened 2023-08-31
[SageMaker] Remove downloading of config file during partition launch.

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Small change to remove the downloading of the config file during partitioning launch.
    * This way the launching entity doesn't need to have S3 access configured.
* Use sm_task_name for base job name. 
    * By explicitly setting `job_name` previously, running the same job twice (with different arguments)
      would lead to an error because SM requires unique job names. By using the base job name instead,
      SM will create unique timestamps for job names.


sagemaker/launch/launch_partition.py
```diff
@@ -78,21 +60,21 @@ def run_job(input_args, image):
...
-        image_uri=container_image_uri,
+        image_uri=image,
         role=role,
         instance_count=instance_count,
         instance_type=instance_type,
         py_version="py3",
         hyperparameters=params,
         sagemaker_session=sagemaker_session,
-        base_job_name=f"gs-partition-{graph_name}",
+        base_job_name=f"gs-partition-{sm_task_name}",
```
classicsong, line 70: if task name is None, can we give a default name instead of ""?

classicsong (approved):

sagemaker/launch/launch_partition.py
```diff
@@ -1,15 +1,13 @@
...
-import json
 import logging
 import boto3 # pylint: disable=import-error
 
 from sagemaker.pytorch.estimator import PyTorch
-from sagemaker.s3 import S3Downloader
 import sagemaker
 
-from common_parser import get_common_parser, parse_estimator_kwargs
+from common_parser import get_common_parser, parse_estimator_kwargs # pylint: disable=wrong-import-order
```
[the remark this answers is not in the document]
thvasilo: I think this is a false positive of pylint, what I'm seeing is:

> third party import "from common_parser import get_common_parser, parse_estimator_kwargs" should be placed before "from sagemaker.pytorch.estimator import PyTorch"

So it thinks `from common_parser` is a third-party import, and `from sagemaker ...`  is a local import, possibly because the root directory is named `sagemaker`.  Because it wants third-party imports to be placed before local imports, we get this false positive here.

sagemaker/launch/launch_partition.py
thvasilo: Sure, I'll replace it with a timestamp

PR #407, opened 2023-08-31
Add first version for the workflow on GSProcessing

jalencato:
*Issue #, if available:*

*Description of changes:*

Will delete the PR after testing


.github/workflow_scripts/gsprocessing.sh
```diff
@@ -0,0 +1,3 @@
+cd /usr/lib/spark/graphstorm/graphstorm-processing/
+pip install .
+pytest .
```
thvasilo, line 3: Let's add a newline at the end of the file.

.github/workflows/continuous-integration.yml
```diff
@@ -246,5 +253,5 @@ jobs:
                                --command "${{ env.COMMAND-E2E-MGPU }}" \
                                --remote https://github.com/'${{ github.event.pull_request.head.repo.full_name }}' \
                                --source-ref '${{ github.event.pull_request.head.sha }}' \
-                               --wait    
-
+                               --wait
+
```
thvasilo, line 257: Same here, add newline.

.github/workflow_scripts/gsprocessing.sh
thvasilo, line 1: Let's rename this file to `gsprocessing-pytest.sh`.

.github/workflows/gsprocessing_workflow.yml
```diff
@@ -0,0 +1,120 @@
...
+  contents: read
+
+defaults:
+  run:
+    shell: bash
+
+# All test scripts split in individual .sh files and moved to .github/workflow scripts
+env:
+  COMMAND-LINT: chmod +x ./lint_check.sh && ./lint_check.sh
+  COMMAND-GSProcessing: chmod +x ./gsprocessing.sh && ./gsprocessing.sh
```
thvasilo, line 33: The commands can just be:

```suggestion
  COMMAND-LINT: bash ./lint_check.sh
  COMMAND-GSProcessing-PYTEST: bash ./gsprocessing-pytest.sh
```

.github/workflows/gsprocessing_workflow.yml
```diff
@@ -0,0 +1,120 @@
...
+
+jobs:
+  lint_check:
```
thvasilo, line 36: Here we are using the common `lint_check.sh` script right? That one does a bunch of checks on the main GS repo that would be unnecessary here (and vice versa), since we know (from our workflow definition) that only files under `./graphstom-processing` were modified.

I see two options to avoid this:
1. Create a new `gsprocessing-lint.sh` workflow script that only checks GSProcessing code and execute that. Then we remove the gsprocessing-related line from `lint_check.sh`.
2. Add an argument to `lint_check.sh` that tells it whether to run pylint on gsprocessing or graphstorm files. The GSProcessing `COMMAND_LINT` could then be `bash ./lint_check.sh gsprorcessing` and similarly, in the graphstorm workflow it would `bash ./lint_check graphstorm`.

I'd be OK with either option of the above.

.github/workflows/gsprocessing_workflow.yml
```diff
@@ -0,0 +1,120 @@
...
+      uses: actions/checkout@v3
+    - name: Set up Python 3.x
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.x'
+    - name: Install dependencies
+      shell: bash
+      run: |
+        python3 -m pip install --upgrade pip
+        python3 -m pip install pytest
```
thvasilo, line 55: I do wonder why we're installing pytest on the Github runner, since the tests will run on Batch. We can skip this right?

.github/workflows/gsprocessing_workflow.yml
```diff
@@ -0,0 +1,120 @@
...
```
thvasilo, line 98: Same here, is `pytest` needed?

.github/workflows/gsprocessing_workflow.yml
thvasilo: Note that the above recommendation changes the name of the command/env var, as well as the name of script for pytest.

thvasilo (changes requested):
Overall looks good, but need some changes. Let's make sure that `pylint` is only ran on files that have possibly changed for each project in the repo (see comments)

.github/workflows/continuous-integration.yml
jalencato: The 257 is the new line, right?

thvasilo: Yup just add a newline at the end of the file.

thvasilo: Unresolving, the trailing whitespace on line 263 is still there.

.github/workflow_scripts/gsprocessing_pytest.sh
```diff
@@ -0,0 +1,4 @@
+cd /usr/lib/spark/graphstorm/graphstorm-processing/
+pip install .
```
thvasilo, line 2: Let's move this to be pip3 as well to ensure we're installing to the right python env.

PR #409, opened 2023-09-01
[Fix] Fix node id mapping back logic

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
This PR fixes the previous node ID mapping logic. Not use the node_mapping.pt node IDs to map back, but use the index of node IDs to map back.


classicsong (approved):
LGTM

PR #410, opened 2023-09-01
Multiple target node training with per node type decoder

CongWeilin:
*Issue #, if available:*

*Description of changes:*

A decoder is created for each node type. This PR also include functions that create per node type loss function, but haven't been used in `create_builtin_node_gnn_model` yet.


python/graphstorm/gsf.py
```diff
@@ -131,14 +131,16 @@ def create_builtin_node_gnn_model(g, config, train_task):
...
         Whether this model is used for training.
+    share_decoder: bool
+        Whether this model share decoder for different node types.
 
     Returns
     -------
     GSgnnModel : The GNN model.
     """
-    return create_builtin_node_model(g, config, train_task)
+    return create_builtin_node_model(g, config, train_task, share_decoder)
```
CongWeilin, line 141: If we can add `share_decoder` as part of config, then user can directly create this model by setting model arguments.

python/graphstorm/model/gnn.py
```diff
@@ -395,6 +412,22 @@ def set_loss_func(self, loss_fn):
                 'The loss function should be the class of nn.Module.'
         self._loss_fn = loss_fn
 
+    def set_loss_func_dict(self, loss_fns):
```
CongWeilin, line 415: I just add this function here, but I did this function when creating graphstorm's model.

python/graphstorm/gsf.py
```diff
@@ -161,18 +165,36 @@ def create_builtin_node_model(g, config, train_task):
...
+                                                else model.node_input_encoder.out_dims,
+                                               config.num_classes,
+                                               config.multilabel))
+        else:
+            decoder = {}
+            for ntype in config.target_ntype:
+                decoder[ntype] = EntityClassifier(model.gnn_encoder.out_dims \
+                                                if model.gnn_encoder is not None \
+                                                else model.node_input_encoder.out_dims,
+                                               config.num_classes,
```
zheng-da, line 180: do we need to extend this? different tasks may have different numbers of classes.

python/graphstorm/gsf.py
```diff
@@ -161,18 +165,36 @@ def create_builtin_node_model(g, config, train_task):
...
+                                               config.multilabel)
```
zheng-da, line 181: similarly to this one.

python/graphstorm/model/gnn.py
```diff
@@ -383,6 +383,23 @@ def set_decoder(self, decoder):
                     + 'match the input dimension of the decoder.'
         self._decoder = decoder
 
+    def set_decoder_dict(self, decoders):
```
zheng-da, line 386: instead of adding a new method, can we reuse `set_decoder` and have it to support a dict of decoders?

python/graphstorm/model/gnn.py
zheng-da: the same comment as above

zheng-da: this function isn't used to set different loss functions for different node types.

python/graphstorm/model/gnn.py
CongWeilin: merged in new commit.

python/graphstorm/gsf.py
CongWeilin: solved

python/graphstorm/model/gnn.py
CongWeilin: solved. In the update, variables `config.num_classess`, `config.multilabel`, `config.multilabel_weights`, and `config.imbalance_class_weights` can be defined as a per node type dictionary.

However, the current implementation only support node classification but does not support node regression, because those variables are not used in regression task in graphstorm.

python/graphstorm/config/argument.py
```diff
@@ -1195,21 +1207,40 @@ def multilabel_weights(self):
...
-                assert w >= 0., "multilabel weights can not be negative values"
-            assert len(weights) == self.num_classes, \
-                "Each class must have an assigned weight"
-
-            return th.tensor(weights)
+        if hasattr(self, "_num_classes") and isinstance(self.num_classes, dict):
+            if hasattr(self, "_multilabel_weights"):
+                ntype_weights = {}
+                for ntype in self.num_classes:
+                    assert self.multilabel[ntype] is True, \
```
zheng-da, line 1214: should we save the `multilabel` result somewhere first? otherwise, it will trigger computations in multilabel every time.

python/graphstorm/config/argument.py
```diff
@@ -1195,21 +1207,40 @@ def multilabel_weights(self):
...
+                        "Must be a multi-label classification task."
+                    try:
+                        weights = self._multilabel_weights[ntype].split(",")
+                        weights = [float(w) for w in weights]
+                    except Exception: # pylint: disable=broad-except
+                        assert False, "The weights should in following format 0.1,0.2,0.3,0.1,0.0"
```
zheng-da, line 1220: can we raise an error instead of using assert? it looks weird.

python/graphstorm/config/argument.py
```diff
@@ -1195,21 +1207,40 @@ def multilabel_weights(self):
...
+                assert self.multilabel is True, "Must be a multi-label classification task."
+                try:
+                    weights = self._multilabel_weights.split(",")
+                    weights = [float(w) for w in weights]
+                except Exception: # pylint: disable=broad-except
+                    assert False, "The weights should in following format 0.1,0.2,0.3,0.1,0.0"
+                for w in weights:
+                    assert w >= 0., "multilabel weights can not be negative values"
+                assert len(weights) == self.num_classes, \
+                    "Each class must have an assigned weight"
```
zheng-da, line 1239: can we wrap the checks here in a function? so we don't need to duplicate the code above.

zheng-da: the same comments apply to other places as well.

python/graphstorm/run/gsgnn_np/gsgnn_np.py
```diff
@@ -35,13 +35,22 @@ def get_evaluator(config):
...
+                                    config.early_stop_rounds,
+                                    config.early_stop_strategy)
+        else:
+            return GSgnnAccEvaluator(config.eval_frequency,
+                                    config.eval_metric,
+                                    config.multilabel,
+                                    config.use_early_stop,
+                                    config.early_stop_burnin_rounds,
+                                    config.early_stop_rounds,
+                                    config.early_stop_strategy)
```
zheng-da, line 53: the only difference is `multilabel`? if so, can we write it like this:
```
multilabel = config.multilabel[config.eval_target_ntype] if isinstance(config.multilabel, dict) else config.multilabel
```

CongWeilin: solved

python/graphstorm/config/argument.py
CongWeilin: solved

python/graphstorm/config/argument.py
CongWeilin: solved

python/graphstorm/config/argument.py
CongWeilin: solved

CongWeilin:
integrate test is provided:

- (step 1) A multi-target node version of movieLens dataset is created in `test/data_gen/create_data.sh` and  `data_gen/process_movielens.py`

- (step 2) A multi-target node single-label classification test case is created in `test/graphstorm-nc/test.sh`

- (step 3) A multi-target node multilabel classification movielens is generated using `test/data_gen/gen_multilabel.py`. Then,  a multi-target node multilabel classification test case is created in `test/graphstorm-nc/test.sh`

docs/source/advanced/advanced-usages.rst
```diff
@@ -16,8 +16,51 @@ GraphStorm provides supports to avoid theses problems:
...
+
+- Evaluate on single node type: During evuation, the users can set a single node type for evaluation. For example, by setting ``eval_target_ntype:  movie``, we will only perform evaluation on "movie" node type. Our current implementation only support evaluating on a single node type.
+
+- Per target node type decoder: The users may also want to use a different decoder on each node type, where the output dimension for each decoder maybe different. We can achieve this by setting ``num_classes`` in model config YAML file. For example, by setting ``num_classes`` as following, GraphStorm will create a decoder with output dimension as 3 for movie node type, and a decoder with output dimension as 7 for user node type.
+
+  .. code-block:: yaml
+
+    num_classes:
+      movie:  3
+      user:  7
```
zheng-da, line 48: can we still use `num_classes: 7` to specify the number of classes for one target node type?

zheng-da: the same question goes to other arguments as well.

zheng-da: if not, we are changing our interface significantly.

python/graphstorm/config/argument.py
```diff
@@ -1302,21 +1335,37 @@ def imbalance_class_weights(self):
...
-            assert len(weights) == self.num_classes, \
+            assert len(weights) == num_classes, \
                 "Each class must have an assigned weight"
-
             return th.tensor(weights)
 
-        return None
+        if hasattr(self, "_num_classes") and isinstance(self.num_classes, dict):
+            if hasattr(self, "_imbalance_class_weights"):
+                assert isinstance(self._imbalance_class_weights, dict)
```
zheng-da, line 1353: please provide an error message.

docs/source/advanced/advanced-usages.rst
CongWeilin: > can we still use `num_classes: 7` to specify the number of classes for one target node type?

Yes, we can still use this command for single node type, same for other arguments. The above instruction is speficially for more than 1 node types.

CongWeilin: Actually, most of the integrate test cases and unit test cases are single node type, and we don't have to change their commands.

python/graphstorm/config/argument.py
CongWeilin: solved.

zheng-da (approved):
LGTM.

zheng-da:
@classicsong can you take a look at this PR?

python/graphstorm/config/argument.py
```diff
@@ -1302,21 +1335,38 @@ def imbalance_class_weights(self):
...
+        if hasattr(self, "_num_classes") and isinstance(self.num_classes, dict):
+            if hasattr(self, "_imbalance_class_weights"):
+                assert isinstance(self._imbalance_class_weights, dict), \
+                    print('The imbalance_class_weights should be dictionary')
+                num_classes = self.num_classes
+                imbalance_class_weights = self._imbalance_class_weights
+                ntype_weights = {}
+                for ntype in num_classes:
+                    ntype_weights[ntype] = check_imbalance_class_weights(
+                        imbalance_class_weights[ntype],
```
classicsong, line 1360: What if some node types have imbalance class weights and some node types do not?

python/graphstorm/config/argument.py
```diff
@@ -1341,6 +1391,24 @@ def target_ntype(self):
...
+        # pylint: disable=no-member
+        if hasattr(self, "_eval_target_ntype"):
+            assert isinstance(self._eval_target_ntype, str), \
+                "Now we only support single ntype evaluation"
+            return self._eval_target_ntype
+        else:
+            if isinstance(self.target_ntype, str):
+                return self.target_ntype
+            elif isinstance(self.target_ntype, list):
+                # (wlcong) Now only support single ntype evaluation
```
classicsong, line 1407: Can you add a warning here?

CongWeilin: solved.

python/graphstorm/config/argument.py
CongWeilin: solved.

PR #412, opened 2023-09-02
Update GSgnnLinkPredictionModelInterface and support mini-batch based full graph inference

classicsong:
*Issue #, if available:*

*Description of changes:*


python/graphstorm/model/gnn.py
```diff
@@ -673,7 +678,14 @@ def get_input_embeds(input_nodes):
                 input_nodes = {data.g.ntypes[0]: input_nodes}
             return {ntype: input_embeds[ntype][ids].to(device) \
                     for ntype, ids in input_nodes.items()}
-        embeddings = model.gnn_encoder.dist_inference(data.g, get_input_embeds,
+        if minibatch:
```
zheng-da, line 681: we should have a function called `mini_batch_inference` outside `full_graph_inference`

python/graphstorm/model/gnn_encoder_base.py
```diff
@@ -104,6 +104,83 @@ def dist_inference(self, g, get_input_embeds, batch_size, fanout,
         return dist_inference(g, self, get_input_embeds, batch_size, fanout,
                               edge_mask=edge_mask, task_tracker=task_tracker)
 
+
+def dist_minibatch_inference(g, gnn_encoder, get_input_embeds, batch_size, fanout,
```
zheng-da, line 108: why do we want to have mini-batch inference on the entire graph?

python/graphstorm/model/gnn.py
classicsong: It is still full graph inference, but in a mini-batch manner.

python/graphstorm/model/gnn_encoder_base.py
classicsong: If there is a custom model, we may not be able to do layer-by-layer full graph inference.
That's why we need mini-batch inference.

python/graphstorm/inference/ep_infer.py
```diff
@@ -77,7 +77,8 @@ def infer(self, loader, save_embed_path, save_prediction_path=None,
         sys_tracker.check('start inferencing')
         self._model.eval()
         embs = do_full_graph_inference(self._model, loader.data, fanout=loader.fanout,
-                                       task_tracker=self.task_tracker)
+                                       task_tracker=self.task_tracker,
+                                       minibatch=use_mini_batch_infer)
```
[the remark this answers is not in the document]
classicsong: Now we have two ways to do full graph inference: 1) layer-by-layer inference and 2) node-batch by node-batch (mini-batch here) inference.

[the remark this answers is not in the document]
zheng-da: my question is why we do mini-batch inference in the function called `do_full_graph_inference`.

python/graphstorm/model/gnn.py
zheng-da: is there any advantage of computing GNN embeddings of all nodes in a graph?

python/graphstorm/model/gnn_encoder_base.py
zheng-da: i see.

python/graphstorm/inference/lp_infer.py
```diff
@@ -43,6 +43,7 @@ class GSgnnLinkPredictionInfer(GSInfer):
     # TODO(zhengda) We only support full-graph inference for now.
     def infer(self, data, loader, save_embed_path,
             edge_mask_for_gnn_embeddings='train_mask',
+            use_mini_batch_infer=False,
```
zheng-da, line 46: should we have this as an argument? if mini-batch inference is used for custom models, we can easily test if the input model is a custom model.

python/graphstorm/model/gnn.py
```diff
@@ -622,7 +623,7 @@ def loss_func(self):
         return self._loss_fn
 
 def do_full_graph_inference(model, data, batch_size=1024, fanout=None, edge_mask=None,
-                            task_tracker=None):
+                            task_tracker=None, minibatch=False):
```
zheng-da, line 626: i think it's better to not invoke mini-batch inference code inside full-graph inference.

zheng-da (approved):
overall, it looks good to me. We need to be careful of using mini-batch inference when we need to compute BERT embeddings on nodes.

PR #420, opened 2023-09-06
Update the implementation of getting edge features for link prediction decoders. 

classicsong:
*Issue #, if available:*
Based on the API change in #412 , the implementation of how decoder edge features are passed to decoder is also changed.

*Description of changes:*
This PR follows https://github.com/awslabs/graphstorm/pull/412


python/graphstorm/model/edge_decoder.py
```diff
@@ -487,17 +487,64 @@ def _compute_logits(self, g, h):
         return out
 
 ##################### Link Prediction Decoders #######################
-class LinkPredictDotDecoder(GSLayerNoParam):
-    """ Link prediction decoder with the score function of dot product
+class LinkPredictNoParamDecoder(GSLayerNoParam):
```
zheng-da, line 490: can you explain why we need to distinguish them?

python/graphstorm/dataloading/dataset.py
```diff
@@ -416,6 +416,44 @@ def eval_etypes(self):
...
+        The graph name
+    part_config : str
+        The path of the partition configuration file.
+    train_etypes : tuple of str or list of tuples
+        Target edge types for training
+    eval_etypes : tuple of str or list of tuples
+        Target edge types for evaluation
+    label_field : str
+        The field for storing labels
+    node_feat_field: str or dict of str
```
zheng-da, line 434: do we support multiple feature fields for one node type? the same question applies to edge features and positive graph features.

python/graphstorm/model/edge_decoder.py
```diff
@@ -927,7 +935,13 @@ def forward(self, g, h):
                 rel_embedding = rel_embedding.repeat(1,dest_emb.shape[0]).T
                 scores_etype = calc_distmult_pos_score(src_emb, dest_emb, rel_embedding)
 
-                weight = _get_edge_weight(g, LP_DECODER_EDGE_WEIGHT, canonical_etype)
+                if e_h is not None and canonical_etype in e_h.keys():
+                    weight = e_h[canonical_etype]
```
zheng-da, line 939: should we assert that edge weight can only be a tensor?

python/graphstorm/model/edge_decoder.py
classicsong: LinkPredictNoParamDecoder Is for decoders that do not have trainable parameters such as LinkPredictDotDecoder.
LinkPredictLearnableDecoder is for decoders that have trainable parameters, i.e., nn.Module.

That's why we need to define two different abstract classes.

python/graphstorm/dataloading/dataset.py
classicsong: We do. Actually node_feat_field should be str or dist of list of str.

classicsong: I will update the docstring.

python/graphstorm/model/edge_decoder.py
```diff
@@ -963,7 +980,12 @@ def forward(self, g, h): # pylint: disable=arguments-differ
                 dest_emb = h[dest_type][v]
                 scores_etype = calc_dot_pos_score(src_emb, dest_emb)
 
-                weight = _get_edge_weight(g, LP_DECODER_EDGE_WEIGHT, canonical_etype)
+                if e_h is not None and canonical_etype in e_h.keys():
+                    weight = e_h[canonical_etype]
```
zheng-da, line 984: we should check here as well.

python/graphstorm/model/edge_decoder.py
```diff
@@ -927,7 +935,15 @@ def forward(self, g, h):
                 rel_embedding = rel_embedding.repeat(1,dest_emb.shape[0]).T
                 scores_etype = calc_distmult_pos_score(src_emb, dest_emb, rel_embedding)
 
-                weight = _get_edge_weight(g, LP_DECODER_EDGE_WEIGHT, canonical_etype)
+                if e_h is not None and canonical_etype in e_h.keys():
+                    weight = e_h[canonical_etype]
+                    assert th.is_tensor(weight), \
+                        "The edge weight for Link prediction must be a torch tensor."
```
zheng-da, line 941: it's possible that there are multiple edge features. I think the error message should say that this decoder doesn't support multiple edge data.

zheng-da (approved):
LGTM

python/graphstorm/model/edge_decoder.py
classicsong: For edge_weight, it should be a single edge feature

PR #421, opened 2023-09-07
Add base classes for dataloader.

zheng-da:
*Issue #, if available:*

*Description of changes:*


python/graphstorm/dataloading/dataloading.py
```diff
@@ -87,7 +87,71 @@ def sample(self, seeds):
         input_nodes = {ntype: block.srcnodes[ntype].data[dgl.NID] for ntype in block.srctypes}
         return block, input_nodes
 
-class GSgnnEdgeDataLoader():
+class GSgnnEdgeDataLoaderBase():
+    """ The base dataloader class for edge tasks.
+
```
zhjwy9343, line 92: Should users extend this class if they want to implement their own dataloader? If so, could you provide examples about what functions they need to overwrite to customize?

This may also apply to the `NodeDataLoaderBase` and `LinkPredictionDataLoaderBase`.

python/graphstorm/dataloading/dataloading.py
```diff
@@ -87,7 +87,71 @@ def sample(self, seeds):
...
+        """
+        return self._data
+
+    @property
+    def target_eidx(self):
+        """ Target edge idx for prediction
+
+        Returns
+        -------
+        dict of Tensors : the target edge IDs.
```
zhjwy9343, line 140: This is inconsistent with the init's doc string.

The targe_eidx is not only a dict of Tensors, but also could be a Tensor, right?

python/graphstorm/dataloading/dataloading.py
```diff
@@ -216,7 +278,74 @@ def fanout(self):
...
 
-class GSgnnLinkPredictionDataLoader():
+class GSgnnLinkPredictionDataLoaderBase():
+    """ The base class of link prediction dataloader.
+
+    Parameters
+    ----------
+    dataset: GSgnnEdgeData
+        The GraphStorm edge dataset
+    target_idx : dict of Tensors
```
zhjwy9343, line 288: Could this target_idx be a Tensor, like in the `EdgeDataLoaderBase` and `NodeDataLoaderBase`?

python/graphstorm/dataloading/dataloading.py
```diff
@@ -216,7 +278,74 @@ def fanout(self):
...
+        self._fanout = fanout
+
+    def __iter__(self):
+        """ Returns an iterator object
+        """
+
+    def __next__(self):
+        """ Return a mini-batch for link prediction.
+
+        A mini-batch of link prediction contains three objects:
```
zhjwy9343, line 305: three -> four

python/graphstorm/dataloading/dataloading.py
```diff
@@ -216,7 +278,74 @@ def fanout(self):
...
+        """
+        return self._fanout
+
+    @property
+    def target_eidx(self):
+        """ The target edges for prediction.
+
+        Returns
+        -------
+        dict of Tensors : the target edge IDs.
```
zhjwy9343, line 344: ditto

python/graphstorm/dataloading/dataloading.py
```diff
@@ -216,7 +278,74 @@ def fanout(self):
 BUILTIN_FAST_LP_LOCALUNIFORM_NEG_SAMPLER = 'fast_localuniform'
 BUILTIN_FAST_LP_LOCALJOINT_NEG_SAMPLER = 'fast_localjoint'
 
-class GSgnnLinkPredictionDataLoader():
+class GSgnnLinkPredictionDataLoaderBase():
+    """ The base class of link prediction dataloader.
+
```
zhjwy9343, line 283: ditto

python/graphstorm/dataloading/dataloading.py
```diff
@@ -766,7 +882,71 @@ def _prepare_negative_sampler(self, num_negative_edges):
 
 ################ Minibatch DataLoader (Node classification) #######################
 
-class GSgnnNodeDataLoader():
+class GSgnnNodeDataLoaderBase():
+    """ The base dataloader class for node tasks.
+
```
zhjwy9343, line 887: ditto

python/graphstorm/dataloading/dataloading.py
```diff
@@ -766,7 +882,71 @@ def _prepare_negative_sampler(self, num_negative_edges):
...
+        """
+        return self._data
+
+    @property
+    def target_nidx(self):
+        """ Target edge idx for prediction
+
+        Returns
+        -------
+        dict of Tensors : the target edge IDs.
```
zhjwy9343, line 935: ditto

zhjwy9343:
Please also import these three base dataloaders in this folder's __init__.py file, so Sphinix can import them from the dataloading package.

python/graphstorm/dataloading/dataloading.py
zheng-da: are you sure docstring is a good place to write examples to illustrate how to customize a data loader?

zheng-da: how complete the example should be?

zhjwy9343: We expose this API mainly because we want users to use this Class to implement their own model. So it is better to give some explanation and example here.

I would suggest to give a sudo code to let users to know they need to extend this class.

zheng-da: The methods that require users to implement are a common interface of Python iterator. any Python users should understand this interface. if we want to provide an example, we should provide a more realistic example.

zhjwy9343: OK

zhjwy9343 (approved):
LGTM.

PR #425, opened 2023-09-08
Test PR

JCy1rus:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -8,7 +8,6 @@ on:
       - main
       - ci_dev
   pull_request_target:
-    types: [labeled]
     
 
 permissions:
@@ -66,7 +65,7 @@ jobs:
                                --source-ref '${{ github.ref }}' \
                                --wait
     - name: Submit Job (for Pull Request)
-      if: ${{ github.event_name == 'pull_request_target'}}
+      if: ${{ github.event_name == 'pull_request_target' }}
       shell: bash
       run: |
         echo "Start submitting job - Pull Request"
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -28,9 +28,12 @@ env:
   COMMAND-E2E-MGPU: chmod +x ./e2e_mgpu_check.sh && ./e2e_mgpu_check.sh
 
 jobs:
-
+  
   lint_check:
-    if: ${{ github.event.label.name != 'draft' }}
+    if: ${{ github.event_name == 'push' ||
+      (github.event.label.name != 'draft'
```
classicsong, line 34: Should this also be ```pull_request.```?

.github/workflows/continuous-integration.yml
```diff
@@ -28,9 +28,12 @@ env:
...
+      && github.event_name == 'pull_request_target'
```
thvasilo, line 35: I don't think we should be using this way to check the event. The [official guidelines](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) from Github security suggest using:

```yaml
on:
      pull_request_target:
        types: [labeled]

    jobs:
      build:
        name: Build and test
        runs-on: ubuntu-latest
        if: contains(github.event.pull_request.labels.*.name, 'safe to test')
```

was there any issue with using this approach?

thvasilo (changes requested):

.github/workflows/continuous-integration.yml
thvasilo: Discussed offline, I'm OK with using this approach to have the condition at the job level instead of the workflow level.

thvasilo (approved):
LGTM now. The workflow only triggers for push which requires write permissions, and for any PRs. When triggered the additional condition will ensure that only PRs with the `ready` tag will be able to run the CI.

PR #426, opened 2023-09-10
Refactor the code of BERT computation to fix a potential bug.

zheng-da:
*Description of changes:*
The original code locates a BERT model based on the index. There is potentially a bug here. If a user fine-tune BERT models on two node types:
```
  node_lm_models:
    -
      lm_type: bert
      model_name: "bert-base-uncased"
      gradient_checkpoint: true
      node_types:
        - paper
    -
      lm_type: bert
      model_name: "bert-base-uncased"
      gradient_checkpoint: true
      node_types:
        - fos
```
If the user trains the GNN model with the following config,
```
  node_lm_models:
    -
      lm_type: bert
      model_name: "bert-base-uncased"
      gradient_checkpoint: true
      node_types:
        - fos
    -
      lm_type: bert
      model_name: "bert-base-uncased"
      gradient_checkpoint: true
      node_types:
        - paper
```
The original code will use the wrong BERT model for the node types.
The fix is to associate the BERT model with node types and the parameter names contain the node types to make sure the parameters are loaded to the right BERT model.

### DIFF
--- a/python/graphstorm/model/lm_embed.py
+++ b/python/graphstorm/model/lm_embed.py
@@ -119,7 +119,9 @@ class LMModels(nn.Module):
                 assert ntype not in self._lm_node_feats, \
                         f"More than one BERT model runs on Node {ntype}."
                 self._lm_node_feats[ntype] = feats
-            key = str(lm_ntypes)
+            # We should sort the node type list before converting it to the key.
+            lm_ntypes.sort()
+            key = ','.join(lm_ntypes)
             self._lm_models[key] = lm_model
             for ntype in lm_ntypes:
                 self._lm_map[ntype] = key
--- a/python/graphstorm/model/lm_embed.py
+++ b/python/graphstorm/model/lm_embed.py
@@ -20,6 +20,7 @@
 import logging
 
 import torch as th
+from torch import nn
 import dgl
 
 from .embed import GSNodeInputLayer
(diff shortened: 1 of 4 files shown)


python/graphstorm/model/lm_embed.py
```diff
@@ -28,109 +29,200 @@
...
+            lm_ntypes = lm_config["node_types"]
+            lm_node_feats = get_lm_node_feats(g, lm_model, lm_ntypes)
+            for ntype, feats in lm_node_feats.items():
+                assert ntype not in self._lm_node_feats, \
+                        f"More than one BERT model runs on Node {ntype}."
+                self._lm_node_feats[ntype] = feats
+            key = str(lm_ntypes)
+            self._lm_models[key] = lm_model
             for ntype in lm_ntypes:
+                self._lm_map[ntype] = key
```
classicsong, line 125: What is this? A map from ntype to ntype?

python/graphstorm/model/lm_embed.py
```diff
@@ -163,31 +255,17 @@ def __init__(self,
...
-            # A list of node types sharing the same lm model
-            lm_ntypes = lm_config["node_types"]
-            lm_node_feats = get_lm_node_feats(g, lm_model, lm_ntypes)
-            lm_models.append(lm_model)
-            lm_models_info.append((lm_ntypes, lm_node_feats))
-            node_with_lm_feats.update(lm_ntypes)
-        assert len(node_with_lm_feats) == len(g.ntypes), \
-            "Every node type in a graph should have text feature"
+        self._lm_models = LMModels(g, node_lm_configs, num_train, lm_infer_batch_size)
+        #assert len(self._lm_models.ntypes) == len(g.ntypes), \
```
classicsong, line 259: Remove this assert

python/graphstorm/model/lm_embed.py
```diff
@@ -28,109 +29,200 @@
...
+            # No bert training, Get cached LM embedding
+            # Note: self.lm_emb_cache is initialized by calling warmup
+            for ntype, idx in input_nodes.items():
+                if ntype in lm_emb_cache:
+                    lm_feats[ntype] = lm_emb_cache[ntype][idx].to(dev)
+        else:
+            # TODO: Release the bert cache properly
+            #       This may need support from DistDGL
+            # Need bert training
+            for ntype in self.ntypes:
```
classicsong, line 153: If multiple ntypes share the same lm_model, can we batch them together to improve efficiency?

python/graphstorm/model/lm_embed.py
zheng-da: a mapping from ntype to the key in `_lm_models`. The key of `_lm_models` is not ntype. a LM model can be used by multiple node types. The key is the string that represents all node types using the LM model.

python/graphstorm/model/lm_embed.py
zheng-da: does it matter? I tested it. there is no difference for batch size of 32, 128, 1024.
also, how can we batch bert computation on different node types? the length of text sequences on different node types can be different.

python/graphstorm/model/lm_embed.py
classicsong: Shall we sort the ntypes before creating the key?

classicsong (approved):
LGTM

PR #430, opened 2023-09-11
Fix categorical transform

classicsong:
*Issue #, if available:*
An error happens when the loaded categorical features include both integers and strings. This may happen when loading features from CSV files.

*Description of changes:*

### DIFF
--- a/python/graphstorm/gconstruct/transform.py
+++ b/python/graphstorm/gconstruct/transform.py
@@ -351,6 +351,9 @@ class CategoricalTransform(TwoPhaseFeatTransform):
     def __init__(self, col_name, feat_name, separator=None, transform_conf=None):
         self._val_dict = {}
         if transform_conf is not None and 'mapping' in transform_conf:
+            # We assume the keys of a categorical mapping are strings.
+            # But previously keys can be integers. So we convert them
+            # into strings.
             self._val_dict = \
                 {str(key): val for key, val in transform_conf['mapping'].items()}
             self._conf = transform_conf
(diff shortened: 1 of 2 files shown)


python/graphstorm/gconstruct/transform.py
```diff
@@ -379,7 +379,7 @@ def pre_process(self, feats):
 
         feats = feats[feats != None] # pylint: disable=singleton-comparison
         if self._separator is None:
-            return {self.feat_name: np.unique(feats)}
+            return {self.feat_name: np.unique(feats.astype(str))}
```
zheng-da, line 382: is it good to convert everything to strings?

zheng-da:
you need to maintain the backward compatibility

classicsong:
> you need to maintain the backward compatibility

I have added.

python/graphstorm/gconstruct/transform.py
```diff
@@ -351,7 +351,8 @@ class CategoricalTransform(TwoPhaseFeatTransform):
     def __init__(self, col_name, feat_name, separator=None, transform_conf=None):
         self._val_dict = {}
         if transform_conf is not None and 'mapping' in transform_conf:
-            self._val_dict = transform_conf['mapping']
+            self._val_dict = \
+                {str(key): val for key, val in transform_conf['mapping'].items()}
```
classicsong, line 355: @zheng-da fix the backward compatibility

zheng-da: you should add comments here to explain why the keys are converted to strings.

zheng-da (approved):
please add comments to explain the data type conversion. otherwise, it looks good to me.

PR #436, opened 2023-09-13
add labeled

jalencato:
*Issue #, if available:*

*Description of changes:*

Allow developer to add ready tag after they creating the PR request


thvasilo (approved):
LGTM this should allow us to trigger CI even if the `ready` label is added after a PR is opened, which is the expected usecase for external contributors.

PR #441, opened 2023-09-17
Store bert embeddings with float16

zheng-da:
*Description of changes:*
We can store bert embeddings with float16 to reduce memory. It doesn't change the model performance.


classicsong (commented):
Shall we provide an option to users to store bert embedding in float32 or float16? There can be other choices like bf16

classicsong (approved):

classicsong (approved):

PR #442, opened 2023-09-17
argument num_classess > 0 instead of > 1

CongWeilin:
*Issue #, if available:*

If our task is binary classification and evaluate by precision-recall, we might need to set `num_classes = 1` and `multilabel=true`.

We may not want to set `num_classes = 2` and `multilabel=false` for this case because precision-recall is computed on the positive class.

*Description of changes:*

### DIFF
--- a/python/graphstorm/config/argument.py
+++ b/python/graphstorm/config/argument.py
@@ -1249,9 +1249,9 @@ class GSConfig:
             "Must provide the number possible labels through num_classes"
         if isinstance(self._num_classes, dict):
             for num_classes in self._num_classes.values():
-                assert num_classes > 1
+                assert num_classes > 0
         else:
-            assert self._num_classes > 1
+            assert self._num_classes > 0
         return self._num_classes
 
     @property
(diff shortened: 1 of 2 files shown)


classicsong (commented):
Can num_classes == 1?
For binary classification, should n't it be 2, i.e., 0 and 1?

classicsong:
Is this related to how precision-recall is implemented?

CongWeilin:
> Is this related to how precision-recall is implemented?

Yes, precision recall only compute on positive score (https://scikit-learn.org/stable/modules/generated/sklearn.metrics.precision_recall_curve.html). So in this case, we may want 1-d prediction instead of 2-d for binary classification

python/graphstorm/config/argument.py
```diff
@@ -1249,9 +1249,9 @@ def num_classes(self):
             "Must provide the number possible labels through num_classes"
         if isinstance(self._num_classes, dict):
             for num_classes in self._num_classes.values():
-                assert num_classes > 1
+                assert num_classes > 0
```
classicsong, line 1252: Can you write some comments here to explain why?

CongWeilin: I added a comments in the code. Please check.

classicsong (approved):

PR #446, opened 2023-09-19
Remove unnecessary rank argument.

zheng-da:
*Description of changes:*
GraphStorm can identify the rank automatically. We don't need to have a rank argument in many classes.


zhjwy9343 (approved):
LGTM

PR #451, opened 2023-09-20
Update np_trainer.py

CongWeilin:
*Issue #, if available:*

when using precision_recall as evaluation metric, we need to set `return_proba=True` because scikit-learn's precision_recall_curve takes probability as input.

*Description of changes:*

### DIFF
--- a/python/graphstorm/trainer/np_trainer.py
+++ b/python/graphstorm/trainer/np_trainer.py
@@ -90,6 +90,7 @@ class GSgnnNodePredictionTrainer(GSgnnTrainer):
         if self.evaluator is not None:
             assert val_loader is not None, \
                     "The evaluator is provided but validation set is not provided."
+            return_proba = True if self.evaluator.metric() == ['precision_recall'] else False
         if not use_mini_batch_infer:
             assert isinstance(self._model, GSgnnModel), \
                     "Only GSgnnModel supports full-graph inference."
@@ -167,9 +168,10 @@ class GSgnnNodePredictionTrainer(GSgnnTrainer):
                 if self.evaluator is not None and \
                     self.evaluator.do_eval(total_steps, epoch_end=False) and \
                     val_loader is not None:
+
                     val_score = self.eval(model.module if is_distributed() else model,
                                           val_loader, test_loader,
-                                          use_mini_batch_infer, total_steps, return_proba=False)
+                                          use_mini_batch_infer, total_steps, return_proba=return_proba)
 
                     if self.evaluator.do_early_stop(val_score):
                         early_stop = True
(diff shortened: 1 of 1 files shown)


python/graphstorm/trainer/np_trainer.py
```diff
@@ -167,9 +168,10 @@ def fit(self, train_loader, num_epochs,
                 if self.evaluator is not None and \
                     self.evaluator.do_eval(total_steps, epoch_end=False) and \
                     val_loader is not None:
+
                     val_score = self.eval(model.module if is_distributed() else model,
```
classicsong, line 172: Can we modify the eval() function instead of the fit() function?
And we should add an warning here that we force the return_proba = True.

python/graphstorm/trainer/np_trainer.py
```diff
@@ -261,6 +261,12 @@ def eval(self, model, val_loader, test_loader, use_mini_batch_infer, total_steps
         """
         teval = time.time()
         sys_tracker.check('before prediction')
+
+
+        if 'precision_recall' in self.evaluator.metric and return_proba==False:
```
classicsong, line 266: compute_roc_auc, compute_precision_recall_auc

python/graphstorm/trainer/np_trainer.py
```diff
@@ -261,6 +261,17 @@ def eval(self, model, val_loader, test_loader, use_mini_batch_infer, total_steps
         """
         teval = time.time()
         sys_tracker.check('before prediction')
+
+        metric = set(self.evaluator.metric)
+        need_proba = metric.intersection({'roc_auc', 'per_class_roc_auc', 'precision_recall'})
+        need_label_pred = metric.intersection({'accuracy', 'f1_score', 'per_class_f1_score'})
+        assert len(need_proba) == 0 or len(need_label_pred) == 0, \
+            print(f"{need_proba} requires return_proba==True, \
```
classicsong, line 269:
```suggestion
            f"{need_proba} requires return_proba==True, \
```

python/graphstorm/trainer/np_trainer.py
```diff
@@ -261,6 +261,17 @@ def eval(self, model, val_loader, test_loader, use_mini_batch_infer, total_steps
...
+                         but {need_label_pred} requires return_proba==False.")
```
classicsong, line 270:
```suggestion
                         but {need_label_pred} requires return_proba==False."
```

python/graphstorm/trainer/np_trainer.py
```diff
@@ -261,6 +261,17 @@ def eval(self, model, val_loader, test_loader, use_mini_batch_infer, total_steps
...
+        if len(need_proba) > 0 and return_proba==False:
+            return_proba = True
+            logging.info(f"{need_proba} requires return_proba==True.")
```
classicsong, line 273:
```suggestion
            logging.warning(f"{need_proba} requires return_proba==True. Set return_proba to True")
```

classicsong (commented):
Can you also change ep_trainer.py?

classicsong (approved):
LGTM

PR #453, opened 2023-09-20
Fix a bug in the default value of fanout

zheng-da:
*Description of changes:*
The default value of fanout should be a list.


classicsong (approved):
LGTM

PR #454, opened 2023-09-20
Fix BERT fine tune

zheng-da:
*Description of changes:*
We want to perform graph-aware fine-tuning and train a GNN model based on the fine-tuned model. The PR is to fix problems in BERT fine-tuning and GNN training.


python/graphstorm/model/lm_embed.py
```diff
@@ -266,9 +266,6 @@ def __init__(self,
             "language model configurations must be provided"
 
         self._lm_models = LMModels(g, node_lm_configs, num_train, lm_infer_batch_size)
-        assert len(self._lm_models.ntypes) == len(g.ntypes), \
```
classicsong, line 269: GSPureLMNodeInputLayer is designed for the case when the encoder layer only has lm models. If there is a ntype does not have a corresponding LM model, you should not use GSPureLMNodeInputLayer.

python/graphstorm/model/utils.py
```diff
@@ -582,9 +582,14 @@ def load_sparse_emb(num_embs, ntype_path):
                         "malicious pickle data which will execute arbitrary code " \
                         "during unpickling. Only load data you trust or " \
                         "update torch to 1.13.0+")
-                load_sparse_emb(num_embs, os.path.join(model_path, ntype))
-            else:
-                load_sparse_emb(num_embs, os.path.join(model_path, ntype))
+            emb_path = os.path.join(model_path, ntype)
+            # If we are loading the model from BERT fine-tuning, we may not have
+            # the sparse embeddings.
+            if os.path.exists(emb_path):
```
classicsong, line 588: If we are loading model from BERT fine-tuning, why do we need to load the sparse embeddings?

zheng-da: when we fine-tune the BERT model with link prediction on the node types with text features, no sparse embeddings are saved in the fine-tuning step. When we need to train the GNN model with fine-tuned BERT model, we need to load the parameters in the input encoder. At this stage, the input encoder can have learnable embeddings. But there are no sparse embeddings saved from the previous step.

python/graphstorm/model/lm_embed.py
zheng-da: why do we have GSPureLMNodeInputLayer? shouldn't we always use `GSLMNodeEncoderInputLayer`?

python/graphstorm/model/utils.py
classicsong, line 588: Add sparse_emb and dense_emb option in model_layer_to_load

tests/unit-tests/test_config.py
```diff
@@ -1305,6 +1306,8 @@ def test_load_io_info():
                          local_rank=0)
         config = GSConfig(args)
         assert config.restore_model_path == "./restore"
+        print(config.restore_model_layers)
```
classicsong, line 1309: remove this print

classicsong (approved):
LGTM

PR #458, opened 2023-09-21
[Wholegraph] Support multiple feature conversion in the script

isratnisa:
Use the `convert_feat_to_wholegraph.py` script with `--dataset-path` pointing to the distDGL folder of partitions and `--feat-names` to the features you want to convert (e.g. '`paper:feat author:feat,feat2 institution:feat`'). It will convert the node features to WholeGraph compatible format.


tools/convert_feat_to_wholegraph.py
```diff
@@ -35,47 +35,64 @@ def main(folder, feat_names):
...
     folder_pattern = re.compile(r"^part[0-9]+$")
     for path in (os.path.join(folder, name) for name in sorted( \
         (f for f in os.listdir(folder) if os.path.isdir(os.path.join(folder, f)) \
         and folder_pattern.match(f)), key=lambda x: int(x.split("part")[1]))):
         node_feats_data.append(dgl.data.utils.load_tensors(f'{path}/node_feat.dgl'))
 
+    # per node type feature
+    fname_dict = {}
+    for feat_name in feat_names:
+        print("feat_name", feat_name)
```
classicsong, line 48: Can we remove this print?

classicsong (approved):
LGTM

tools/convert_feat_to_wholegraph.py
isratnisa: yes

PR #459, opened 2023-09-21
Fix logging

zheng-da:
*Description of changes:*
We need to set up the logging as early as possible. otherwise, the log messages are not printed out correctly.


classicsong (approved):
LGTM

PR #464, opened 2023-09-22
Fix a bug in RGCN encoder

zheng-da:
*Description of changes:*
It is possible that the number of destination nodes is 0. The code will trigger the assert even though the input data are correct.

### DIFF
--- a/python/graphstorm/model/rgcn_encoder.py
+++ b/python/graphstorm/model/rgcn_encoder.py
@@ -168,8 +168,9 @@ class RelGraphConvLayer(nn.Module):
                     # tensor. This tensor won't be used for computing embeddings.
                     # We need this just to fulfill the requirements of DGL message passing
                     # modules.
-                    assert not self.self_loop, \
-                            f"We cannot allow self-loop if node {k} doesn't have input features."
+                    if g.num_dst_nodes(k) > 0:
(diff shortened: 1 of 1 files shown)


python/graphstorm/model/rgcn_encoder.py
```diff
@@ -168,8 +168,9 @@ def forward(self, g, inputs):
                     # tensor. This tensor won't be used for computing embeddings.
                     # We need this just to fulfill the requirements of DGL message passing
                     # modules.
-                    assert not self.self_loop, \
-                            f"We cannot allow self-loop if node {k} doesn't have input features."
+                    if g.num_dst_nodes(k) > 0:
```
classicsong, line 171: In which case g.num_dst_nodes(k) == 0?
If dst node is zero, why there is an MFG?

classicsong (approved):

python/graphstorm/model/rgcn_encoder.py
zheng-da: I'm not sure why. I got this issue when running training on the mag dataset.

zhjwy9343: If we remove this assertion, will there be any potential error?

PR #473, opened 2023-09-25
Support load conf for min/max transformation.

classicsong:
*Issue #, if available:
Feature transformation of inference graphs should follow the same transformation rule as the corresponding training graph. However, for min/max transformation, the rule is not saved. Thus when processing inference graphs, gconstruct can not load the transformation info.

*Description of changes:*
Fix the issue.


zhjwy9343 (approved):
LGTM

python/graphstorm/gconstruct/transform.py
```diff
@@ -454,11 +454,21 @@ class NumericalMinMaxTransform(TwoPhaseFeatTransform):
     out_dtype:
         The dtype of the transformed feature.
         Default: None, we will not do data type casting.
+    transform_conf : dict
```
zheng-da, line 457: i suggest to pass min_val and max_val directly. in this way, the doc string can explain what are these two values and what is their type.

python/graphstorm/gconstruct/transform.py
```diff
@@ -448,17 +449,29 @@ class NumericalMinMaxTransform(TwoPhaseFeatTransform):
     feat_name : str
         The feature name used in the constructed graph.
     max_bound : float
-        The maximum float value.
+        The maximum float value. Any number larger than max_bound will be set to max_bound.
     min_bound : float
-        The minimum float value
+        The minimum float value. Any number smaller than min_bound will be set to min_bound.
+    max_val : list of float
+        Define the value of `max` in the Min-Max normalization formula. When max_val is set
```
zheng-da, line 456:
```suggestion
        Define the value of `max` in the Min-Max normalization formula for each feature. If max_val is set
```

python/graphstorm/gconstruct/transform.py
```diff
@@ -448,17 +449,29 @@ class NumericalMinMaxTransform(TwoPhaseFeatTransform):
...
+        max_bound will be ignored.
+    min_val : list of float
+        Define the value of `min` in the Min-Max normalization formula. When min_val is set
```
zheng-da, line 459:
```suggestion
        Define the value of `min` in the Min-Max normalization formula for each feature. If min_val is set
```

python/graphstorm/gconstruct/transform.py
```diff
@@ -525,6 +551,11 @@ def update_info(self, info):
         self._max_val = max_val
         self._min_val = min_val
 
+        # We need to save the max_val and min_val in the config object.
+        if self._conf is not None:
+            self._conf['max_val'] = self._max_val.tolist()
+            self._conf['min_val'] = self._min_val.tolist()
```
zheng-da, line 557: was there a bug before? it didn't change the max_val and min_val in conf.

classicsong: Previously, max_val and min_val were not saved back into conf.

python/graphstorm/gconstruct/transform.py
classicsong: Updated

docs/source/configuration/configuration-gconstruction.rst
```diff
@@ -87,7 +87,7 @@ Currently, the graph construction pipeline supports the following feature transf
...
+* **Numerical MAX_MIN transformation** normalizes numerical input features with `val = (val-min)/(max-min)`, where `val` is the feature value, `max` is the maximum number in the feature and `min` is the minimum number in the feature. The ``name`` field in the feature transformation dictionary is ``max_min_norm``. The dict can contains four optional fields. ``max_bound`` specifies the maximum value allowed in the feature. Any number larger than ``max_bound`` will be set to ``max_bound``. Here, `max` = min(np.amax(feats), ``max_bound``). ``min_bound`` specifies the minimum value allowed in the feature. Any number smaller than ``min_bound`` will be set to ``min_bound``. Here, `min` = max(np.amin(feats), ``min_bound``). ``max_val`` defines the `max` in the transformation formular. When ``max_val`` is provided, `max` always equal to ``max_val``. ``min_val`` defines the `min` in the transformation formular.  When ``min_val`` is provided, `min` always equal to ``min_val``.
```
zheng-da, line 90:
```suggestion
* **Numerical MAX_MIN transformation** normalizes numerical input features with `val = (val-min)/(max-min)`, where `val` is the feature value, `max` is the maximum number in the feature and `min` is the minimum number in the feature. The ``name`` field in the feature transformation dictionary is ``max_min_norm``. The dict can contain four optional fields: ``max_bound``, ``min_bound``, ``max_val`` and ``min_val``. ``max_bound`` specifies the maximum value allowed in the feature. Any number larger than ``max_bound`` will be set to ``max_bound``. Here, `max` = min(np.amax(feats), ``max_bound``). ``min_bound`` specifies the minimum value allowed in the feature. Any number smaller than ``min_bound`` will be set to ``min_bound``. Here, `min` = max(np.amin(feats), ``min_bound``). ``max_val`` defines the `max` in the transformation formula. When ``max_val`` is provided, `max` is always equal to ``max_val``. ``min_val`` defines the `min` in the transformation formula.  When ``min_val`` is provided, `min` is always equal to ``min_val``. ``max_val`` and ``min_val`` are mainly used in the inference stage, where we want to use the max & min values computed in the training stage to normalize inference data.
```

zheng-da (approved):

PR #477, opened 2023-09-25
[Bug Fix] Disable storing embeddings when inferring on ec/er task

jalencato:
*Issue #, if available:*

*Description of changes:*

Disable storing embeddings when inferring on ec/er task


jalencato:
> Can you check if node classification has the same issue?
Nope, for the node classification, we calculate the embeddings on the mini batch only. So there will no such issue. On the link prediction, we calculate them as well. So it is the only issue for the edge task.

python/graphstorm/inference/ep_infer.py
```diff
@@ -77,6 +77,9 @@ def infer(self, loader, save_embed_path, save_prediction_path=None,
                 "A label field must be provided for edge classification " \
                 "or regression inference when evaluation is required."
 
+        if use_mini_batch_infer:
+            assert save_embed_path is None, \
+                "Not allowed to save embedding when using mini batch inference"
```
classicsong, line 82:
```suggestion
                "Unable to save the node embeddings when using mini batch inference." \
                "It is not guaranteed that mini-batch prediction will cover all the nodes."
```

classicsong (approved):

PR #479, opened 2023-09-25
Use zfill to make sure the saved embeddings or prediction results are sortable.

classicsong:
*Issue #, if available:*
The saved embeddings or prediction results are named as emb-0.pt, emb-1.pt, .... emb-10.pt ... When there are more than 10 workers in the graph, the output files can not be easily sorted.

*Description of changes:*
We pad file index with extra zeros, so that the saved files will be named as emb-00000.pt, emb-00001.pt, etc.


zhjwy9343:
Could you add descriptions for this PR?

zhjwy9343 (approved):
LGTM

PR #481, opened 2023-09-26
Change save prediction logic of edge prediction tasks

classicsong:
Originally, we save prediction results only. However, unlink node prediction tasks, it is impossible to find the corresponding edges for the edge prediction results. In this PR, other than storing the prediction results, we also store the source node ids and destination node ids of the corresponding target edges. So users can map a prediction result back to the right edge according to its source node id and its destination node id.

Second PR of #508 


python/graphstorm/model/utils.py
```diff
@@ -703,8 +703,85 @@ def shuffle_predict(predictions, id_mapping_file, pred_type,
...
+            graph partition algorithm.
+        rank: int
+            Rank of the current process in a distributed environment.
+    """
+    num_nodes = g.num_nodes(ntype)
+    id_mapping_info = create_dist_tensor((num_nodes,), dtype=th.int64,
+                                          name=f"mapping-{ntype}",
+                                          part_policy=g.get_node_partition_policy(ntype),
+                                          # TODO: this makes the tensor persistent in memory.
+                                          persistent=True)
```
zheng-da, line 728: let's not turn on persistent here. are we sure that there is still a bug in DistDGL?

python/graphstorm/model/utils.py
```diff
@@ -703,8 +703,85 @@ def shuffle_predict(predictions, id_mapping_file, pred_type,
                 len(predictions)).cpu() # predictions are stored in CPU
     return predictions[local_id_mapping]
 
+def shuffle_nids(g, ntype, nids, node_id_mapping_file, rank):
```
zheng-da, line 706: should we make it stateful? currently, we create a distributed tensor, which is already stateful.

python/graphstorm/model/utils.py
```diff
@@ -703,8 +703,85 @@ def shuffle_predict(predictions, id_mapping_file, pred_type,
...
+        id_mapping = id_mappings[ntype] if isinstance(id_mappings, dict) else id_mappings
+        assert id_mapping.shape[0] == num_nodes, \
+            "id mapping should have the same size of num_nodes"
+        # Save ID mapping into dist tensor
+        id_mapping_info[th.arange(num_nodes)] = id_mapping
+    barrier()
+
+    return id_mapping_info[nids]
+
+def save_ep_prediction_results(predictions, src_nids, dst_nids,
```
zheng-da, line 744: do you mean `save_edge_prediction_results`? ep is already edge prediction.

python/graphstorm/model/utils.py
classicsong: Done.

python/graphstorm/model/utils.py
classicsong: Yes, this is save_edge_prediction. let me change the name.

python/graphstorm/model/utils.py
classicsong: As we need rank0 to load the mapping file (we want to avoid loading mapping files every each inferer), we need to create a distributed tensor.

zheng-da: i understand. what i mean is that the function is exactly stateful because the distributed tensor is persistent. to avoid creating the distributed tensor and fill the id map to it again and again. i think it's better to make it stateful. in this way, we can control when we should destroy the distributed tensors.

classicsong: Updated

python/graphstorm/model/utils.py
```diff
@@ -703,8 +703,109 @@ def shuffle_predict(predictions, id_mapping_file, pred_type,
...
+            and os.path.exists(node_id_mapping_file), \
+            f"{node_id_mapping_file} must exist."
+
+        # _id_mapping_info stores node ids in distributed tensors.
+        # ID mappings are not load in __init__ function
+        # The mappings are loaded on-demand.
+        self._id_mapping_info = {}
+        # _id_mappings is used by rank0 to cache the node id mapping file.
+        if get_rank() == 0:
+            self._id_mappings = th.load(node_id_mapping_file)
```
zheng-da, line 729: this is saved in rank 0 process and the data are stored here whenever the object exists. should we save the data in `self._id_mapping_info` directly and avoid storing `self._id_mappings`?

classicsong: Storing data in self._id_mapping_info requires storing data as Dist Tensors, i.e., on shared memory, while holding self._id_mappings only consumes non shared memory.
This implementation try to reduce the overall cost of shared memory, as in some cases, not all the node id mappings are required.

classicsong: > this is saved in rank 0 process and the data are stored here whenever the object exists. should we save the data in `self._id_mapping_info` directly and avoid storing `self._id_mappings`?

Storing data in self._id_mapping_info requires storing data as Dist Tensors, i.e., on shared memory, while holding self._id_mappings only consumes non shared memory.
This implementation try to reduce the overall cost of shared memory, as in some cases, not all the node id mappings are required.

zheng-da: at the cost of having one process to store the data of all nodes? i'm not sure if it will make it more scalable.

zheng-da: i see some benefits if there is only one machine. otherwise, storing data in shared memory should be better.

classicsong: Updated

python/graphstorm/model/utils.py
```diff
@@ -703,8 +703,111 @@ def shuffle_predict(predictions, id_mapping_file, pred_type,
...
+
+        ntypes = ntypes if ntypes is not None else g.ntypes
+        assert isinstance(ntypes, list) and len(ntypes) > 0, \
+            f"ntypes is not a list or is an empty list {ntypes}"
+        id_mappings = th.load(node_id_mapping_file) if get_rank() == 0 else None
+
+        self._id_mapping_info = {
+            ntype: self._load_id_mapping(g, ntype, id_mappings) \
+                for ntype in ntypes
+        }
```
zheng-da, line 733: you need to test `ntypes` here.

python/graphstorm/model/utils.py
```diff
@@ -703,8 +703,111 @@ def shuffle_predict(predictions, id_mapping_file, pred_type,
...
+        ----------
+        g: Dist DGLGraph
+            Graph.
+        node_id_mapping_file:
+            Path to the file storing node id mapping generated by the
+            graph partition algorithm.
+        ntypes: list of str
+            The node types that will have node ids shuffled.
+    """
+    def __init__(self, g, node_id_mapping_file, ntypes=None):
```
zheng-da, line 719: i think users need to provide `ntypes`. otherwise, the code probably doesn't work.

zheng-da (approved):

PR #482, opened 2023-09-26
Fix a bug in restoring models in the inference script

zheng-da:
*Description of changes:*
We should allow users to customize the model layers to load in the inference script.


classicsong (approved):
LGTM

PR #484, opened 2023-09-26
Enable nccl for embedding save

classicsong:
*Issue #, if available:*

*Description of changes:*


python/graphstorm/model/utils.py
```diff
@@ -342,7 +345,8 @@ def _exchange_node_id_mapping(rank, world_size, device,
         alltoallv_cpu(rank, world_size, gather_list, data_tensors)
     else: # backend == "nccl"
         th.distributed.all_to_all(gather_list, data_tensors)
```
isratnisa, line 347: Can we create a wrapper in the utils.py?

isratnisa (approved):

python/graphstorm/model/utils.py
classicsong: Yes.

PR #486, opened 2023-09-26
Remove unnecessary caching

zheng-da:
*Description of changes:*
We don't need to cache the BERT embeddings and the input embeddings twice.


classicsong (approved):

zheng-da:
i decide to use a different way to avoid caching twice.

PR #490, opened 2023-09-26
[GSProcessing] Name reverse edges as `dst:relation-rev:src` 

thvasilo:
*Issue #, if available:*

*Description of changes:*

* We use `dst:relation-rev:src`  for reverse edge names instead of `dst:rev-relation:src` to match GConstruct.

*Testing*:

* pytest and test job (with re-partition follow-up) succeed


classicsong (approved):
LGTM

PR #495, opened 2023-09-27
[Bug fix]Fix hgt_encoder no edge error

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
This PR fixed the error when no edge exists in a block, but the relation type still exist.

The previous code assign `edge_fn` to all relation types in a mini-batch. But some mini-batchs could have the case where there is no edge for a specific relation type. In such case, the that relation type has an `edge_fn`, but no nodes and edges. When call `multi_update_all()` function, that relation type will fail due to no node or edge data.

The new implementation will filter up such no-edge relation type, and only assign `edge_fn` to relation types having edges. Therefore fix this issue.

### DIFF
--- a/python/graphstorm/model/hgt_encoder.py
+++ b/python/graphstorm/model/hgt_encoder.py
@@ -189,6 +189,7 @@ class HGTLayer(nn.Module):
         """
         # pylint: disable=no-member
         with g.local_scope():
+            edge_fn = {}
             for srctype, etype, dsttype in g.canonical_etypes:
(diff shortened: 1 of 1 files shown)


zheng-da:
why this can fix the bug?

zhjwy9343:
> why this can fix the bug?

If there is no edges for a specific relation type, this fix will not create an edge-type aggregation function. Therefore this type of edge will not be used in `multi_update_all()`. 

The previous code always set an aggregation function no matter if the `num_edges()==0`.

zheng-da (approved):

PR #496, opened 2023-09-27
Fix lint does not work with astroid==3.0 error.

classicsong:
`python3 -m pip install --upgrade prospector pip` has a change to install astroid==3.0 (do not know why) which is not compatible with lint.

Fix astroid version to 2.15.7 in this PR.


jalencato (approved):
LGTM. I have tested on my branch after deploying the changes. Never see the problem.

PR #483, opened 2023-09-26
Glem restore bug fix

wangz10:
*Issue #, if available:*

*Description of changes:* Modify the `GLEM.restore_dense_model` and `GLEM.restore_sparse_model` such that it loads the correct layers for LM and GNN from either GLEM saved checkpoints or `GSgnnNodePredictionTrainer` saved checkpoints.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #498, opened 2023-09-27
Glem full graph inference

wangz10:
*Issue #, if available:*

*Description of changes:* Enabled in full-graph inference for GLEM model. Added integration test to ensure the inference results are consistent between mini-batch and full-graph inference.


wangz10:
Should be ready after merging #483

python/graphstorm/inference/np_infer.py
```diff
@@ -96,9 +97,15 @@ def infer(self, loader, save_embed_path, save_prediction_path=None,
             else:
                 embs = {ntype: embs}
         else:
-            embs = do_full_graph_inference(self._model, loader.data, fanout=loader.fanout,
+            if isinstance(self._model, GLEM):
+                embedding_model = self._model if self._model.inference_using_gnn else self._model.lm
+                decoding_model = self._model.gnn if self._model.inference_using_gnn else self._model.lm
```
zheng-da, line 102: instead of changing the code here, can we have GLEM to return the right layers based on `inference_using_gnn`?

zheng-da: basically, we need to define `node_input_encoder`, `gnn_encoder`, `decoder` in GLEM correctly.

python/graphstorm/model/gnn.py
```diff
@@ -942,7 +942,8 @@ def do_full_graph_inference(model, data, batch_size=1024, fanout=None, edge_mask
     -------
     dict of th.Tensor : node embeddings.
     """
-    assert isinstance(model, GSgnnModel), "Only GSgnnModel supports full-graph inference."
+    assert isinstance(model, GSgnnModel) or type(model).__name__ == 'GLEM',\
```
zheng-da, line 945: why do we need the assert? GLEM is a child class of `GSgnnNodeModelBase`

python/graphstorm/model/node_glem.py
```diff
@@ -128,6 +128,21 @@ def gnn_encoder(self):
...
+
+    @property
+    def _node_input_encoder(self):
+        """Alias for accessing the node_input_encoder"""
+        return self.lm.node_input_encoder
+
+    @property
+    def decoder(self):
+        """Alias for accessing the decoder"""
+        return self.gnn.decoder if self.inference_using_gnn else self.lm.decoder
```
zheng-da, line 144: you already define these. why do you still need the code above?

python/graphstorm/inference/np_infer.py
wangz10: Addressed.

python/graphstorm/model/gnn.py
wangz10: GLEM is not a subclass of `GSgnnModel`.

python/graphstorm/model/node_glem.py
wangz10: They need to be accessed in other functions in `embed.py`, `do_full_graph_inference`, etc.

python/graphstorm/model/gnn.py
zheng-da: it is not ideal to explicitly check if it's GLEM. let me check which class we should assert.

python/graphstorm/model/node_glem.py
zheng-da: i understand. what i mean is that we should redefine these properties and we don't need the modifications in np_trainer.py or np_infer.py

python/graphstorm/model/gnn.py
zheng-da: let's have it like this for now.

python/graphstorm/model/node_glem.py
```diff
@@ -125,8 +143,23 @@ def set_gnn_encoder(self, gnn_encoder):
...
+        """Alias for accessing the gnn_encoder. Hide gnn_encoder if the inference route is lm"""
+        return self.gnn.gnn_encoder if self.inference_route_is_gnn else None
+
+    @property
+    def node_input_encoder(self):
+        """Alias for accessing the node_input_encoder"""
+        return self.lm.node_input_encoder
+
+    @property
+    def _node_input_encoder(self):
```
zheng-da, line 155: why do we have this property?

python/graphstorm/model/node_glem.py
```diff
@@ -125,8 +143,23 @@ def set_gnn_encoder(self, gnn_encoder):
 
     @property
     def gnn_encoder(self):
```
zheng-da, line 145: can you also add comment that these properties are only used for model inference and evaluation?

python/graphstorm/model/node_glem.py
zheng-da: i don't see this property is used. Can you remove it?

wangz10: removed

zheng-da (approved):

PR #505, opened 2023-09-28
[Doc] Improving instruction for a distributed setting

DominikaJedynak:
This PR fixes some small issues in the instructions on how to setup a (distributed) environment for GraphStorm. It: 

- changes some links so that they point to GraphStorm's docs instead of to deprecated wiki,
- changes/adds some links so that they point to more accurate parts of the instruction,
- adds the tip how to avoid having to preface all Docker commands with _sudo_ (which is the case by default on some AWS instances),
- adds _-o StrictHostKeyChecking=no_ flag for checking the connectivity between containers (which is also used for connecting in _python/graphstorm/run/launch.py_ file and without which it seems impossible to connect following the current instruction),
- adds a step of exchanging the ssh keys between Docker containers

### DIFF
--- a/docs/source/scale/distributed.rst
+++ b/docs/source/scale/distributed.rst
@@ -70,7 +70,7 @@ In the container environment, users can check the connectivity with the command 
 
 .. code-block:: bash
 
-    ssh 172.38.12.143 -p 2222
+    ssh 172.38.12.143 -o StrictHostKeyChecking=no -p 2222
 
 If succeeds, you should login to the container in the ``<ip-in-the-cluster>`` instance. 
 
--- a/docs/source/scale/distributed.rst
+++ b/docs/source/scale/distributed.rst
@@ -16,7 +16,7 @@ Create a GraphStorm Cluster
 
 Setup the instance of a cluster
 .......................................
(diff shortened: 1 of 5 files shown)


docs/source/scale/distributed.rst
```diff
@@ -66,15 +66,16 @@ Pick one instance and run the following command to connect to the GraphStorm Doc
...
 
 .. code-block:: bash
 
-    ssh 172.38.12.143 -p 2222
+    ssh 172.38.12.143 -o StrictHostKeyChecking=no -p 2222
 
 If succeeds, you should login to the container in the ``<ip-in-the-cluster>`` instance. 
 
 If not, please make sure there is no limitation of port 2222. 
+You may also have to exchange the keys from inside the GraphStorm Docker containers to allow their communication. For that, copy the keys from the ``/root/.ssh/id_rsa.pub`` from this container to ``/root/.ssh/authorized_keys`` in containers on every machine in your cluster.
```
DominikaJedynak, line 78: Maybe key exchanging should be a separate step of the instruction? Is there a plan of GraphStorm Docker images to be available for download instead of building them manually? With the current instructions, it seems that every container has different keys as keys are created inside Dockerfile for every machine separately.

zhjwy9343: So far we have our own internal Docker images, but not publicly available.

Thanks for the new explanation added.

Issue #508, opened 2023-09-28
Update the inference output format


zheng-da:
Currently, the node prediction and edge prediction inference scripts save the embeddings of all nodes or edges. It is very common that we only need to inference the embeddings of a small number of nodes and edges. We should define a compact output format to handle this case.

One possible output format is that we output two files:
* the original node/edge IDs,
* the corresponding node/edge embeddings and predictions.

PR #524, opened 2023-10-03
Update save prediction

classicsong:
Currently, the node task and edge task inference scripts can only save prediction results for one node or edge type. This PR makes it more general that the scripts can save prediction results from multiple node types and edge types.

First PR of #508 


python/graphstorm/inference/ep_infer.py
```diff
@@ -136,20 +136,22 @@ def infer(self, loader, save_embed_path, save_prediction_path=None,
...
-                # edges that have predictions may be just a subset of the
-                # entire edge set.
-                pred_data[loader.target_eidx[etype]] = pred.cpu()
-
-                pred = shuffle_predict(pred_data, edge_id_mapping_file, etype, get_rank(),
-                                       get_world_size(), device=device)
-            save_prediction_results(pred, save_prediction_path, get_rank())
+                shuffled_preds = {}
+                for etype, pred in preds.items():
+                    assert etype in infer_data.eval_etypes
```
zheng-da, line 141: print an error message?

python/graphstorm/inference/ep_infer.py
```diff
@@ -136,20 +136,22 @@ def infer(self, loader, save_embed_path, save_prediction_path=None,
...
+                    pred_shape = list(pred.shape)
+                    pred_shape[0] = g.num_edges(etype)
+                    pred_data = create_dist_tensor(pred_shape, dtype=pred.dtype,
+                        name='predict-'+'-'.join(etype),
+                        part_policy=g.get_edge_partition_policy(etype),
+                        # TODO: this makes the tensor persistent in memory.
+                        persistent=True)
```
zheng-da, line 148: can we remove this? this will keep the distributed tensor persistent even if we will never use the tensor again.

python/graphstorm/model/utils.py
```diff
@@ -681,6 +688,57 @@ def save_prediction_results(predictions, prediction_path, rank):
...
+        save_prediction_results(pred,
+                                os.path.join(prediction_path, ntype),
+                                rank)
+    if rank == 0:
+        meta_fname = os.path.join(prediction_path, "result_info.json")
+        meta_info = {
+            "format": "pytorch",
+            "world_size": world_size,
+            "ntypes": list(predictions.keys())
+        }
```
zheng-da, line 713: why is this useful?

zheng-da:
do we currently save the embeddings in the same format as saving the predictions?

classicsong:
> do we currently save the embeddings in the same format as saving the predictions?

Here is the PR for saving the embeddings in the same format as saving predictions: https://github.com/awslabs/graphstorm/pull/527

python/graphstorm/inference/ep_infer.py
classicsong: We can. But I am not sure whether DGL solved the bug of GC a DistTensor.

python/graphstorm/model/utils.py
classicsong: format and world_size are useful for remapping the id into raw ID space.
ntypes is recording which node type(s) have prediction results. the remapping script can use it to do some sanity checks.

python/graphstorm/inference/ep_infer.py
classicsong: I have changed it into False

zheng-da: we should test it. if it still doesn't work, we need to inform the DGL team to fix it.

zheng-da (approved):

zheng-da:
do we need to finalize the format?

PR #530, opened 2023-10-05
[Documentation] Standalone mode demo notebook

wangz10:
*Issue #, if available:*

*Description of changes:* Added Jupyter notebook to demonstrate standalone mode.


examples/standalone_mode_demo.ipynb
```diff
@@ -0,0 +1,609 @@
...
+   "metadata": {},
+   "source": [
+    "## Develop a GNN model for node-classification \n",
+    "\n",
+    "Next, we use a GNN model to go over the model development cycle of a node classification problem on the MovieLens dataset. \n",
+    "\n",
+    "\n",
+    "### 0. Load the constructed graph data\n",
+    "\n",
+    "We can use the following codes to set up the `GSConfig` object by providing the yaml config file of the training script and the json config file about the constructed graph dataset. `GSConfig` handles the parsing of all the config files normally provided through command line. Therefore, we can simulate this in the Notebook using `argparse.Namespace`. "
```
zheng-da, line 83: let's remove GSConfig. when people develop custom models, they don't have to use GSConfig to set up the hyperparameters. GSConfig is used for the command line tool.

examples/standalone_mode_demo.ipynb
```diff
@@ -0,0 +1,609 @@
...
+    "        'RANK': '0',\n",
+    "        'WORLD_SIZE': '1',\n",
+    "        'MASTER_ADDR': '127.0.0.1',\n",
+    "        'MASTER_PORT': '54321'\n",
+    "    }\n",
+    "    # set up required ENV_VAR for distributed context\n",
+    "    for key, val in env_vars.items():\n",
+    "        os.environ[key] = val\n",
+    "\n",
+    "    gs.initialize(ip_config=env_vars['MASTER_ADDR'], backend='gloo')\n",
```
zheng-da, line 65: `env_vars` is not necessary. you can set `ip_conifg=None`.

examples/standalone_mode_demo.ipynb
```diff
@@ -0,0 +1,609 @@
...
+    "# constructed graphstorm data config:\n",
+    "constructed_graph_config = '/mnt/efs/gsf-data/movielen_100k_train_val_1p_4t/movie-lens-100k.json'\n",
+    "\n",
+    "cmd_args = {\n",
+    "    'yaml_config_file': yaml_config_file,\n",
+    "    'part_config': constructed_graph_config,\n",
+    "    'local_rank': os.environ['RANK'],\n",
+    "}\n",
+    "cmd_args = Namespace(**cmd_args)\n",
+    "config = GSConfig(cmd_args)"
```
zheng-da, line 103: let's not use GSConfig.

examples/standalone_mode_demo.ipynb
```diff
@@ -0,0 +1,609 @@
...
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "# create a built-in gsf model\n",
+    "model = gs.create_builtin_node_gnn_model(\n",
+    "    train_data.g, \n",
+    "    config, \n",
+    "    train_task=True)\n",
```
zheng-da, line 258: Here we should show how to define the model architecture.

examples/standalone_mode_demo.ipynb
```diff
@@ -0,0 +1,609 @@
...
+   "cell_type": "code",
+   "execution_count": 12,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Extract accuracies from the trainer's evaluator:\n",
+    "val_accs, test_accs = [], []\n",
+    "for val_acc, test_acc in trainer.evaluator._history:\n",
+    "    val_accs.append(val_acc['accuracy'])\n",
+    "    test_accs.append(test_acc['accuracy'])"
```
zheng-da, line 391: is it important to expose this to users? if so, I suggest to add `history` property.

examples/standalone_mode_demo.ipynb
zhjwy9343: One big challenge of removing GSConfig comes from those settings to be used in model generation. If we want to use gs.create_builtin_node_gnn_model(), we need to pass a GSConfig object as an argument. If we do not use .create_builtin_node_gnn_model() method, then we need to expose the details of each layer, which will make this demo much complex.

zheng-da: when i designed the API, i didn't intend to expose `create_builtin_node_gnn_model`. i wrote that as a helper function.
if we expose GSConfig, users need to learn the entire GSConfig. i wonder if this is necessary.

examples/standalone_mode_demo.ipynb
```diff
@@ -0,0 +1,609 @@
...
+    "    batch_size=config.batch_size,\n",
+    "    device=device, train_task=False)"
+   ]
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### 2. Create a GNN model\n",
```
zhjwy9343, line 190: Could you try to use the code in the `gs.create_builtin_node_gnn_model()` to build a model, and check what GSConfig arguments will be used? I think this is the place where many GSConfig arguments are used, and need to be replaced.

Also, we expose the encoder and gnn layer APIs upon which users can use to build their own customer models. In such case, we better reveal the details of create_builtin_node_gnn_model().

examples/standalone_mode_demo.ipynb
zhjwy9343: Agree, I added a new comment, asking to use details in the `create_builtin_node_gnn_model` to build a model. We didn't expose the `create_builtin_node_gnn_model` API.

examples/standalone_mode_demo.ipynb
```diff
@@ -0,0 +1,609 @@
...
+     "output_type": "stream",
+     "text": [
+      "Number of model parameters: 84480\n",
+      "Number of trainable parameters: 84480\n"
+     ]
+    }
+   ],
+   "source": [
+    "num_params = sum(p.numel() for p in model.parameters())\n",
+    "num_trainable_params = sum(p.numel() for p in model.parameters() if p.requires_grad)\n",
```
zhjwy9343, line 278: Because the results of the two variables are the same, which is 84480, readers may ask what is the difference between the two variables. Do we have an example having different values for the two variables?

examples/standalone_mode_demo.ipynb
```diff
@@ -0,0 +1,609 @@
...
+      "INFO:root:Epoch 8 take 0.110 seconds\n",
+      "INFO:root:Part 0 | Epoch 00009 | Batch 000 | Loss: 0.4196 | Time: 0.0102\n",
+      "INFO:root:Epoch 9 take 0.109 seconds\n",
+      "INFO:root:Peak GPU Mem alloc: 20.3286 MB\n"
+     ]
+    }
+   ],
+   "source": [
+    "# Train the model with the trainer\n",
+    "trainer.fit(\n",
```
zhjwy9343, line 364: If you can set up a task_tracker for the trainer, you will see the evaluation and testing results during the training epochs. Then no need to call the `trainer.evaluator._history` internal attribute. We didn't expose this attribute.

examples/standalone_mode_demo.ipynb
```diff
@@ -0,0 +1,609 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "total 16K\n",
+      "-rw-rw-r-- 1 ubuntu ubuntu 14K Oct  5 20:47 predict-00000.pt\n"
+     ]
+    }
+   ],
+   "source": [
```
zhjwy9343, line 582: We are changing the output of prediction. Please talk to Xiang for the new output format.

examples/standalone_mode_demo.ipynb
```diff
@@ -0,0 +1,609 @@
...
+    "infer_dataloader = GSgnnNodeDataLoader(\n",
+    "    infer_data, infer_data.test_idxs, fanout=config.fanout, \n",
+    "    batch_size=100, device=device,\n",
+    "    train_task=False)\n",
+    "\n",
+    "\n",
+    "# Create an Inferrer object\n",
+    "infer = GSgnnNodePredictionInferrer(model)\n",
+    "infer.setup_device(device=device)\n",
+    "infer.setup_evaluator(evaluator)\n"
```
zhjwy9343, line 524: Also set task_tracker to show testing results during inference.

examples/standalone_mode_demo.ipynb
wangz10: task_tracker is not exposed API. Also, setting up task_tracker requires the `GSConfig` again. We can use task_tracker   if initializing doesn't require GSConfig.

examples/standalone_mode_demo.ipynb
wangz10: https://github.com/awslabs/graphstorm/pull/530#discussion_r1350650746

examples/standalone_mode_demo.ipynb
wangz10: We could do this, but it will leads to this bug: https://github.com/awslabs/graphstorm/issues/535
Will switch to the following after bug fix.

examples/standalone_mode_demo.ipynb
wangz10: The new prediction output will be updated.

zheng-da (approved):

PR #533, opened 2023-10-07
Add support for remapping edge prediction results.

classicsong:
Add graphstrom.gconstruct.remap_result.py to remap edge prediction results.
Will add support for remapping node prediction results and embeddings.

#508 


python/graphstorm/gconstruct/id_map.py
```diff
@@ -66,6 +70,64 @@ def save(self, file_path):
...
+        tensor: A tensor of raw IDs.
+        """
+        if len(ids) == 0:
+            return np.array([], dtype=np.str)
+
+        return self._ids[ids]
+
+    @property
+    def size(self):
+        "Get the approximated szie of the id mapping"
```
zhjwy9343, line 128: "szie" -> "size".

classicsong: We do not need it now.

classicsong: I will remove it.

tests/end2end-tests/graphstorm-ec/mgpu_test.sh
```diff
@@ -129,7 +129,7 @@ then
 fi
 
 cnt=$(ls -l /data/gsgnn_ec/prediction/user_rating_movie/ | grep predict | wc -l)
-if test $cnt != $NUM_INFO_TRAINERS
+if test $cnt != $NUM_INFO_TRAINERS * 2
```
zheng-da, line 132: why do we have this change?

classicsong: If we set --preserve-input True, then we have both old prediction result and new prediction result under /data/gsgnn_ec/prediction/user_rating_movie/

python/graphstorm/gconstruct/remap_result.py
```diff
@@ -0,0 +1,477 @@
...
+    """ Do one edge remapping task
+
+        Parameters
+        ----------
+        pred_file_path: str
+            The path to the prediction result.
+        src_nid_path: str
+            The path to the file storing source node ids
+        dst_nid_path: str
+            The path to the file storing destination node ids
```
zheng-da, line 58: should we store them in separate files?

python/graphstorm/gconstruct/remap_result.py
```diff
@@ -0,0 +1,477 @@
...
+        chunk_size: int
+            Max number of raws per output file.
+        preserve_input: bool
+            Whether the input data should be removed.
+    """
+    pred_result = th.load(pred_file_path).numpy()
+    src_nids = th.load(src_nid_path).numpy()
+    dst_nids = th.load(dst_nid_path).numpy()
+    src_id_map = id_maps[src_type]
+    dst_id_map = id_maps[dst_type]
```
zheng-da, line 74: id_maps is a global variable?

zheng-da: if so, why not make it a class?

python/graphstorm/gconstruct/remap_result.py
```diff
@@ -0,0 +1,477 @@
...
+        data = {"pred": pred,
+                "src_nid": src_nid,
+                "dst_nid": dst_nid}
+
+        write_data_parquet(data, output_fname)
+
+    if preserve_input is False:
+        os.remove(pred_file_path)
+        os.remove(src_nid_path)
+        os.remove(dst_nid_path)
```
zheng-da, line 93: do we need to remove the files directly?

python/graphstorm/gconstruct/remap_result.py
```diff
@@ -0,0 +1,477 @@
...
+    try:
+        while True:
+            # If the queue is empty, it will raise the Empty exception.
+            idx, task_args = task_queue.get_nowait()
+            logging.debug("worker %d Processing %s task", worker_id, idx)
+            func(**task_args)
+    except queue.Empty:
+        pass
+
+def multiprocessing_remap(tasks, num_proc, remap_func):
```
zheng-da, line 116: should we make this a standardized utility function for multiprocessing?

python/graphstorm/gconstruct/remap_result.py
```diff
@@ -0,0 +1,477 @@
...
+        #    src_nids-00000.pt
+        #    src_nids-00001.pt
+        #    ...
+        #    dst_nids-00000.pt
+        #    dst_nids-00001.pt
+        #    ...
+
+        The output emb files will be
+        #    predict-00000_00000.parquet
+        #    predict-00000_00001.parquet
```
zheng-da, line 207: why the remapped data are stored in a different format?

python/graphstorm/gconstruct/remap_result.py
```diff
@@ -0,0 +1,477 @@
...
+        out_pred_dir = os.path.join(output_dir, "_".join(etype))
+        etype_pred_files = os.listdir(input_pred_dir)
+        src_nid_files = [fname for fname in etype_pred_files if fname.startswith("src_nids")]
+        dst_nid_files = [fname for fname in etype_pred_files if fname.startswith("dst_nids")]
+        pred_files = [fname for fname in etype_pred_files if fname.startswith("predict")]
+        src_nid_files.sort()
+        dst_nid_files.sort()
+        pred_files.sort()
+
+        num_parts = len(pred_files)
```
zheng-da, line 244: should you verify if src_nid_files and dst_nid_files have the same number of files?

python/graphstorm/gconstruct/remap_result.py
```diff
@@ -0,0 +1,477 @@
...
+                       help="The logging level. The possible values: debug, info, warning, \
+                                   error. The default value is info.")
+
+    group.add_argument("--pred-etypes", type=str, nargs="+", default=None,
+                       help="[Optional] A list of canonical edge types which have"
+                                "prediction results For example, "
+                                "--pred-etypes user,rate,movie user,watch,movie"
+                                "If pred_etypes is not provided, result_info.json "
+                                "under prediction_dir will be used to retrive the pred_etypes")
+    group.add_argument("--output-chunk-size", type=int, default=100000,
```
zheng-da, line 466: why do we want to output the remapped results in a different number of files?

python/graphstorm/run/launch.py
```diff
@@ -323,6 +323,58 @@ def construct_torch_dist_launcher_cmd(
...
+
+    udf_command[0] = "graphstorm.gconstruct.remap_result"
+
+    # Add remap related arguments
+    udf_command += ["--rank", str(rank)]
+    udf_command += ["--world-size", str(world_size)]
+    udf_command += ["--with-shared-fs", "True" if with_shared_fs else "False"]
+    udf_command += ["--num-processes", str(num_trainers)]
+    udf_command += ["--output-chunk-size", str(output_chunk_size)]
+    udf_command += ["--preserve-input", "True" if preserve_input else "False"]
```
zheng-da, line 367: why do we give users this option?

python/graphstorm/gconstruct/remap_result.py
```diff
@@ -0,0 +1,477 @@
...
+        src_nid_files = [fname for fname in etype_pred_files if fname.startswith("src_nids")]
+        dst_nid_files = [fname for fname in etype_pred_files if fname.startswith("dst_nids")]
+        pred_files = [fname for fname in etype_pred_files if fname.startswith("predict")]
+        src_nid_files.sort()
+        dst_nid_files.sort()
+        pred_files.sort()
+
+        num_parts = len(pred_files)
+        logging.debug("%s has %d embedding files", etype, num_parts)
+        if with_shared_fs:
```
zheng-da, line 246: i suppose only this option has been tested.

zheng-da: did you test the other option?
if not, maybe we can remove that option first.

python/graphstorm/gconstruct/remap_result.py
classicsong: We already store them in separate files. src_nids_xxx and dst_nids_xxx.

python/graphstorm/gconstruct/remap_result.py
classicsong: It is a global variable. The purpose is to share it across forked processes with out passing id_maps to other processes (using the copy-on-write mechanism provided by Linux).

python/graphstorm/gconstruct/remap_result.py
classicsong: By default we can remove them. Users only need the remapped data.

python/graphstorm/gconstruct/remap_result.py
classicsong: You mean we move it into utils.py?
In the following PRs, both remap_node_predictions and remap_node_embeddings will use this function.

python/graphstorm/gconstruct/remap_result.py
classicsong: We can hardly store strings in torch tensor.
We supports parquet now. But we can extend it into csv/json or other formats if needed.

python/graphstorm/gconstruct/remap_result.py
classicsong: Good point

python/graphstorm/run/launch.py
classicsong: This is for debug purpose, in case the remap function has a bug.

classicsong: By default, it is False.

python/graphstorm/gconstruct/remap_result.py
classicsong: In following PRs, I will add sagemaker support. with_shared_fs == False will be tested.

classicsong: BTW, let me see if I can write some unitest for it.

classicsong: Add a test in data_process/test.sh

python/graphstorm/gconstruct/remap_result.py
classicsong: Currently the number of prediction result or embedding files are same as the number of trainers. 
When the graph is large, it is possible that a single output file can be tens of GBs or even larger.
With this parameter, we can split those files into smaller files. For example thousands of parquet files, each of which takes only hundreds of MB space.

python/graphstorm/gconstruct/remap_result.py
zheng-da: please write comments to explain this.

python/graphstorm/gconstruct/remap_result.py
zheng-da: these are intermediate files for command line tools, but they are not for customer model API. the users of customer model API and the users of command line tools have very different user experience.

python/graphstorm/gconstruct/remap_result.py
zheng-da: i think so. we might want to generalize it for other purpose as well. for example, gconstruct has a multi-processing function as well.

python/graphstorm/gconstruct/remap_result.py
zheng-da: why can we not store them in torch tensors?

python/graphstorm/gconstruct/remap_result.py
zheng-da: i see. the default behavior should be not splitting the files.

python/graphstorm/gconstruct/remap_result.py
classicsong: Does torch support string tensors?
I think parquet is a general format that can be easily accepted by many downstream applications.

python/graphstorm/gconstruct/remap_result.py
classicsong: If users do not call remap, those files are not removed.

python/graphstorm/gconstruct/remap_result.py
classicsong: OK. I moved it into utils.py as named it multiprocessing_exec_no_return.

Issue #535, opened 2023-10-09
GSgnnTrainer doesn't support saving models in  standalone mode


wangz10:
In `save_model`, it assumes model is DDP wrapped without asserting whether it's in distributed training: 

```
File [/opt/conda/envs/pytorch/lib/python3.10/site-packages/graphstorm/trainer/gsgnn_trainer.py:195](https://vscode-remote+ssh-002dremote-002b52-002e24-002e153-002e167.vscode-resource.vscode-cdn.net/opt/conda/envs/pytorch/lib/python3.10/site-packages/graphstorm/trainer/gsgnn_trainer.py:195), in GSgnnTrainer.save_model(self, model, epoch, i, save_model_path)
    193 barrier()
    194 if save_model_path is not None:
--> 195     assert isinstance(model.module, (GSgnnModel, GSgnnModelBase)), \
    196         "Please make sure the model derives from GSgnnModel or GSgnnModelBase, " \
    197         "which provides a scalable model saving implementation."
    198     save_model_path = self._gen_model_path(save_model_path, epoch, i)
    199     model.module.save_model(save_model_path)

File [/opt/conda/envs/pytorch/lib/python3.10/site-packages/torch/nn/modules/module.py:1614](https://vscode-remote+ssh-002dremote-002b52-002e24-002e153-002e167.vscode-resource.vscode-cdn.net/opt/conda/envs/pytorch/lib/python3.10/site-packages/torch/nn/modules/module.py:1614), in Module.__getattr__(self, name)
   1612     if name in modules:
   1613         return modules[name]
-> 1614 raise AttributeError("'{}' object has no attribute '{}'".format(
   1615     type(self).__name__, name))

AttributeError: 'GSgnnNodeModel' object has no attribute 'module'
```

PR #537, opened 2023-10-09
Bug fix dist model save

wangz10:
*Issue #, if available:* #535

*Description of changes:*


python/graphstorm/trainer/gsgnn_trainer.py
```diff
@@ -192,11 +192,12 @@ def save_model(self, model, epoch, i, save_model_path):
         '''
         barrier()
         if save_model_path is not None:
-            assert isinstance(model.module, (GSgnnModel, GSgnnModelBase)), \
```
zheng-da, line 195: there are integration tests for the standalone mode. Can you trigger model saving in integration test of standalone mode?

zheng-da: here is the integration test. https://github.com/awslabs/graphstorm/blob/main/tests/end2end-tests/graphstorm-nc/test.sh#L30

wangz10: done

python/graphstorm/eval/evaluator.py
```diff
@@ -326,6 +326,11 @@ def best_iter_num(self):
         """
         return self._best_iter
 
+    @property
+    def history(self):
+        """ Evaluation history"""
+        return self._history
```
zheng-da, line 332: maybe revert this change and add this to the tutorial?

zheng-da (approved):

PR #548, opened 2023-10-11
Add support to run random graph partition on Linux instead of SageMaker

classicsong:
*Issue #, if available:*

*Description of changes:*


python/graphstorm/gpartition/dist_partition_graph.py
```diff
@@ -0,0 +1,154 @@
...
+    output_path = args.output_path
+    metadata_file=args.meta_info_path
+    local_output_path = os.path.join(output_path, "tmp")
+
+    part_start = time.time()
+    if args.part_algorithm == "random":
+        partitioner = LocalRandomPartitioner(metadata_file, local_output_path)
+        local_partition_path = partitioner.create_partitions(args.num_parts)
+    else:
+        raise RuntimeError(f"Unknow partition algorithm {args.part_algorighm}")
```
thvasilo, line 108: See comment in `random_partition.py`, let's write this more as 

```
if args.part_algorithm == "random":
    partitioner = RandomPartitionAlgorithm(metadata_file, local_output_path)
elif args.part_algorithm == "range":
    partitioner = RangePartitionAlgorithm(metadata_file, local_output_path)
else:
    raise NotImplementedError()

local_partition_path = partitioner.create_partitions(args.num_parts)
```

We should also have a contract that `partitioner.create_partitions()` is the main common interface between different partition algorithm implementations, and what the expected output is from calling that function, that is, one partition assignment file named this specific way is created in this specific path (and write it in docstring)

python/graphstorm/gpartition/random_partition.py
```diff
@@ -0,0 +1,137 @@
...
+"""
+import os
+import logging
+import json
+
+import numpy as np
+import pyarrow as pa
+import pyarrow.csv as pa_csv
+
+class RandomPartitionAlgorithm():
```
thvasilo, line 26: Do we plan to implement more single-instance partition algorithms? (e.g. range)

If yes, it's worth creating a class hierarchy with a shared interface, i.e. a `PartitionAlgorithm` base class from which `RandomPartitionAlgorithm` inherits, where we define a public `create_partitions(args.num_parts, local_output_path)` method, that's implemented as

```
def create_partitions(self, args.num_parts: int, local_output_path: str) -> str:
    partition_dir = os.path.join(self.local_output_path, "partition_assignment")
    os.makedirs(partition_dir, exist_ok=True)
    self._run_partitioning(args.num_parts, local_output_path)
    self._create_metadata(args.num_parts, local_output_path)

    return partition_dir
```

If we do that then we can probably remove `LocalRandomPartitioner` and only have a hierarchy of `PartitionAlgorithm` classes (any common init code can go to the base `PartitionAlgorithm` class)
that implement the two "private" functions, `_run_partitioning` and `_create_metadata`.

So in `python/graphstorm/gpartition/dist_partition_graph.py` we'd have:


```
if args.part_algorithm == "random":
    partitioner = RandomPartitionAlgorithm(metadata_file)
elif args.part_algorithm == "range":
    partitioner = RangePartitionAlgorithm(metadata_file)
else:
    raise NotImplementedError()

local_partition_path = partitioner.create_partitions(args.num_parts, local_output_path)

[...]
```

Note that we move `local_output_path` from object state to function argument (the same partitioner should be able to output partitions anywhere, so let's remove it from the object's state).

thvasilo (changes requested):
Some suggestions to re-design and make our local partitioners easily extensible in the future.

python/graphstorm/gpartition/random_partition.py
classicsong: SageMakerPartitioner also calls RandomPartitionAlgorithm.run_partitioning.
If we change the API here, we need to change the SageMakerPartitioner.

thvasilo: Agreed, right now we have

```python
        random_part = RandomPartitionAlgorithm(self.local_output_path,
                                               self.metadata,
                                               self.rank)
        partition_dir = random_part.run_partitioning(num_partitions)
```

in `SageMakerPartitioner()._run_partitioning(num_partitions: int)`. What I'm suggesting would change that to:


```python
        random_part = RandomPartitionAlgorithm(
                                               self.metadata,
                                               self.rank)
        partition_dir = random_part.create_partitions(num_partitions, self.local_output_path)
```

Does this make sense?

thvasilo:
Closing in favor of #764

PR #561, opened 2023-10-14
Fix bug in constructing node features from neighbor nodes with only BERT embeddings

zheng-da:
*Description of changes:*
When constructing node features from neighbor nodes, the sampler reads the node features from the dataset to identify the right neighbor nodes to compute node features. If the features of the neighbor nodes are text features, the current datasets don't know these features. This PR basically changes this behavior. When the BERT embeddings are cached, they are added to the datasets as node features. In this way, the sampler for constructing node features can identify the right neighbor nodes.


classicsong:
Can you provide a short description of the bug?

zheng-da:
I added the description.

python/graphstorm/gsf.py
```diff
@@ -542,6 +542,9 @@ def set_encoder(model, g, config, train_task):
                                           use_node_embeddings=config.use_node_embeddings,
                                           force_no_embeddings=config.construct_feat_ntype,
                                           num_ffn_layers_in_input=config.num_ffn_layers_in_input)
+    # The number of feature dimensions can change. For example, the feature dimensions
+    # of BERT embeddings are determined when the input encoder is created.
+    feat_size = encoder.in_dims
```
classicsong, line 547: It seems feat_size is not used in the following code.

python/graphstorm/model/gnn.py
```diff
@@ -666,6 +666,7 @@ def freeze_input_encoder(self, train_data):
     def unfreeze_input_encoder(self):
         """ Unfreeze input layer for model training
         """
+        # TODO(zhengda) we need to remove the computed BERT embeddings as node features.
```
classicsong, line 669: Can you provide more information for this TODO?

python/graphstorm/model/lm_embed.py
```diff
@@ -262,15 +262,17 @@ def _get_model_name(self, ntype):
...
             embed_path = os.path.join(os.path.join(self._embed_path, ntype),
                     self._get_model_name(ntype))
             if os.path.exists(embed_path):
                 if get_rank() == 0:
                     logging.info("load LM embedding from %s for node type %s",
                             embed_path, ntype)
+                embed_name = embed_ndata_names[ntype]
                 self._lm_emb_cache[ntype] = load_pytorch_embedding(embed_path,
-                        self._g.get_node_partition_policy(ntype), "bert_emb")
+                        self._g.get_node_partition_policy(ntype), embed_name)
```
classicsong, line 275: It seems it is still using "bert_emb". Any specific reason we what to define the embed_ndata_name?

python/graphstorm/model/lm_embed.py
```diff
@@ -473,6 +488,13 @@ def freeze(self, _):
         self.lm_emb_cache.update_cache(self.lm_infer_batch_size, use_fp16=self.use_fp16)
         self.use_cache = True
 
+    def unfreeze(self):
+        """ Disable Bert caching
+        """
+        if self.num_train != 0:
+            self.use_cache = False
+        self.lm_emb_cache.clear_cache()
```
classicsong, line 496: If num_train is 0, do we need to clear cache?

python/graphstorm/gsf.py
zheng-da: you can search for `feat_size` in the hidden code.

python/graphstorm/model/lm_embed.py
zheng-da: i think it's better to not hard code the embedding name in multiple places.

python/graphstorm/model/embed.py
```diff
@@ -335,12 +352,27 @@ def get_sparse_params(self):
...
+        """
+        return self.feat_size
+
     @property
     def out_dims(self):
         """ The number of output dimensions.
         """
         return self.embed_size
 
+def _gen_emb(g, feat_field, embed_layer, ntype):
```
classicsong, line 367: This function is tricky.

python/graphstorm/model/embed.py
```diff
@@ -378,6 +410,10 @@ def compute_node_input_embeddings(g, batch_size, embed_layer,
     start = time.time()
     with th.no_grad():
         for ntype in target_ntypes:
+            # If we cannot generate embeddings on the node type, let's skip the node type.
+            if not _gen_emb(g, feat_field, embed_layer, ntype):
```
classicsong, line 414: Can you elaborate more why we can skip this node type?

python/graphstorm/model/embed.py
```diff
@@ -378,6 +431,10 @@ def compute_node_input_embeddings(g, batch_size, embed_layer,
     start = time.time()
     with th.no_grad():
         for ntype in target_ntypes:
+            # If we cannot generate embeddings on the node type, let's skip the node type.
```
classicsong, line 434: Can we add more information like:
```
# For node types that we can not generate embeddings, reconstructed_embed is recommended 
# to create the corresponding embeddings.
```

classicsong (approved):

tests/end2end-tests/graphstorm-nc/mgpu_test.sh
```diff
@@ -229,6 +229,21 @@ python3 $GS_HOME/tests/end2end-tests/check_infer.py --train_embout /data/gsgnn_n
...
+
+python3 -m graphstorm.run.gs_node_classification --inference --workspace $GS_HOME/inference_scripts/np_infer/ --num-trainers $NUM_INFERs --num-servers 1 --num-samplers 0 --part-config /data/movielen_100k_movie_lm_encoder_train_val_1p_4t/movie-lens-100k-text.json --ip-config ip_list.txt --ssh-port 2222 --cf ml_nc_movie_text_infer.yaml --use-mini-batch-infer true --save-embed-path /data/gsgnn_nc_ml_text/infer-emb/ --restore-model-path /data/gsgnn_nc_ml_movie_text/epoch-0/ --save-prediction-path /data/gsgnn_nc_ml_text/prediction/ --logging-file /tmp/log.txt --construct-feat-ntype user
+
+error_and_exit $?
+
+# Here we test the GNN embeddings with full-graph inference and mini-batch inference.
+python3 $GS_HOME/tests/end2end-tests/check_infer.py --train_embout /data/gsgnn_nc_ml_text/emb/ --infer_embout /data/gsgnn_nc_ml_text/infer-emb/ --mini-batch-infer
+
+error_and_exit $?
+
```
classicsong, line 246: You may need to rm the /tmp/log.txt

PR #565, opened 2023-10-16
[Wholegraph-feature] Enable inference support using WholeGraph

isratnisa:
Add inference support using WholeGraph


classicsong (approved):
LGTM

PR #569, opened 2023-10-16
fix no in-degree destination nodes error

zhjwy9343:
*Issue #586, if available:*

*Description of changes:*
This PR include a fix of no in-degree destination nodes issue in the hgt_encoder.

The previous implementation of hgt_encoder use source node's feature to compute the output of destination nodes if there is no edges in a block. This implementation assumes if there is no edge, should no destination nodes, which is not true if the destination nodes are the target nodes and no in-degree. This case will cause the input tensor has more rows than the next block's source nodes, which cause an error of different number of rows.

The current implementation will use only the destination nodes to compute the outputs, which solve the different number of rows issue.

This PR also fixed the issue #586 when adding new test model cases.

This PR also add the logic to handle zero destination nodes in a block.

Related: #563


python/graphstorm/model/hgt_encoder.py
```diff
@@ -259,10 +259,12 @@ def forward(self, g, h):
...
                     else:                       # Nodes not really in destination side.
                         warnings.warn("Warning. Graph convolution returned empty "
-                          f"dictionary, for node with type: {str(k)}")
-                        # So add psudo self-loop with feature copy.
-                        trans_out = self.drop(self.a_linears[k](h[k]))
-                        trans_out = trans_out * alpha + h[k] * (1-alpha)
+                          f"dictionary for node with type: {str(k)}. Pleaes check your data \
+                            for no in-degree {str(k)} nodes.")
+                        # So add psudo self-loop for the destination nodes with its own feature.
+                        dst_h = h[k][:g.num_dst_nodes(k)]
```
classicsong, line 265: So the case is there are dst_nodes but there is no in edge for those dst nodes, right?
Can you add a unit test in test_nn_model.py?

zhjwy9343: Add a unit test for this case.

python/graphstorm/model/hgt_encoder.py
```diff
@@ -252,17 +252,20 @@ def forward(self, g, h):
...
                     else:                       # Nodes not really in destination side.
                         warnings.warn("Warning. Graph convolution returned empty "
-                          f"dictionary, for node with type: {str(k)}")
-                        # So add psudo self-loop with feature copy.
-                        trans_out = self.drop(self.a_linears[k](h[k]))
-                        trans_out = trans_out * alpha + h[k] * (1-alpha)
+                          f"dictionary for node with type: {str(k)}. Pleaes check your data \
+                            for no in-degree {str(k)} nodes.")
+                        # So add psudo self-loop for the destination nodes with its own feature.
+                        dst_h = self.a_linears[k](h[k][:g.num_dst_nodes(k)])
```
classicsong, line 266: So the shape mis-match happens here?

tests/unit-tests/test_nn_model.py
```diff
@@ -106,7 +139,26 @@ def test_rgat_with_zero_input(input_dim, output_dim):
...
+    assert outputs['n1'].shape[1] == output_dim
+    assert outputs['n2'].shape[0] == 64
+    assert outputs['n2'].shape[1] == output_dim
 
 if __name__ == '__main__':
     test_rgcn_with_zero_input(32,32)
-    test_rgat_with_zero_input(32,64)
\ No newline at end of file
+    test_rgat_with_zero_input(32,64)
+    test_hgt_with_no_indegree_dstnodes(32, 64)
```
classicsong, line 164: Can you also check whether hgt can pass the test_rgat_with_zero_input test?

classicsong (approved):

python/graphstorm/model/hgt_encoder.py
zhjwy9343: Yes. Original implementation does not take destination nodes as output.

tests/unit-tests/test_nn_model.py
zhjwy9343: Added, but the outputs are different from RGCN and RGAT. Original HGT algorithm does not handle such case. We need to decide if we want to support 0 number of dst node.

zhjwy9343: Confirm with Minjie, the expected outputs of zero edges are fixed, which is same as the rgcn and rgat.

python/graphstorm/model/rgcn_encoder.py
```diff
@@ -228,7 +228,8 @@ def forward(self, g, inputs):
 
         def _apply(ntype, h):
             if self.self_loop:
-                h = h + th.matmul(inputs_dst[ntype], self.loop_weight)
+                h = h + inputs_dst[ntype]
```
classicsong, line 231: This actually change the model implementation.

tests/unit-tests/test_nn_model.py
```diff
@@ -107,6 +167,68 @@ def test_rgat_with_zero_input(input_dim, output_dim):
...
+    assert outputs['n1'].shape[1] == output_dim
+    assert outputs['n2'].shape[0] == 64
+    assert outputs['n2'].shape[1] == output_dim
+
 if __name__ == '__main__':
-    test_rgcn_with_zero_input(32,32)
-    test_rgat_with_zero_input(32,64)
\ No newline at end of file
+    test_rgcn_with_zero_input(32, 64)
+    test_rgat_with_zero_input(32, 64)
```
classicsong, line 229: Is there any tailing spaces?

tests/unit-tests/test_nn_model.py
```diff
@@ -62,12 +63,46 @@ def create_dummy_test_graph(dim):
...
+    ("n0", "r0", "n1"): (th.arange(256, 1024), th.concat([th.arange(64), th.randint(0, 64, (704,))])),
+    ("n0", "r1", "n2"): (th.arange(256), th.concat([th.arange(64), th.randint(0, 64, (192,))])),
+    ("n1", "r2", "n2"): (th.arange(64, 128), th.arange(64))
+    }
+
+    block = dgl.create_block(edges,
+        num_src_nodes=num_src_nodes,
+        num_dst_nodes=num_nodes_dict)
+
+    inputs = {"n0": th.zeros((1024, dim)),
```
classicsong, line 95: So in this case n0 does not have in edges, right?

zhjwy9343: only the first 256 n0 nodes do not have in edges. The other 768 n0 nodes have.

tests/unit-tests/test_nn_model.py
zhjwy9343: I add a tailing line at the end.

python/graphstorm/model/rgcn_encoder.py
zhjwy9343: Revert to the origin. Let's have another PR to fix this.

PR #575, opened 2023-10-17
[GSProcessing] Add support for numerical and multi-numerical transformations.

### LINKED ISSUE #572: [GSProcessing] Numerical transformation
Provide missing data imputation and value normalization for single and multi-column numerical data.

thvasilo:
*Issue #, if available:*
Closes #572 

*Description of changes:*
* Adds a single-column and multi-column numerical transformations.
* We support missing data imputation, and value normalization.
    * For missing data we support `mean`, `median` and `most_frequent` imputation.
    * For normalization we support `mix-max` and `standard` which divides all values
      by their column sum.
* We update the GConstruct config converter to allow pass-through of the min-max transformation.
* We change the GSProcessing feature config key that describes the transformation from `transform` to `transformation`.

### DIFF
--- a/docs/source/gs-processing/developer/input-configuration.rst
+++ b/docs/source/gs-processing/developer/input-configuration.rst
@@ -12,8 +12,8 @@ between other config formats, such as the one used
 by the single-machine GConstruct module.
 
 GSProcessing can take a GConstruct-formatted file
-directly, and we also provide `a script <https://github.com/awslabs/graphstorm/blob/main/graphstorm-processing/scripts/convert_gconstruct_config.py>`
-that can convert a `GConstruct <https://graphstorm.readthedocs.io/en/latest/configuration/configuration-gconstruction.html#configuration-json-explanations>`
+directly, and we also provide `a script <https://github.com/awslabs/graphstorm/blob/main/graphstorm-processing/scripts/convert_gconstruct_config.py>`_
+that can convert a `GConstruct <https://graphstorm.readthedocs.io/en/latest/configuration/configuration-gconstruction.html#configuration-json-explanations>`_
 input configuration file into the ``GSProcessing`` format,
 although this is mostly aimed at developers, users are
 can rely on the automatic conversion.
@@ -30,11 +30,11 @@ The GSProcessing input data configuration has two top-level objects:
 -  ``version`` (String, required): The version of configuration file being used. We include
    the package name to allow self-contained identification of the file format.
 -  ``graph`` (JSON object, required): one configuration object that defines each
-   of the node types and edge types that describe the graph.
+   of the edge and node types that constitute the graph.
 
 We describe the ``graph`` object next.
 
-``graph`` configuration object
+Contents of the ``graph`` configuration object
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 The ``graph`` configuration object can have two top-level objects:
@@ -135,13 +135,12 @@ objects:
    ``source`` key, with a JSON object that contains
    ``{“column: String, and ”type“: String}``.
 -  ``relation``: (JSON object, required): Describes the relation
-   modeled by the edges. A relation can be common among all edges, or it
-   can have sub-types. The top-level objects for the object are:
+   modeled by the edges. The top-level keys for the object are:
 
    -  ``type`` (String, required): The type of the relation described by
       the edges. For example, for a source type ``user``, destination
-      ``movie`` we can have a relation type ``interacted_with`` for an
-      edge type ``user:interacted_with:movie``.
+      ``movie`` we can have a relation type ``rated`` for an
+      edge type ``user:rated:movie``.
 
 -  ``labels`` (List of JSON objects, optional): Describes the label
    for the current edge type. The label object has the following
@@ -171,9 +170,9 @@ objects:
       -  ``train``: The percentage of the data with available labels to
          assign to the train set (0.0, 1.0].
       -  ``val``: The percentage of the data with available labels to
-         assign to the train set [0.0, 1.0).
+         assign to the validation set [0.0, 1.0).
       -  ``test``: The percentage of the data with available labels to
-         assign to the train set [0.0, 1.0).
+         assign to the test set [0.0, 1.0).
 
 -  ``features`` (List of JSON objects, optional)\ **:** Describes
    the set of features for the current edge type. See the :ref:`features-object` section for details.
@@ -248,12 +247,12 @@ following top-level keys:
       -  ``train``: The percentage of the data with available labels to
          assign to the train set (0.0, 1.0].
       -  ``val``: The percentage of the data with available labels to
-         assign to the train set [0.0, 1.0).
+         assign to the validation set [0.0, 1.0).
       -  ``test``: The percentage of the data with available labels to
-         assign to the train set [0.0, 1.0).
+         assign to the test set [0.0, 1.0).
 
 -  ``features`` (List of JSON objects, optional): Describes
-   the set of features for the current edge type. See the next section, :ref:`features-object`
+   the set of features for the current node type. See the next section, :ref:`features-object`
    for details.
 
 --------------
@@ -272,10 +271,10 @@ can contain the following top-level keys:
         "column": "String",
         "name": "String",
         "transformation": {
-        "name": "String",
-        "kwargs": {
-            "arg_name": "<value>"
-        }
+            "name": "String",
+            "kwargs": {
+                "arg_name": "<value>"
+            }
         },
         "data": {
             "format": "String",
@@ -285,7 +284,7 @@ can contain the following top-level keys:
     }
 
 -  ``column`` (String, required): The column that contains the raw
-   feature values in the dataset
+   feature values in the data.
 -  ``transformation`` (JSON object, optional): The type of
    transformation that will be applied to the feature. For details on
    the individual transformations supported see :ref:`supported-transformations`.
@@ -309,7 +308,7 @@ can contain the following top-level keys:
 
         # Example node config with multiple features
         {
-            # This is where the node structure data exist just need an id col
+            # This is where the node structure data exist, just need an id col in these files
             "data": {
                 "format": "parquet",
                 "files": ["path/to/node_ids"]
@@ -356,7 +355,7 @@ Supported transformations
 
 In this section we'll describe the transformations we support.
 The name of the transformation is the value that would appear
-in the ``transform['name']`` element of the feature configuration,
+in the ``['transformation']['name']`` element of the feature configuration,
 with the attached ``kwargs`` for the transformations that support
 arguments.
 
@@ -373,6 +372,35 @@ arguments.
          split the values in the column and create a vector column
          output. Example: for a separator ``'|'`` the CSV value
          ``1|2|3`` would be transformed to a vector, ``[1, 2, 3]``.
+-  ``numerical``
+
+   -  Transforms a numerical column using a missing data imputer and an
+      optional normalizer.
+   -  ``kwargs``:
+
+      -  ``imputer`` (String, optional): A method to fill in missing values in the data.
+         Valid values are:
+         ``none``, ``mean`` (Default), ``median``, and ``most_frequent``. Missing values will be replaced
+         with the respective value computed from the data.
+      - ``normalizer`` (String, optional): Applies a normalization to the data, after
+         imputation. Can take the following values:
+         - ``none``: (Default) Don't normalize the numerical values during encoding.
+         - ``min-max``: Normalize each value by subtracting the minimum value from it,
+        and then dividing it by the difference between the maximum value and the minimum.
+        - ``standard``: Normalize each value by dividing it by the sum of all the values.
+-  ``multi-numerical``
+
+   -  Column-wise transformation for vector-like numerical data using a missing data imputer and an
+      optional normalizer.
+   -  ``kwargs``:
+
+      - ``imputer`` (String, optional): Same as for ``numerical`` transformation, will
+        apply the ``mean`` transformation by default.
+      - ``normalizer`` (String, optional): Same as for ``numerical`` transformation, no
+        normalization is applied by default.
+      - ``separator`` (String, optional): Same as for ``no-op`` transformation, used to separate numerical
+        values in CSV input. If the input data are in Parquet format, each value in the
+        column is assumed to be an array of floats.
 
 --------------
 
@@ -403,6 +431,8 @@ OAG-Paper dataset
             ],
             "nodes" : [
                 {
+                    "type": "paper",
+                    "column": "ID",
                     "data": {
                         "format": "csv",
                         "separator": ",",
@@ -410,8 +440,18 @@ OAG-Paper dataset
                             "node_feat.csv"
                         ]
                     },
-                    "type": "paper",
-                    "column": "ID",
+                    "features": [
+                        {
+                            "column": "n_citation",
+                            "transformation": {
+                                "name": "numerical",
+                                "kwargs": {
+                                    "imputer": "mean",
+                                    "normalizer": "min-max"
+                                }
+                            }
+                        }
+                    ],
                     "labels": [
                         {
                             "column": "field",
--- a/docs/source/gs-processing/developer/input-configuration.rst
+++ b/docs/source/gs-processing/developer/input-configuration.rst
@@ -379,7 +379,7 @@ arguments.
 
       -  ``imputer`` (String, optional): A method to fill in missing values in the data.
          Valid values are:
-         ``none``, ``mean`` (Default), ``median``, and ``most_frequent``. Missing values will be replaced
+         ``none`` (Default), ``mean``, ``median``, and ``most_frequent``. Missing values will be replaced
          with the respective value computed from the data.
       - ``normalizer`` (String, optional): Applies a normalization to the data, after
          imputation. Can take the following values:
--- a/docs/source/gs-processing/developer/input-configuration.rst
+++ b/docs/source/gs-processing/developer/input-configuration.rst
@@ -71,7 +71,7 @@ objects:
      },
      "source": {"column": "String", "type": "String"},
      "relation": {"type": "String"},
-     "destination": {"column": "String", "type": "String"},
+     "dest": {"column": "String", "type": "String"},
      "labels" : [
             {
                 "column": "String",
--- a/docs/source/gs-processing/developer/input-configuration.rst
+++ b/docs/source/gs-processing/developer/input-configuration.rst
@@ -193,9 +193,8 @@ following top-level keys:
             "files": ["String"],
             "separator": "String"
         },
-        "source": {"column": "String", "type": "String"},
-        "relation": {"type": "String"},
-        "destination": {"column": "String", "type": "String"},
+        "column": "String",
+        "type": "String",
         "labels" : [
                 {
                     "column": "String",
@@ -207,15 +206,15 @@ following top-level keys:
                     }
                 }
             ],
-            "features": [{}]
+        "features": [{}]
     }
 
 -  ``data``: (JSON object, required): Has the same definition as for
    the edges object, with one top-level key for the ``format`` that
    takes a String value, and one for the ``files`` that takes an array
    of String values.
--  ``column``: (String, required): The column in the data that
-   corresponds to the column that stores the node ids.
+-  ``column``: (String, required): The name of the column in the data that
+   stores the node ids.
 -  ``type:`` (String, optional): A type name for the nodes described
    in this object. If not provided the ``column`` value is used as the
    node type.
@@ -252,7 +251,7 @@ following top-level keys:
          assign to the test set [0.0, 1.0).
 
 -  ``features`` (List of JSON objects, optional): Describes
-   the set of features for the current node type. See the next section, :ref:`features-object`
+   the set of features for the current node type. See the section :ref:`features-object`
    for details.
 
 --------------
--- a/docs/source/gs-processing/developer/input-configuration.rst
+++ b/docs/source/gs-processing/developer/input-configuration.rst
@@ -82,8 +82,8 @@ objects:
                     "test": "Float"
                 }
             },
-      ]
-       "features": [{}]
+      ],
+      "features": [{}]
    }
 
 -  ``data`` (JSON Object, required): Describes the physical files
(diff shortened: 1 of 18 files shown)


docs/source/gs-processing/developer/input-configuration.rst
```diff
@@ -285,7 +284,7 @@ can contain the following top-level keys:
     }
 
 -  ``column`` (String, required): The column that contains the raw
```
jalencato, line 286:
```suggestion
-  ``column`` (String, required): The column name that contains the raw
```

graphstorm-processing/tests/test_dist_numerical_transformation.py
```diff
@@ -0,0 +1,299 @@
...
+    DistMultiNumericalTransformation,
+)
+
+_ROOT = os.path.abspath(os.path.dirname(__file__))
+
+
+@pytest.fixture(scope="function", name="multi_num_df_with_missing")
+def multi_num_df_with_missing_fixture(spark: SparkSession):
+    """Create a multi-numerical text DataFrame with missing data"""
+    data = [("1|10",), ("5|2",), ("3|",), ("1|3",), ("5|2",)]
```
jalencato, line 35: Do we need to add a test for all empty data?

thvasilo: I think the process will just error out in that case, which would be expected, there's not much we can do when an entire column is missing.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
```diff
@@ -0,0 +1,349 @@
...
+    @staticmethod
+    def get_transformation_name() -> str:
+        return "DistMultiNumericalTransformation"
+
+    @staticmethod
+    def apply_norm_vector(vector_col: str, shared_norm: str, vector_df: DataFrame) -> DataFrame:
+        """
+        Applies normalizer column-wise with a single vector column as input.
+        """
+        other_cols = list(set(vector_df.columns).difference([vector_col]))
```
classicsong, line 193: Why we need this?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
```diff
@@ -0,0 +1,349 @@
...
+        return scaled_df
+
+    @staticmethod
+    def get_transformation_name() -> str:
+        return "DistNumericalTransformation"
+
+
+class DistMultiNumericalTransformation(DistNumericalTransformation):
+    """Transformation to apply missing value imputation and various forms of normalization
+     to a multi-column numerical input, where the input is a string separated by a delimiter.
```
classicsong, line 151: Are we going to have one imputer for each column and one normalizer for each column?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
```diff
@@ -0,0 +1,349 @@
...
+    @staticmethod
+    def apply_norm_vector(vector_col: str, shared_norm: str, vector_df: DataFrame) -> DataFrame:
+        """
+        Applies normalizer column-wise with a single vector column as input.
+        """
+        other_cols = list(set(vector_df.columns).difference([vector_col]))
+
+        if shared_norm == "none":
+            scaled_df = vector_df
+        elif shared_norm == "min-max":
```
classicsong, line 197: Why not using `apply_norm(cols: Sequence[str], shared_norm: str, imputed_df: DataFrame) -> DataFrame`?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
```diff
@@ -0,0 +1,349 @@
...
+            scaled_df = elwise_divider.transform(vector_df).drop(vector_col)
+
+            scaled_df = scaled_df.withColumnRenamed(vector_col + "_scaled", vector_col)
+            scaled_df = scaled_df.select([vector_col] + other_cols)
+        else:
+            raise RuntimeError(f"Unknown normalizer requested for col {vector_col}: {shared_norm}")
+
+        return scaled_df
+
+    def apply(self, input_df: DataFrame) -> DataFrame:
```
classicsong, line 232: Docstring for this func.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
```diff
@@ -0,0 +1,349 @@
...
+            input_has_nan = vector_df_has_nan(vector_df, self.multi_column)
+
+            if input_has_nan:
+                # If the input has NaN, check how values exist in the vectors
+                feat_row = input_df.take(1)[0]
+
+                len_array = len(feat_row[self.multi_column].split(self.original_separator))
+
+                # Doing the column splitting in memory can be expensive
+                if len_array > 50:
```
classicsong, line 292: Why len_array > 50?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
```diff
@@ -0,0 +1,349 @@
...
+                    logging.warning(
+                        "Attempting imputation on %d columns can lead to OOM errors", len_array
+                    )
+
+                # Splitting the vectors requires an array DF
+                if multi_col_type == "string":
+                    assert self.separator, "Separator must be provided for string-separated vectors"
+                    split_array_df = input_df.select(
+                        # After split, replace empty strings with 'NaN' and cast to Array<Float>
+                        F.transform(
```
classicsong, line 302: Can we call convert_multistring_to_vector_df here?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
```diff
@@ -0,0 +1,349 @@
...
+            return vector_df
+
+        def vector_df_has_nan(vector_df: DataFrame, vector_col: str) -> bool:
+            """
+            Returns true if there exists at least one NaN value in `vector_df[vector_col]`
+            """
+            sum_vector_df = vector_df.select(Summarizer.mean(vector_df[vector_col]).alias("sum"))
+            sums_vector = sum_vector_df.take(1)[0]["sum"]
+            for val in sums_vector:
+                if np.isnan(val):
```
classicsong, line 266: Will an empty raw be treated as NaN?

graphstorm-processing/tests/test_dist_heterogenous_loader.py
```diff
@@ -58,12 +58,13 @@ def tempdir_fixture():
 
 @pytest.fixture(scope="function", name="data_configs_with_label")
 def data_configs_with_label_fixture():
-    """Create data configuration object that contain labels"""
-    config_path = os.path.join(_ROOT, "resources/small_heterogeneous_graph/gconstruct-config.json")
+    """Create data configuration object that contain features and labels"""
+    config_path = os.path.join(
+        _ROOT, "resources/small_heterogeneous_graph/gsprocessing-config.json"
```
classicsong, line 63: Can we use /tmp/ ?

graphstorm-processing/tests/test_dist_numerical_transformation.py
```diff
@@ -0,0 +1,299 @@
...
+    """Create a multi-numerical text DataFrame without missing data"""
+    data = [("1|10",), ("5|2",), ("3|2",), ("1|3",), ("5|2",)]
+
+    schema = StructType([StructField("ratings", StringType(), True)])
+    df = spark.createDataFrame(data, schema=schema)
+
+    yield df
+
+
+def test_numerical_transformation_with_mode_imputer(input_df: DataFrame):
```
classicsong, line 54: Where is input_df?

graphstorm-processing/tests/test_dist_numerical_transformation.py
```diff
@@ -0,0 +1,299 @@
...
+    check_df_schema(transformed_df)
+
+    transformed_rows = transformed_df.collect()
+
+    for i, row in zip(range(1, 10, 2), transformed_rows):
+        assert row["ratings"][0] == float(i)
+        assert row["ratings"][1] == float(i + 1)
+
+
+def test_multi_numerical_transformation_with_imputer(
```
classicsong, line 181:
```suggestion
def test_multi_numerical_transformation_with_mean_imputer(
```

graphstorm-processing/graphstorm_processing/config/config_conversion/gconstruct_converter.py
```diff
@@ -80,27 +82,40 @@ def _convert_feature(feats: list[dict]) -> list[dict]:
         list[dict]
             The feature information in the GSProcessing format
         """
-        feats_list = []
+        gsp_feats_list = []
```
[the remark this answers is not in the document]
thvasilo: Yes.

graphstorm-processing/graphstorm_processing/config/numerical_configs.py
```diff
@@ -0,0 +1,92 @@
...
+    def __init__(self, config: Mapping):
+        super().__init__(config)
+        self.imputer = self._transformation_kwargs.get("imputer", "mean")
+        self.norm = self._transformation_kwargs.get("normalizer", "none")
+
+        self._sanity_check()
+
+    def _sanity_check(self) -> None:
+        super()._sanity_check()
+        valid_imputers = ["mean", "median", "most_frequent"]
```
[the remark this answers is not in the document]
thvasilo: Sure. We can either add them the common `constants.py` or at the top of this file, I think having everything in `constants.py` might be a good idea?

docs/source/gs-processing/developer/input-configuration.rst
```diff
@@ -373,6 +372,35 @@ arguments.
...
+         with the respective value computed from the data.
+      - ``normalizer`` (String, optional): Applies a normalization to the data, after
+         imputation. Can take the following values:
+         - ``none``: (Default) Don't normalize the numerical values during encoding.
+         - ``min-max``: Normalize each value by subtracting the minimum value from it,
+        and then dividing it by the difference between the maximum value and the minimum.
+        - ``standard``: Normalize each value by dividing it by the sum of all the values.
+-  ``multi-numerical``
+
+   -  Column-wise transformation for vector-like numerical data using a missing data imputer and an
```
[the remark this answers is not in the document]
thvasilo: We have the option to define it as `none` would you like us to make the default? I'm following NeptuneML here, and it's the "safest" choice for users (otherwise the user will figure out the issue way down the pipeline).

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
thvasilo: This is empirical, it will depend on dataset size really, but it's good to provide a warning to the user.

graphstorm-processing/graphstorm_processing/constants.py
```diff
@@ -19,9 +19,6 @@
 MISSING_CATEGORY = "GSP_CONSTANT_UNKNOWN"
 SINGLE_CATEGORY_COL = "SINGLE_CATEGORY"
 
-################ Multi-numerical Limits #####################
-MAX_COLUMNS_TO_IMPUTE = 50
```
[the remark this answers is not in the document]
thvasilo: We had an alternative implementation for multi-numerical that used a different codepath when we had multiple columns. I removed it here for simplicity, if scaling becomes an issue we can bring it back (it wasn't much help anyway). 

The original cause for this is that to apply imputations to a vector df (where each entry is a numerical vector) we needed to first convert each value in the vector to a column, then apply the imputation to each column. This would often trigger OOM errors when the dimensions are large.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
```diff
@@ -0,0 +1,349 @@
...
+from ..spark_utils import rename_multiple_cols
+
+
+def apply_imputation(cols: Sequence[str], shared_imputation: str, input_df: DataFrame) -> DataFrame:
+    """
+    Applies a single imputation to input dataframe, individually to each of the columns
+    provided in the cols argument.
+    """
+    assert shared_imputation in [
+        "mean",
```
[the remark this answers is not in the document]
thvasilo: Yes, for transformation we'll need to do that, but we haven't come up with a design yet. The overall plan is to have a JSON representation of the transformation from which we can then create a transformer object. For some of the SparkML transformations they have their own serializes so we could also use that.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
```diff
@@ -0,0 +1,349 @@
...
+        "median",
+        "mode",
```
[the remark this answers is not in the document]
thvasilo: It's used in stats to indicate the most frequent value in a dataset: https://en.wikipedia.org/wiki/Mode_(statistics)

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
```diff
@@ -0,0 +1,349 @@
...
+                for scaled_col_name, orig_col in zip(scaled_cols, cols)
+            ]
+            + other_cols
+        )
+    elif shared_norm == "standard":
+        col_sums = imputed_df.agg({col: "sum" for col in cols}).collect()[0].asDict()
+        # TODO: See if it's possible to exclude NaN values from the sum
+        for col, val in col_sums.items():
+            if np.isinf(val) or np.isnan(val):
+                col_sums[col] = 0
```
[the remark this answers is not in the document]
thvasilo: Hmm yeah, I'll just raise an error here and suggest that an imputer is used on the data.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
```diff
@@ -0,0 +1,349 @@
...
+        Valid values are "none", "min-max", and "standard".
+    imputer : str
+        The type of missing value imputation to apply to the column.
+        Valid values are "mean", "median" and "most_frequent".
+    """
+
+    def __init__(self, cols: Sequence[str], normalizer: str, imputer: str) -> None:
+        super().__init__(cols)
+        self.cols = cols
+        self.shared_norm = normalizer
```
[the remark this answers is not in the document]
thvasilo: Not sure if I see the connection, bucket-numerical will have its own independent transformation class. What is the concern here?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
thvasilo: Ah I see this reads as if we do. I'll change the language.

To answer the question in any case, was not planning on supporting separate imputers/normalizers for each col to at least for this iteration. We could implement it if we have a user request.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
thvasilo: We use it to keep track of the other columns in the dataframe to make sure they don't get lost during the call to `select`. On second thought I'm not 100% sure we need this, because each transformer is only supposed to operate on the input columns, the rest are selected away. There might be an edges case I'm forgetting though.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
thvasilo: I'll double-check I think it had to do with retaining all columns.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
thvasilo: The docstring is provided at the parent class. Do we want to have a different `apply` docstring for each transformation?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
thvasilo: I believe so. The `Summarizer` will treat `None`/`NaN` the same I think, any such value in a column will result in a `Nan` being returned.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
thvasilo: Not quite, because that return a `DataFrame[DenseVector]` and in this case we want `DataFrame[ArrayType]`. I could modify that function though to support both return types.

graphstorm-processing/tests/test_dist_heterogenous_loader.py
thvasilo: These input files are bundled with the repo, we use temp directories for the output.

graphstorm-processing/tests/test_dist_numerical_transformation.py
thvasilo: It's defined as an auto-use fixture in `conftest.py` to make it available to all test modules because we use it pretty much everywhere.

graphstorm-processing/graphstorm_processing/config/numerical_configs.py
```diff
@@ -0,0 +1,92 @@
...
+        A separator to use when splitting a delimited string into multiple numerical values
+        as a vector. Only applicable to CSV input. Example: for a separator `'|'` the CSV
+        value `1|2|3` would be transformed to a vector, `[1, 2, 3]`. When `None` the expected
+        input format is an array of numerical values.
+
+    """
+
+    def __init__(self, config: Mapping):
+        super().__init__(config)
+        self.separator = self._transformation_kwargs.get("separator", None)
```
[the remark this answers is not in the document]
thvasilo: It will be provided by the user. It's optional here because it only applies to CSV data.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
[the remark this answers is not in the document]
classicsong: Can we store it back into the config file?
For gconstruct, we do the similar thing. for example, with min-max norm, after processing the graph we store back the actual min value bound and max value bound into a new config.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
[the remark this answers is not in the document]
classicsong: Can you add https://en.wikipedia.org/wiki/Mode_(statistics) as a comment here?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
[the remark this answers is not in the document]
classicsong: I c.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
classicsong: Basically the implementation can be shared.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
classicsong: Can you double check or write some unit test for it?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
classicsong: I think you can unify the implementation.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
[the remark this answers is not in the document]
thvasilo: Will do. Just a note that `mode` is only needed by the Spark transformer internally, we expose `most_frequent` to the user. I'll modify this function to accept both, since it's possible other callers might exist.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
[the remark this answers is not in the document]
thvasilo: Will design more permanent solution for this resolving for this PR.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
thvasilo: Yes verified this is to ensure we have the same behavior as `apply_norm` and `apply_imputation` which will apply the transformations to the columns in `cols` and leave the rest of the columns in the DF unchanged.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
thvasilo: Verified, this, we want to avoid using `apply_norm` here because it tries to use VectorAssembler to concat the existing columns, but our input is already in Vector form here so we're skipping one pass through the data this way.

docs/source/gs-processing/developer/input-configuration.rst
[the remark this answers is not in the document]
classicsong: According to https://github.com/awslabs/graphstorm/pull/575/files#diff-ec035bfb83573a8c31659fd21e4d5211df702c12f6199ee764e3bfab3d2ca846R46 We are now use none as the default value, can you also update the doc accordingly?

classicsong (approved):

PR #578, opened 2023-10-17
Add support for remapping node prediction results

classicsong:
Update graphstrom.gconstruct.remap_result.py to remap node prediction results.
Will add support for remapping embeddings.

https://github.com/awslabs/graphstorm/issues/508


python/graphstorm/inference/np_infer.py
```diff
@@ -146,7 +146,7 @@ def infer(self, loader, save_embed_path, save_prediction_path=None,
         if save_prediction_path is not None:
             # save_embed_path may be None. In that case, we need to init nid_shuffler
             if nid_shuffler is None:
-                nid_shuffler = NodeIDShuffler(g, node_id_mapping_file, list(preds.keys)) \
```
zheng-da, line 149: is this a bug? so it's not triggered in the test?

classicsong: This is a bug. It is not triggerred previously. But as I added more tests, this is triggered.

python/graphstorm/gconstruct/remap_result.py
```diff
@@ -318,32 +463,77 @@ def main(args, gs_config_args):
...
+        # If shared file system is not avaliable
+        # result_info.json is not guaranteed to exist
+        # on each instances. So users must use
+        # --pred-ntypes or --pred-etypes instead.
+        #
+        # In case when both --pred-ntypes or --pred-etypes
+        # are provided while result_info.json is also avaliable,
+        # GraphStorm remaping will follow --pred-ntypes or --pred-etypes
+        # and ignore the result_info.json.
+        if os.path.exists(os.path.join(predict_dir, "result_info.json")):
```
zheng-da, line 504: what does result_info.json provide?

python/graphstorm/gconstruct/remap_result.py
```diff
@@ -318,32 +463,77 @@ def main(args, gs_config_args):
...
+
     for ntype in set(ntypes):
+        mapping_file = os.path.join(id_mapping_path, ntype + "_id_remap.parquet")
         logging.debug("loading mapping file %s",
-                      os.path.join(id_mapping_path, ntype + "_id_remap.parquet"))
-        id_maps[ntype] = \
-            IdReverseMap(os.path.join(id_mapping_path, ntype + "_id_remap.parquet"))
+                      mapping_file)
+        if os.path.exists(mapping_file):
+            id_maps[ntype] = \
```
zheng-da, line 532: where is this used?

zheng-da: i cannot find the code below that uses `id_map`

python/graphstorm/gconstruct/remap_result.py
classicsong: https://github.com/awslabs/graphstorm/blob/id-remap-dev/python/graphstorm/model/utils.py#L1099-L1103
the result_info.json provides: which node type(s) or edge type(s) has the prediction results saved and also how many partitions are there.

python/graphstorm/gconstruct/remap_result.py
classicsong: id_map is a global variable.
it is accessed in worker_remap_node_pred.

classicsong: Here is an short intro of what id_map is
https://github.com/awslabs/graphstorm/pull/578/files#diff-8ed4342d598e284201f07618bf95b5a085b567face2180fa825155738ac53697R43-R50

python/graphstorm/gconstruct/remap_result.py
zheng-da: i feel that we have multiple files that try to save the metadata. do users need to understand these files? i'm a little worried that there will be too many such files and eventually get out of control.

python/graphstorm/gconstruct/remap_result.py
zheng-da: i see. will it be better if we have a class. if we want the child process to inherit the data, we can create a global object of the class. i think this can make the code easier to manage and understand.

python/graphstorm/gconstruct/remap_result.py
classicsong: Most users do not need to understand these configs if they are just using the cmdline to run graphstorm.

zheng-da (approved):

PR #584, opened 2023-10-19
Fix a bug in gconstruct

zheng-da:
*Description of changes:*
The data fields we want to read from the input files are the same. We only need to read a data field once.


python/graphstorm/gconstruct/file_io.py
```diff
@@ -345,6 +345,7 @@ def _parse_file_format(conf, is_node, in_mem):
         for label_conf in conf["labels"]:
             if "label_col" in label_conf:
                 keys.append(label_conf["label_col"])
+    keys = list(set(keys))
```
classicsong, line 348: Why not directly raise an error?

classicsong (approved):
LGTM

PR #587, opened 2023-10-20
Fix load embeding

classicsong:
Now the embeddings saved by save_pytorch_embeddings are saved as per node type.

This PR fixes the path issue of load cached LM embeddings.

### DIFF
--- a/python/graphstorm/model/lm_embed.py
+++ b/python/graphstorm/model/lm_embed.py
@@ -31,8 +31,8 @@ from .embed import GSNodeInputLayer
 from .embed import GSNodeEncoderInputLayer
 from .lm_model import init_lm_model
 from .lm_model import get_lm_node_feats
-from .utils import (load_pytorch_embedding,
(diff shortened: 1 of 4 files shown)


python/graphstorm/model/lm_embed.py
```diff
@@ -258,6 +258,7 @@ def _load_embeddings(self):
         for ntype in self._lm_models.ntypes:
             embed_path = os.path.join(os.path.join(self._embed_path, ntype),
                     self._get_model_name(ntype))
+            embed_path = os.path.join(embed_path, ntype)
```
zheng-da, line 261: why do we add ntype in the path again?

classicsong: Because, the node embedding saving logic has been changed.
It automatically save embeddings from different ntypes into embedding/<ntype>/ folder.

zheng-da (approved):

PR #588, opened 2023-10-20
[Enhancement] Always cache LM when training GLEM GNN to speed up training

wangz10:
*Issue #, if available:*

*Description of changes:*
When using GLEM to train GNN, cache the LM embeddings to speed up its training. 


python/graphstorm/trainer/glem_np_trainer.py
```diff
@@ -158,17 +158,25 @@ def fit(self, train_loader, num_epochs,
...
-                no_pl = False
-
             use_gnn = self._model.em_order_gnn_first
             # `use_gnn`` determines which part to train, if `em_order_gnn_first`
             # 1st round: train GNN, fix LM; 2nd round: train LM fix gnn
             for _ in range(2):
                 stage_start_time = time.time()
                 part_to_train = 'gnn' if use_gnn else 'lm'
                 self._model.toggle(part_to_train)
+                if part_to_train == 'gnn' and not self._model.node_input_encoder.use_cache:
```
zheng-da, line 168: the code here is difficult to understand. I think we need to handle the cache correctly in `freeze_input_encoder`, instead of handling here.

when freezing the input encoder, we need to check the existence of the cache and update the cache if the LM model is updated. However, i think we should clear up the cache in `toggle` when we switch from GNN to LM.

wangz10: Moved to `update_cache`

python/graphstorm/trainer/glem_np_trainer.py
```diff
@@ -158,17 +158,20 @@ def fit(self, train_loader, num_epochs,
...
                 stage_start_time = time.time()
                 part_to_train = 'gnn' if use_gnn else 'lm'
-                self._model.toggle(part_to_train)
+                self._model.toggle(part_to_train, data)
+                if part_to_train == 'lm':
+                    # when training lm, unfreeze LM if past warmup epochs
+                    # otherwise, only train the LM decoder
+                    if freeze_input_layer_epochs <= epoch:
+                        no_pl = False
+                        self._model.lm.unfreeze_input_encoder()
```
zheng-da, line 173: `toggle` doesn't unfreeze the LM model?

wangz10: Unfreezing of LM models is handled in trainer. Because the `freeze_input_layer_epochs` can decide whether to unfreeze LM. If still in warm up epochs, do not unfreeze LM.

zheng-da (approved):

PR #589, opened 2023-10-20
Raise an error when gconstruct reads empty files.

classicsong:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/python/graphstorm/gconstruct/file_io.py
+++ b/python/graphstorm/gconstruct/file_io.py
@@ -20,7 +20,6 @@ from functools import partial
 import glob
 import json
 import os
-import logging
 
 import pyarrow.parquet as pq
 import pyarrow as pa
--- a/python/graphstorm/gconstruct/file_io.py
+++ b/python/graphstorm/gconstruct/file_io.py
@@ -81,6 +81,9 @@ def read_data_csv(data_file, data_fields=None, delimiter=','):
     dict of Numpy arrays.
     """
     data = pd.read_csv(data_file, delimiter=delimiter)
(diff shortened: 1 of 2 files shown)


python/graphstorm/gconstruct/file_io.py
```diff
@@ -132,6 +135,8 @@ def read_data_json(data_file, data_fields):
         for line in json_file.readlines():
             record = json.loads(line)
             data_records.append(record)
+    assert len(data_records) > 0, \
+        f"{data_file} is empty {data_records}."
```
zheng-da, line 139: if it's empty, why do you need to include data_records in the message?

python/graphstorm/gconstruct/file_io.py
```diff
@@ -80,6 +80,9 @@ def read_data_csv(data_file, data_fields=None, delimiter=','):
     dict of Numpy arrays.
     """
     data = pd.read_csv(data_file, delimiter=delimiter)
+    assert data.shape[0] > 0, \
+        f"{data_file} has an empty data. The data frame shape is {data.shape}"
```
zheng-da, line 84: if the file is empty, what is the shape of the data?

zheng-da (approved):

python/graphstorm/gconstruct/file_io.py
classicsong: It still contains the column names.

python/graphstorm/gconstruct/file_io.py
classicsong: It still contains the column names.

PR #590, opened 2023-10-23
Multi Column Feature for Gconstruct

jalencato:
*Issue #, if available:*

*Description of changes:*

Draft for allow multi column feature for gconstruct


python/graphstorm/gconstruct/transform.py
```diff
@@ -1061,14 +1076,19 @@ def preprocess_features(data, ops):
...
-        for key, val in res.items():
-            pre_data[key] = val
+        if isinstance(op.col_name, str):
+            col_name = [op.col_name]
+        else:
+            col_name = op.col_name
+        for col in col_name:
+            res = op.pre_process(data[col])
+            assert isinstance(res, dict)
+            for key, val in res.items():
```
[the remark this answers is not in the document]
classicsong: There is a bug here.

python/graphstorm/gconstruct/transform.py
```diff
@@ -1080,23 +1100,55 @@ def process_features(data, ops):
...
+            res = op(data[col])
+            assert isinstance(res, dict)
+            for key, val in res.items():
+                # Check if it has 1D features. If yes, convert to 2D features
+                if len(val.shape) == 1:
+                    if isinstance(val, ExtMemArrayWrapper):
+                        val = val.to_numpy().reshape(-1, 1)
+                    else:
+                        val = val.reshape(-1, 1)
+                if len(col_name) == 1:
```
classicsong, line 1131: What if res.items() returns multiple key, val pairs? Are we going to only store the first one?

python/graphstorm/gconstruct/transform.py
```diff
@@ -1080,23 +1100,55 @@ def process_features(data, ops):
...
-                if isinstance(val, ExtMemArrayWrapper):
-                    val = val.to_numpy().reshape(-1, 1)
+        hash_hex_feature_path = generate_hash()
+        feature_path = 'feature_{}_{}'.format(op.feat_name, hash_hex_feature_path)
+        feature_path = ext_mem + feature_path if ext_mem is not None \
+            else feature_path
+        if isinstance(op.col_name, str):
+            col_name = [op.col_name]
+        else:
+            col_name = op.col_name
```
classicsong, line 1119: We can move L1112 to L1114 here and change the logic into something like:
```
if ext_mem is not None:
    hash_hex_feature_path = generate_hash()
    feature_path = 'feature_{}_{}'.format(op.feat_name, hash_hex_feature_path)
    feature_path = os.path.join(ext_mem, feature_path)
    os.makedirs(feature_path)
    wrapper = ExtFeatureWrapper(feature_path
else:
    wrapper = None
```

python/graphstorm/gconstruct/transform.py
```diff
@@ -1080,23 +1100,55 @@ def process_features(data, ops):
...
+                        val = val.reshape(-1, 1)
+                if len(col_name) == 1:
+                    new_data[key] = val
+                    continue
+                tmp_key = key
+                # Use external memory if it is available
+                if ext_mem is not None:
+                    if not os.path.exists(feature_path):
+                        os.makedirs(feature_path)
+                        wrapper = ExtFeatureWrapper(feature_path, val.shape, val.dtype)
```
classicsong, line 1139: Can we create ExtFeatureWrapper before L1121?

python/graphstorm/gconstruct/transform.py
```diff
@@ -1080,23 +1100,55 @@ def process_features(data, ops):
...
+
+                    hash_hex = generate_hash()
+                    val_path = feature_path + '/{}_{}.npy'.format(col, hash_hex)
```
classicsong, line 1142: Why we need val_path here?

python/graphstorm/gconstruct/utils.py
```diff
@@ -141,6 +144,14 @@ def _estimate_sizeof(data):
 
     return data_size
 
+def generate_hash():
+    """ Generate unique hashcode
+    """
+    random_uuid = uuid.uuid4()
```
classicsong, line 150: Can we use uuid directly?

python/graphstorm/gconstruct/utils.py
```diff
@@ -530,6 +541,117 @@ def to_tensor(self):
...
+        The shape of the array.
+    dtype : numpy dtype
+        The data type.
+    merged_file: str
+        The merged file name
+    """
+    def __init__(self, arr_path, shape, dtype,
+                 merged_file="merged_feature.npy"):
+        self._arr_path = arr_path
+        self._shape = (shape[0], 0)
```
classicsong, line 561: Do we really need shape here?

python/graphstorm/gconstruct/utils.py
```diff
@@ -530,6 +541,117 @@ def to_tensor(self):
...
+        self._orig_dtype = self._dtype = dtype
+        self._arr = None
+        self.merged_file = merged_file
+        self.wrapper = []
+
+    @property
```
classicsong, line 567: Can we inherit from ExtNumpyWrapper? Many functions are same. And this class can be treated as a sub-class of ExtNumpyWrapper as it also uses Numpy mmap.

python/graphstorm/gconstruct/utils.py
```diff
@@ -530,6 +541,117 @@ def to_tensor(self):
...
+            return arr
+        else:
+            return self._arr.astype(self._dtype)
+
+    def to_tensor(self):
+        """ Return Pytorch tensor.
+        """
+        return th.tensor(self.to_numpy())
+
+    def append(self, wrap, path=None):
```
classicsong, line 625: The path is not necessary, we can generate a temporal path.
BTW, please add doc strings.

python/graphstorm/gconstruct/utils.py
```diff
@@ -530,6 +541,117 @@ def to_tensor(self):
...
+            ext_val[:] = wrap[:]
+            ext_val.flush()
+            wrap = ExtNumpyWrapper(path, wrap.shape, wrap.dtype)
+        self.wrapper.append(wrap)
+
+    def merge(self):
+        """ Return feature col-wised.
+        """
+        for wrap in self.wrapper:
+            self._shape = (wrap.shape[0], self._shape[1] + wrap.shape[1])
```
classicsong, line 643: Can you do something like:
```
feat_dim = sum([wrap.shape[1] for wrap in self.wrapper])
self._shape = (wrap.shape[0], feat_dim)
```

python/graphstorm/gconstruct/transform.py
jalencato: I think we can put the val_path inside the function.

python/graphstorm/gconstruct/utils.py
jalencato: Let me check if we can directly use str(random_uuid), uuid here is an object, so we need to use the string conversion

python/graphstorm/gconstruct/utils.py
jalencato: The only concern I did not inherit is about the shape definition here. Let me check if we can do so. Mainly about the shape.

python/graphstorm/gconstruct/utils.py
jalencato: Let me move the outside path inside so we will not need temporal path here.

python/graphstorm/gconstruct/transform.py
classicsong: You did not resolve my question.

classicsong: If we only expect res to return 1 (key, val). Please add an assert here and also a TODO.

python/graphstorm/gconstruct/utils.py
```diff
@@ -530,6 +538,91 @@ def to_tensor(self):
         """
         return th.tensor(self.to_numpy())
 
+class ExtFeatureWrapper(ExtNumpyWrapper):
+    """ The wrapper to memory-mapped Numpy array when combining feature
+
+    Parameters
+    ----------
+    arr_path : str
+        The path to different feature files directory
```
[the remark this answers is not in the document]
classicsong: and dtype

python/graphstorm/gconstruct/utils.py
```diff
@@ -530,6 +538,91 @@ def to_tensor(self):
...
+
+    def __del__(self):
+        self.cleanup()
+
+    def cleanup(self):
+        """ Clean up the array.
+        """
+        self._arr.flush()
+        self._arr = None
+        shutil.rmtree(self.directory_path)
```
classicsong, line 576: Why remove the entire directory, not the self._arr_path?

python/graphstorm/gconstruct/utils.py
```diff
@@ -530,6 +538,91 @@ def to_tensor(self):
...
+            return self._arr.astype(self._dtype)
+
+    def append(self, wrap):
+        """Add external memory wrapper to the wrapper
+        """
+        if not self.shape:
+            self._shape = (wrap.shape[0], 0)
+        if self.shape and self.shape[0] != wrap.shape[0]:
+            raise RuntimeError("Expect that ExtFeatureWrapper has a "
+                               "first dimension that is the same")
```
classicsong, line 598: Expect that the first dimension of the ExtFeatureWrapper and wrap is same but get xxx and xxx.

python/graphstorm/gconstruct/utils.py
```diff
@@ -530,6 +538,91 @@ def to_tensor(self):
...
+            raise RuntimeError("The ExtFeatureWrapper needs to be merge first")
+        if self._arr is None:
+            arr = np.memmap(self._arr_path, self._orig_dtype, mode="r", shape=self._shape)
+            if self._dtype != self._orig_dtype:
+                arr = arr.astype(self._dtype)
+            return arr
+        else:
+            return self._arr.astype(self._dtype)
+
+    def append(self, wrap):
```
classicsong, line 591: Add docstring for parameters.

python/graphstorm/gconstruct/utils.py
```diff
@@ -530,6 +538,91 @@ def to_tensor(self):
...
+        """Add external memory wrapper to the wrapper
```
classicsong, line 592:
```suggestion
        """Add an external memory wrapper.
```

classicsong: Actually, the input may not be an wrapper, it can also be an array.
You can just call it feature.

python/graphstorm/gconstruct/utils.py
```diff
@@ -530,6 +538,91 @@ def to_tensor(self):
...
+            ext_val = np.memmap(path, wrap.dtype, mode="w+", shape=wrap.shape)
+            ext_val[:] = wrap[:]
+            ext_val.flush()
+            wrap = ExtNumpyWrapper(path, wrap.shape, wrap.dtype)
+        self.wrapper.append(wrap)
+
+    def merge(self):
+        """ Merge feature col-wised.
+        """
+        assert self.wrapper, "Cannot merge None list, " \
```
classicsong, line 611:
```suggestion
        assert self.wrapper, "Cannot merge an empty list, " \
```

tests/unit-tests/gconstruct/test_construct_graph.py
```diff
@@ -1274,6 +1277,407 @@ def test_parse_edge_data():
...
+        assert str(e) == "Do not support categorical feature " \
+                         "transformation on multiple columns"
+
+    feat_op8 = [{
+        "feature_col": [],
+        "feature_name": "test3"
+    }]
+    try:
+        (res, _, _) = parse_feat_ops(feat_op8)
+        assert False, "expected Error raised for invalid feature column"
```
classicsong, line 1560: Please be more specific of what corner case you are testing. So other maintainer can understand what error you are testing here.

python/graphstorm/gconstruct/transform.py
```diff
@@ -1061,14 +1082,29 @@ def preprocess_features(data, ops):
...
-        assert isinstance(res, dict)
-        for key, val in res.items():
-            pre_data[key] = val
+        if isinstance(op.col_name, str):
+            col_name = [op.col_name]
+        else:
+            col_name = op.col_name
+        for col in col_name:
+            res = op.pre_process(data[col])
+            # Do not expect multiple keys for multiple columns, expect output is
```
classicsong, line 1090:
```suggestion
            # Do not expect multiple keys for multiple columns, the expected output will only have 1 key/val pair.
```

python/graphstorm/gconstruct/transform.py
```diff
@@ -1061,14 +1082,29 @@ def preprocess_features(data, ops):
...
+            # {feat_name: encoding}. But for single column, some feature transformation like
```
classicsong, line 1091:
```suggestion
            # But for single column, some feature transformations like
```

python/graphstorm/gconstruct/transform.py
```diff
@@ -1080,23 +1116,60 @@ def process_features(data, ops):
...
+                if len(val.shape) == 1:
+                    if isinstance(val, ExtMemArrayWrapper):
+                        val = val.to_numpy().reshape(-1, 1)
+                    else:
+                        val = val.reshape(-1, 1)
+                if len(col_name) == 1:
+                    new_data[key] = val
+                    continue
+                tmp_key = key
+                # Use external memory if it is available
```
classicsong, line 1161:
```suggestion
                # Use external memory if it is required
```

python/graphstorm/gconstruct/transform.py
```diff
@@ -1080,23 +1116,60 @@ def process_features(data, ops):
...
+            hash_hex_feature_path = generate_hash()
+            feature_path = 'feature_{}_{}'.format(op.feat_name, hash_hex_feature_path)
+            feature_path = ext_mem_path + feature_path
+            os.makedirs(feature_path)
+            wrapper = ExtFeatureWrapper(feature_path)
+        else:
+            wrapper = None
+        for col in col_name:
+            res = op(data[col])
+            # Do not expect multiple keys for multiple columns, expect output is
```
classicsong, line 1143: Same comment as in preprocess_features.

python/graphstorm/gconstruct/transform.py
```diff
@@ -1080,23 +1116,60 @@ def process_features(data, ops):
...
+            # {feat_name: encoding}. But for single column, some feature transformation like
+            # Tokenizer will return multiple key-val pairs, so do not check for single column
+            if len(col_name) > 1:
+                assert isinstance(res, dict) and len(res) == 1, \
+                    f"It is expected only have one feature name after the process_features " \
+                    f"for multiple columns, but get {len(res)}"
```
classicsong, line 1149:
```suggestion
                    f"for multiple column feature transformation, but get {len(res)}"
```

python/graphstorm/gconstruct/utils.py
```diff
@@ -530,6 +538,107 @@ def to_tensor(self):
         """
         return th.tensor(self.to_numpy())
 
+class ExtFeatureWrapper(ExtNumpyWrapper):
+    """ The wrapper to memory-mapped Numpy array when combining features
+
+    Parameters
+    ----------
+    arr_path : str
+        The path to different feature files directory
```
classicsong, line 547:
```suggestion
        A path to the directory of different feature files.
```

python/graphstorm/gconstruct/utils.py
```diff
@@ -530,6 +538,107 @@ def to_tensor(self):
...
+            The value needs to be packed.
+        """
+        if not self._shape:
+            self._shape = (feature.shape[0], 0)
+        if self._shape and self._shape[0] != feature.shape[0]:
+            raise RuntimeError(f"Expect that ExtFeatureWrapper has a "
+                               f"first dimension that is the same but get "
+                               f"{self.shape[0]} and {feature.shape[0]}")
+        # Convert the numpy array into an ExtNumpyWrapper
+        # By converting into an ExtNumpyWrapper, it is possible to avoid loading
```
classicsong, line 611:
```suggestion
        # By converting it into an ExtNumpyWrapper, it will avoid loading
```

python/graphstorm/gconstruct/utils.py
```diff
@@ -530,6 +538,107 @@ def to_tensor(self):
...
+    def __getitem__(self, idx):
+        if not self._shape:
+            raise RuntimeError("Call ExtFeatureWrapper.merge() first before calling __getitem__")
+        if self._arr is None:
+            self._arr = np.memmap(self._arr_path, self._dtype, mode="r",
+                                  shape=self._shape)
+        return self._arr[idx].astype(self.dtype)
+
+    def __del__(self):
+        self.cleanup()
```
classicsong, line 571: Can you add a unitest to check if __del__ works in the unit test?

python/graphstorm/gconstruct/transform.py
```diff
@@ -989,6 +995,13 @@ def parse_feat_ops(confs):
                                       infer_batch_size=infer_batch_size,
                                       out_dtype=out_dtype)
             elif conf['name'] == 'max_min_norm':
+                # TODO: Not support max_min_norm feature transformation on multiple columns
+                # without explicitly defining max_val and min_val. Otherwise, it is not clear to
```
classicsong, line 999:
```suggestion
                # without explicitly defining max_val and min_val. Otherwise, the definition of max_val and min_val for each column is unclear.
```

python/graphstorm/gconstruct/utils.py
jalencato: It is already there, checkout https://github.com/awslabs/graphstorm/blob/7d140e150d1fc3c3ef12fbaa9b55e17d0915b429/tests/unit-tests/gconstruct/test_construct_graph.py#L1686.

classicsong (approved):
LGTM

PR #593, opened 2023-10-24
Speed up sampling for constructing node features.

zheng-da:
*Description of changes:*
When sampling more neighbors to construct node features, we should run this in the sampling process instead of in the trainer process.


python/graphstorm/dataloading/dataloading.py
```diff
@@ -168,6 +168,23 @@ def fanout(self):
         """
         return self._fanout
 
+class MultiLayerNeighborSamplerForReconstruct(dgl.dataloading.BlockSampler):
+    """
+    """
```
classicsong, line 173: Docstring for MultiLayerNeighborSamplerForReconstruct

classicsong (approved):
LGTM

PR #596, opened 2023-10-26
Add support in-batch joint negative

classicsong:
In addition to sharing negatives in the joint negative sampler, we add more negatives by using in-batch negatives.

### DIFF
--- a/tests/unit-tests/test_dataloading.py
+++ b/tests/unit-tests/test_dataloading.py
@@ -42,6 +42,7 @@ from graphstorm.dataloading import (GSgnnNodeDataLoader,
                                     GSgnnNodeSemiSupDataLoader)
 from graphstorm.dataloading import (GSgnnLinkPredictionDataLoader,
                                     GSgnnLPJointNegDataLoader,
+                                    GSgnnLPInBatchJointNegDataLoader,
                                     GSgnnLPLocalUniformNegDataLoader,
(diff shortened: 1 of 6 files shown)


tests/unit-tests/test_dataloading.py
```diff
@@ -1306,7 +1313,32 @@ def test_lp_dataloader_len(batch_size):
...
+    src = th.arange(num_pos)
+    dst = th.arange(num_pos)
+    g = dgl.heterograph({
+        ("n0", "r0", "n1"): (src, dst),
+    })
+    sampler = InbatchJointUniform(num_neg)
+    src, dst = sampler._generate(g, th.arange(num_pos), ("n0", "r0", "n1"))
+    assert len(src) == num_pos * (num_pos - 1) + num_pos * num_neg
+    assert len(dst) == num_pos * (num_pos - 1) + num_pos * num_neg
+    in_batch_src = src[-num_pos * (num_pos - 1):]
```
zhjwy9343, line 1328: Why do we need to use this method to get in-batch source and destinations? 

I thought the outputs of  sampler._generate() is the in-batch source and destination.

zhjwy9343: I see. It is concat in the _generate() method.

zhjwy9343 (approved):
LGTM

PR #615, opened 2023-11-02
[GSProcessing] Add support for Rank-Gauss Feature Transformation

jalencato:
*Issue #, if available:*

*Description of changes:*


docs/source/gs-processing/developer/input-configuration.rst
```diff
@@ -385,8 +385,13 @@ arguments.
          imputation. Can take the following values:
          - ``none``: (Default) Don't normalize the numerical values during encoding.
          - ``min-max``: Normalize each value by subtracting the minimum value from it,
-        and then dividing it by the difference between the maximum value and the minimum.
-        - ``standard``: Normalize each value by dividing it by the sum of all the values.
+         and then dividing it by the difference between the maximum value and the minimum.
+         - ``standard``: Normalize each value by dividing it by the sum of all the values.
+         - ``rank-gauss``: Normalize each value by rank-gauss normalization. It maps the numeric feature values
+         to gaussian distribution based on ranking. Rank-gauss normalization requires lots of data reshuffling,
+         so it will be a great cost for memory. It is only for single column.
```
thvasilo, line 392:
```suggestion
         - ``rank-gauss``: Normalize each value using Rank-Gauss normalization. Rank-gauss first ranks all values,
         converts the ranks to the -1/1 range, and applies the `inverse of the error function <https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.erfinv.html>`_ to make the values conform
         to a Gaussian distribution shape. This transformation only supports a single column as input.
```

graphstorm-processing/tests/test_dist_numerical_transformation.py
```diff
@@ -17,10 +17,11 @@
 
 import pytest
 import pandas as pd
-from numpy.testing import assert_array_equal, assert_array_almost_equal
+import numpy as np
+from numpy.testing import assert_array_equal, assert_array_almost_equal, assert_almost_equal
 from pyspark.sql import SparkSession, DataFrame, functions as F
 from pyspark.sql.types import ArrayType, FloatType, StructField, StructType, StringType
-
+from scipy.special import erfinv
```
thvasilo, line 24:
```suggestion
from scipy.special import erfinv

```

Add a blank line between third-party and local/library imports: https://peps.python.org/pep-0008/#imports

graphstorm-processing/tests/test_dist_numerical_transformation.py
```diff
@@ -297,3 +298,67 @@ def test_multi_numerical_transformation_with_array_input(spark: SparkSession, ch
...
+    feat = j / divider
+    feat = feat - upper
+    return erfinv(feat)
+
+
+@pytest.mark.parametrize("epsilon", [0.0, 1e-6])
+def test_rank_gauss(spark: SparkSession, check_df_schema, epsilon):
+    data = [("mark", 0.0, None), ("john", 15.0, 10000), ("tara", 26.0, 20000), ("jen", 40.0, 10000)]
+
+    columns = ["name", "age", "salary"]
```
thvasilo, line 320: We don't need the extra columns here right? Can just have data have the float values, and have a single "value" column.

thvasilo (changes requested):
Looks good overall, just need some small changes.

Did we perform a small integration test to ensure the config value get propagated as expected all the way to the transformation from the a GConstruct/GSProcessing config? The unit tests do not cover that AFAIK.

jalencato:
> Looks good overall, just need some small changes.
> 
> Did we perform a small integration test to ensure the config value get propagated as expected all the way to the transformation from the a GConstruct/GSProcessing config? The unit tests do not cover that AFAIK.

Actually that is what I do first. I always first to make a GConstruct/GSProcessing config convert/run correctly, and then start to write unit-test and e2e test. I did not add that here is that we do not have a GSProcessing e2e test. Maybe it will be necessary in the future. For feature transformation, I think the single unit tests will cover most of the cases. But at least for my PRs, I always do some local e2e test first.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
```diff
@@ -146,6 +156,35 @@ def single_vec_to_float(vec):
         scaled_df = imputed_df.select(
             [(F.col(c) / col_sums[f"sum({c})"]).alias(c) for c in cols] + other_cols
         )
+    elif shared_norm == "rank-gauss":
+        assert len(cols) == 1, "Rank-Guass numerical transformation only supports single column"
+        column_name = cols[0]
+        select_df = imputed_df.select(column_name)
+        # original id is the original order for the input data frame, index id is the sorted order.
```
thvasilo, line 163:
```suggestion
        # original id is the original order for the input data frame, value rank indicates the rank of each value in the column
```

thvasilo (approved):
LGTM, just one small change before merging.

PR #616, opened 2023-11-02
Add self-loop in message passing for GAT

classicsong:
Currently GAT model does not include self-loop. This PR will automatically add self-loop in the message passing graph.

*Description of changes:*

### DIFF
--- a/tests/unit-tests/test_gnn.py
+++ b/tests/unit-tests/test_gnn.py
@@ -1527,6 +1527,8 @@ def test_node_mini_batch_gnn_predict():
     th.distributed.destroy_process_group()
 
 if __name__ == '__main__':
+    test_rgcn_node_prediction(None)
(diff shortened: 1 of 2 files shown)


tests/unit-tests/test_gnn.py
```diff
@@ -580,6 +581,7 @@ def test_gat_node_prediction():
                                      train_ntypes=['_N'], label_field='label',
```
zhjwy9343, line 581: Should we change the "_N" here to "DEFAULT_NTYPE"?

zhjwy9343 (approved):
LGTM

tests/unit-tests/test_gnn.py
classicsong: Let me create another PR to update all the places.

PR #619, opened 2023-11-03
Add contrastive loss for link prediction tasks.

classicsong:
Add a contrastive loss for link prediction.


GentleZhu (approved):
LGTM

docs/source/configuration/configuration-run.rst
```diff
@@ -477,10 +477,10 @@ Link Prediction Task
     - Yaml: ``gamma: 10.0``
     - Argument: ``--gamma 10.0``
     - Default value: ``12.0``
-- **lp_loss_func**: Link prediction loss function. Builtin loss functions include ``cross_entropy`` and ``logsigmoid``.
```
zheng-da, line 480: logsigmoid is gone? is this API change?

python/graphstorm/config/argument.py
```diff
@@ -1771,6 +1773,23 @@ def lp_decoder_type(self):
         # Set default value to distmult
         return BUILTIN_LP_DISTMULT_DECODER
 
+    @property
+    def contrastive_loss_temp(self):
```
zheng-da, line 1777: how about using the full name? it's hard for me to think temp is temperature.

python/graphstorm/gsf.py
```diff
@@ -458,44 +463,50 @@ def create_builtin_lp_model(g, config, train_task):
...
+                    if model.gnn_encoder is not None \
+                    else model.node_input_encoder.out_dims
     if config.lp_decoder_type == BUILTIN_LP_DOT_DECODER:
         # if the training set only contains one edge type or it is specified in the arguments,
         # we use dot product as the score function.
         if get_rank() == 0:
             logging.debug('use dot product for single-etype task.')
             logging.debug("Using inner product objective for supervision")
         if config.lp_edge_weight_for_loss is None:
-            decoder = LinkPredictDotDecoder(model.gnn_encoder.out_dims \
```
zheng-da, line 468: are we still going to use `LinkPredictDotDecoder` anywhere?

python/graphstorm/model/edge_decoder.py
```diff
@@ -580,9 +582,8 @@ def forward(self, g, h, e_h=None):
                 src_emb = h[src_type][u]
                 dest_emb = h[dest_type][v]
                 scores_etype = calc_dot_pos_score(src_emb, dest_emb)
-                scores.append(scores_etype)
+                scores[canonical_etype] = scores_etype
 
-            scores=th.cat(scores)
```
zheng-da, line 585: here we actually changed API. is this API exposed to users?

python/graphstorm/model/edge_decoder.py
```diff
@@ -713,6 +714,45 @@ def out_dims(self):
...
+                #
+                # For example:
+                #
+                # pos pairs   |  neg pairs
+                # (10, 20)    |  (10, 3), (10, 1), (10, 0), (10, 22)
+                # (13, 6)     |  (13, 3), (13, 1), (13, 0), (13, 22)
+                # (29, 8)     |  (29, 3), (29, 1), (29, 0), (29, 22)
+                # TODO: use stable to keep the order of negatives. This may not
+                # be necessary
+                u_sort_idx = th.argsort(u, stable=True)
```
zheng-da, line 746: this is really a temporary solution. the comment above actually doesn't make too much sense to be here because there are no positive and negative pairs in this function. the comment should be in the GNN module where we can see positive edges and negative edges.

python/graphstorm/model/edge_decoder.py
```diff
@@ -713,6 +714,45 @@ def out_dims(self):
         """
         return 1
 
+class LinkPredictContrastiveDotDecoder(LinkPredictDotDecoder):
```
zheng-da, line 717: once we get to a better solution, this class is no longer needed?

zheng-da: if so, we shouldn't expose this API to users. in the docstring, we should say that this is a temporary solution.

python/graphstorm/model/loss_func.py
```diff
@@ -110,9 +110,25 @@ class LinkPredictLossFunc(GSLayer):
```
zheng-da, line 110: this class name and the docstring above doesn't make much sense. is this cross-entry loss? we refer to this as link prediction loss?

docs/source/configuration/configuration-run.rst
classicsong: The is no code of logsigmoid. I think we mistakenly add this in our doc.

python/graphstorm/model/edge_decoder.py
classicsong: It is not exposed.

python/graphstorm/model/loss_func.py
```diff
@@ -172,3 +207,68 @@ def out_dims(self):
...
+            nscore.append(n_s.reshape(p_s.shape[0], -1))
+        pscore = th.cat(pscore, dim=0)
+        nscore = th.cat(nscore, dim=0)
+
+        pscore = th.div(pscore, self._temp)
+        nscore = th.div(nscore, self._temp)
+        score = th.cat([pscore.unsqueeze(1), nscore], dim=1)
+
+        exp_logits = th.exp(score)
+        log_prob = pscore - th.log(exp_logits.sum(1))
```
zheng-da, line 252: can you write the formula of the contrastive loss? it's not very clear to me.

python/graphstorm/model/edge_decoder.py
classicsong: > once we get to a better solution, this class is no longer needed?

I think so. But it may take time for us to implement a better solution.
We can hide it from our users.

classicsong: Add a TODO.

python/graphstorm/model/edge_decoder.py
classicsong: You are right.
It works only because we know how neg_graph is built.

python/graphstorm/model/loss_func.py
classicsong: A better name should be LinkPredictBinaryClassLossFunc

python/graphstorm/model/loss_func.py
classicsong: Added in the class description

python/graphstorm/model/edge_decoder.py
zheng-da: can you add a TODO here?

zheng-da (approved):

python/graphstorm/model/edge_decoder.py
classicsong: I added a TODO in L724: 
```
TODO(xiang): develop a better solution to unify the implementation of link prediction decoders.
```

classicsong: I added a TODO in L725 (class description).

PR #623, opened 2023-11-07
[GSProcessing] Categorical Feature Transformation

jalencato:
*Issue #, if available:*

*Description of changes:*


docs/source/gs-processing/developer/input-configuration.rst
```diff
@@ -425,6 +425,17 @@ arguments.
         by specifying a slide-window size ``s``, where ``s`` can an integer or float. GSProcessing then transforms each
         numeric value ``v`` of the property into a range from ``v - s/2`` through ``v + s/2`` , and assigns the value v
         to every bucket that the range covers.
+
+-  ``categorical``
+
+   -  Transforms one value from a fixed list of possible values which can be treated as a categorical feature which are encoded using one-hot encoding. No kwargs required for categorical feature.
```
thvasilo, line 431:
```suggestion
   -  Transforms values from a fixed list of possible values (categorical features) to a one-hot encoding.
   The length of the resulting vector will be the number of categories in the data minus one, with a 1 in
   the index of the single category, and zero everywhere else.
```

docs/source/gs-processing/developer/input-configuration.rst
```diff
@@ -425,6 +425,17 @@ arguments.
...
+
+-  ``multi-categorical``
+
+   -  Apply column-wise transformation for vector-like data from a fixed list of possible values which can be treated as a categorical feature which are encoded using multi-hot encoding.
```
thvasilo, line 435:
```suggestion
   -  Encodes vector-like data from a fixed list of possible values (i.e. multi-label/multi-categorical data) using a multi-hot encoding. The length of the resulting vector will be the number of categories in the data minus one, and each value will have a 1 value for every category that appears, and 0 everwhere else.
```

docs/source/gs-processing/developer/input-configuration.rst
```diff
@@ -425,6 +425,17 @@ arguments.
...
+   -  ``kwargs``:
+      - ``separator`` (String, optional): Same as the one in the No-op operation, the separator is used to
+        split input value in CSV format. If it is not provided, then the whole value will be considered as an array.
```
thvasilo, line 438:
```suggestion
        split multiple input values for CSV files e.g. ``detective|noir``. If it is not provided, then the whole value will be considered as an array.
```

graphstorm-processing/graphstorm_processing/graph_loaders/schema_utils.py
```diff
@@ -92,9 +92,11 @@ def determine_spark_feature_type(feature_type: str) -> Type[DataType]:
         In case an unsupported feature_type is provided.
     """
     # TODO: Replace with pattern matching after moving to Python 3.10?
-    if feature_type in ["no-op", "multi-numerical"] or feature_type.startswith("text"):
+    if feature_type in ["no-op", "multi-numerical", "multi-categorical"] or feature_type.startswith(
+        "text"
+    ):
         return StringType
-    if feature_type in ["numerical", "bucket-numerical", "none"]:
+    if feature_type in ["numerical", "bucket-numerical", "none", "categorical"]:
```
thvasilo, line 99: Categorical features should be parsed as `StringType` (text), move to the case above.

graphstorm-processing/tests/test_converter.py
```diff
@@ -227,6 +227,14 @@ def test_convert_gsprocessing(converter: GConstructConfigConverter):
...
                     "transform": {"name": "rank_gauss", "epsilon": 0.1},
                 },
+                {
+                    "feature_col": ["num_citations"],
+                    "transform": {"name": "to_categorical", "mapping": {"1", "2", "3"}},
+                },
+                {
+                    "feature_col": ["num_citations"],
+                    "transform": {"name": "to_categorical", "separator": ","},
+                },
```
thvasilo, line 237: We are adding these values to the config, but I don't see the logic of the tests changing. Shouldn't we be testing that the config is indeed parsed as expected?

docs/source/gs-processing/developer/input-configuration.rst
```diff
@@ -425,6 +425,17 @@ arguments.
...
 --------------
```
thvasilo, line 439: Let's create a note about hard-coded limit of 100 categories, as in https://docs.aws.amazon.com/neptune/latest/userguide/machine-learning-feature-encoding.html#machine-learning-categorical-features

Use the `.. note::` rst env: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/notes_warnings.html

graphstorm-processing/graphstorm_processing/config/categorical_configs.py
```diff
@@ -0,0 +1,37 @@
...
+class MultiCategoricalFeatureConfig(FeatureConfig):
+    """Feature configuration for multi-column categorical features.
+
+    Supported kwargs
+    ----------------
+    separator: str, optional
+        A separator to use when splitting a delimited string into multiple numerical values
+        as a vector. Only applicable to CSV input. Example: for a separator `'|'` the CSV
+        value `1|2|3` would be transformed to a vector, `[1, 2, 3]`. When `None` the expected
+        input format is an array of numerical values.
```
thvasilo, line 29:
```suggestion
        input format is an array of string values.
```

thvasilo (changes requested):
Let's fix a couple of issues.

Importantly: Did we test cat and multi-cat with both CSV and Parquet input? We should find a way to do so in our unit tests, possibly by generating Parquet files.

graphstorm-processing/tests/test_dist_category_transformation.py
```diff
@@ -174,3 +176,79 @@ def test_multi_category_limited_categories(multi_cat_df_and_separator):
...
+    expected_rows = [
+        [1.0, 1.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0],
+        [0.0, 0.0, 1.0, 1.0, 0.0, 0.0, 0.0, 0.0],
+        [0.0, 0.0, 0.0, 0.0, 1.0, 1.0, 0.0, 0.0],
+        [0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 1.0, 1.0],
+    ]
+    for row, expected_row in zip(transformed_rows, expected_rows):
+        assert row["names"] == expected_row
+
+    shutil.rmtree(parquet_path)
```
thvasilo, line 254: Let's try to put the file creation entire context under a `with tempfile.TemporaryDirectory() as tmpdirname:
`, and write/read from the tempdir, so that even if the test fails, the file cleanup still happens. As it stands, if the assertion fails, `shutil.rmtree` will not be called, leaving the temp file lingering.

See https://docs.python.org/3.9/library/tempfile.html#examples

thvasilo (approved):
LGTM, let's make these two small changes and merge.

PR #627, opened 2023-11-08
[SageMaker] Move SageMaker partition job to use SageMaker Processing instead of Training

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Convert the SageMaker partition jobs from training to processing.
* Add Docker login to SM Docker build script
* Add and pin dependencies to SM Docker container

### DIFF
--- a/docker/sagemaker/Dockerfile.sm
+++ b/docker/sagemaker/Dockerfile.sm
@@ -18,7 +18,15 @@ FROM branch-${DEVICE} AS final
 LABEL maintainer="Amazon AI Graph ML team"
 
 # Install related Python packages
-RUN pip3 install ogb==1.3.6 scipy pyarrow boto3 scikit-learn transformers==4.28.1 \
+RUN pip3 install \
+        boto3 \
+        numba==0.58.1 \
+        numpy==1.26.1 \
+        ogb==1.3.6 \
+        pyarrow \
+        scikit-learn \
+        scipy \
+        transformers==4.28.1 \
     && rm -rf /root/.cache
 
 # Install MPI etc needed by DistDGL
--- a/python/graphstorm/sagemaker/sagemaker_partition.py
+++ b/python/graphstorm/sagemaker/sagemaker_partition.py
@@ -214,7 +214,8 @@ def run_partition(job_config: PartitionJobConfig):
     metadata_filename = job_config.metadata_filename
     skip_partitioning = job_config.skip_partitioning == 'true'
 
-    sm_env = json.loads(os.environ['SM_TRAINING_ENV'])
+    with open("/opt/ml/config/resourceconfig.json", "r", encoding="utf-8") as f:
+        sm_env = json.load(f)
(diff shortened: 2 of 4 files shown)


docker/sagemaker/Dockerfile.sm
```diff
@@ -18,7 +18,15 @@ FROM branch-${DEVICE} AS final
 LABEL maintainer="Amazon AI Graph ML team"
 
 # Install related Python packages
-RUN pip3 install ogb==1.3.6 scipy pyarrow boto3 scikit-learn transformers==4.28.1 \
+RUN pip3 install \
+        boto3 \
+        numba==0.58.1 \
+        numpy==1.26.1 \
```
classicsong, line 24: Does SageMaker partition rely on specify numpy and numba version?

python/graphstorm/sagemaker/sagemaker_partition.py
```diff
@@ -214,7 +214,8 @@ def run_partition(job_config: PartitionJobConfig):
     metadata_filename = job_config.metadata_filename
     skip_partitioning = job_config.skip_partitioning == 'true'
 
-    sm_env = json.loads(os.environ['SM_TRAINING_ENV'])
+    with open("/opt/ml/config/resourceconfig.json", "r", encoding="utf-8") as f:
```
classicsong, line 217: Is SageMaker API changed?

sagemaker/launch/launch_partition.py
```diff
@@ -45,40 +44,43 @@ def run_job(input_args, image):
...
-    est = PyTorch(
-        disable_profiler=True,
-        debugger_hook_config=False,
-        entry_point=os.path.basename(entry_point),
-        source_dir=os.path.dirname(entry_point),
+    script_processor = ScriptProcessor(
         image_uri=image,
         role=role,
-        instance_count=instance_count,
+        instance_count=int(instance_count),
```
classicsong, line 67: Shouldn't instance_count be a int? We can define the type of --instance-count

classicsong (approved):

docker/sagemaker/Dockerfile.sm
thvasilo: I've encountered this particular error when not pinning versions: https://github.com/numba/numba/issues/7339

python/graphstorm/sagemaker/sagemaker_partition.py
thvasilo: No the launch calls remain the same from the user's perspective. The information that exists as env vars in Training exists in this file in Processing jobs.

sagemaker/launch/launch_partition.py
thvasilo: Yeah actually that's already the case, so I'll remove this : https://github.com/awslabs/graphstorm/blob/main/sagemaker/launch/common_parser.py#L33

PR #628, opened 2023-11-08
[GSProcessing] Add thread-parallelism to in-memory repartition implementation.

thvasilo:
Also add more detailed docs about re-partitioning step.

*Issue #, if available:*

*Description of changes:*

* We use thread-parallelism for in-memory file repartition to speed up the process. In our experiments with 10B+ edges, we observe at least 2x speedup.
* Add more detailed documentation about the re-partitioning step, what it does and why.

### DIFF
--- a/docs/source/gs-processing/usage/row-count-alignment.rst
+++ b/docs/source/gs-processing/usage/row-count-alignment.rst
@@ -1,0 +1,143 @@
+Row count alignment
+===================
+
+After the data processing step we need to perform an additional step
+to ensure that our processed data conform to the assumptions of the distributed
+partitioning pipeline. In particular DistPartitioning expects that:
+
+* For each node/edge type:
+    * Every file output has the same number of files.
+        * For example, for an edge type ``x:to:y``, that had
+          two features, ``feat1`` and ``feat2``, the number
+          (partition count) of the files produced separately
+          for ``feat1``, ``feat2`` and the edge structure
+          needs to be the same.
+    * Each respective file in the output has the same row count.
+        * For example, assuming ``feat1``, ``feat2``, and ``edges``
+          had 2 part-files each, the number of rows in file-part-1
+          needs to be the same across all three file sources, and the
+          number of rows in file-part-2 needs to be the same
+          across all three file sources.
+
+
+In code the above means:
+
+.. code-block:: python
+
+    files_for_feat1 = os.listdir("edge_data/x:to:y-feat1/")
+    files_for_feat2 = os.listdir("edge_data/x:to:y-feat2/")
+    files_for_edges = os.listdir("edges/x:to:y-feat2/")
+
+    num_feat1_files = len(files_for_feat1)
+    num_feat2_files = len(files_for_feat2)
+    num_edges_files = len(files_for_edges)
+
+    assert num_feat1_files == num_feat2_files == num_edges_files
+
+In addition, for each node/edge type, the row counts of each respective file
+in their output needs to match, i.e.:
+
+.. code-block:: python
+
+    from pyarrow import parquet as pq
+
+    row_counts_feat1 = [pq.read_metadata(fpath).num_rows for fpath in files_for_feat1]
+    row_counts_feat2 = [pq.read_metadata(fpath).num_rows for fpath in files_for_feat2]
+    row_counts_edges = [pq.read_metadata(fpath).num_rows for fpath in files_for_edges]
+
+    assert row_counts_feat1 == row_counts_feat2 == row_counts_edges
+
+Note that these assumptions only apply `per type`; file counts and per-file
+row counts do not need to match between different node/edge types.
+
+Because of the distributed and speculative nature of Spark execution, it's
+not possible to guarantee that the row counts will match between the file
+outputs we create for every node types features, or the structure and
+features of an edge type.
+
+Therefore and additional step which we call `repartitioning` is necessary
+after the processing step. This step performs two functions:
+
+1. Align the row counts for each edge/node type.
+2. Ensure that data shapes for masks and labels match what
+   what DistPartitioning expects, which are flat ``(N,)`` arrays,
+   instead of what Spark produces which is ``(N, 1)`` Parquet output.
+
+Local repartitioning
+--------------------
+
+The simplest way to apply the re-partitioning step is to do so using a local
+installation of GSProcessing:
+
+.. code-block:: bash
+
+    gs-repartition --input-prefix local_or_s3_path_to_processed_data
+
+The repartitioning command will call the ``graphstorm_processing/repartition_files.py``
+Python script and execute the step locally. The script only requires the
+``input-prefix`` argument to function, but provides optional arguments
+to customize the input/output file names and whether to use an
+in-memory or file streaming implementation for row-count alignment.
+
+You can use `gs-repartition --help` for more details on the arguments.
+
+Repartitioning on SageMaker
+---------------------------
+
+To avoid local processing it is also possible to run re-partitioning on
+SageMaker. You would need to complete the steps described in
+:doc:`distributed-processing-setup` to build and push a SageMaker
+ECR image, and then you're able to launch the re-partitioning job
+on SageMaker:
+
+.. code-block:: bash
+
+    bash docker/build_gsprocessing_image.sh --environment sagemaker --region ${REGION}
+    bash docker/push_gsprocessing_image.sh --environment sagemaker --region ${REGION}
+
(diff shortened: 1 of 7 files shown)


docs/source/gs-processing/usage/row-count-alignment.rst
```diff
@@ -0,0 +1,143 @@
...
+          across all three file sources.
+
+
+In code the above means:
+
+.. code-block:: python
+
+    files_for_feat1 = os.listdir("edge_data/x:to:y-feat1/")
+    files_for_feat2 = os.listdir("edge_data/x:to:y-feat2/")
+    files_for_edges = os.listdir("edges/x:to:y-feat2/")
```
classicsong, line 29:
```suggestion
    files_for_edges = os.listdir("edges/x:to:y")
```

docs/source/gs-processing/usage/row-count-alignment.rst
```diff
@@ -0,0 +1,143 @@
...
+-----------------------------
+
+The default implementation of re-partitioning will load each
+feature/edge type in memory and perform the row-count alignment.
+Using SageMaker Processing with instances such as ``ml.r5.24xlarge``
+with 768GB of memory, you should be able to process data with
+billions of nodes/edges and hundreds of features.
+
+If however your data are so large that they cause out-of-memory
+errors even on SageMaker, you can use the file streaming implementation
```
classicsong, line 117: Can we make streaming processing as the default option? What is the drawback of streaming ?

docs/source/gs-processing/usage/row-count-alignment.rst
```diff
@@ -0,0 +1,143 @@
...
+A similar modification can be applied to the SageMaker launch call:
+
+.. code-block:: bash
+
+    python scripts/run_repartitioning.py --s3-input-prefix ${PROCESSED_OUTPUT} \
+        --role ${ROLE} --image ${IMAGE_URI}  --config-filename "metadata.json" \
+        --instance-type ${INSTANCE_TYPE} --wait-for-job \
+        --streaming-repartitioning True
+
+The file streaming implementation will hold at most 2 files worth of data
```
classicsong, line 136: How doe streaming process decides the number of file partitions?

graphstorm-processing/graphstorm_processing/repartition_files.py
```diff
@@ -135,17 +142,45 @@ def write_parquet_to_relative_path(
 
     @staticmethod
     def create_new_relative_path_from_existing(
-        original_relative_path: str, repartitioned_file_index: int
+        original_relative_path: str, repartitioned_file_index: int, suffix: str = None
     ) -> str:
-        """
+        """Changes the index in the `original_relative_path` to `original_relative_path`.
```
classicsong, line 147: What does this mean?

graphstorm-processing/graphstorm_processing/repartition_files.py
```diff
@@ -135,17 +142,45 @@ def write_parquet_to_relative_path(
...
+
         Given a path of the form 'path/to/parquet/part-00001-filename.snappy.parquet', changes the
         numerical part of the filename to match the provided `repartitioned_file_index`,
-        and changes the path prefix to 'path/to/parquet-repartitioned/'.
+        and changes the path prefix to 'path/to/parquet-repartitioned/', or
+        'path/to/parquet-repartitioned-{suffix}/' if `suffix` is provided.
```
classicsong, line 152: Can you move the examples in L176 here?

classicsong: It seems we will also add a uuid if suffix is not provided, but this is not mentioned in the docstring here.

graphstorm-processing/graphstorm_processing/repartition_files.py
```diff
@@ -260,31 +309,37 @@ def repartition_parquet_files_in_memory(
...
+        zero_copy_slices = [
+            table.slice(offset=offset, length=desired_count)
+            for offset, desired_count in zip(offsets, desired_counts)
+        ]
+        uid_for_entry = uuid.uuid4().hex[:8]
+        relative_paths = [
+            self.create_new_relative_path_from_existing(datafile_list[0], idx, uid_for_entry)
+            for idx in range(len(desired_counts))
+        ]
+        with Parallel(n_jobs=min(16, os.cpu_count()), verbose=10, prefer="threads") as parallel:
```
classicsong, line 324: what if len(desired_counts) < 16? 
Parallel will automatically handle it?

docs/source/gs-processing/usage/row-count-alignment.rst
thvasilo: It's significantly slower. As an example, for a dataset with 10B+ edges where all the edge-related files needed to be re-aligned (4x10B rows repartitioned) in-memory took 25min vs. 6 hours for file streaming.

docs/source/gs-processing/usage/row-count-alignment.rst
thvasilo: The number of file partitions is determined during processing, and that's something that we seem to be able to guarantee from Spark (i.e. Spark always produces the same number of files/partitions for the same node/edge type). In the in-memory implementation we could output arbitrary file counts if we wanted, but for now we're just sticking to what existed before.

graphstorm-processing/graphstorm_processing/repartition_files.py
thvasilo: I'll need to fix this. What I wanted to say is that we change the file index indicator in `original_relative_path`, which will be something like `original/path/part-00003.parquet` to `original/path/part-00001.parquet` if `repartitioned_file_index` was 1.

graphstorm-processing/graphstorm_processing/repartition_files.py
thvasilo: Yeah it will launch as many tasks as needed.

graphstorm-processing/graphstorm_processing/repartition_files.py
thvasilo: > It seems we will also add a uuid if suffix is not provided, but this is not mentioned in the docstring here.

Not sure if I understand correctly, the function has

```python
new_sub_path = (
            "parquet-repartitioned" if suffix is None else f"parquet-repartitioned-{suffix}"
        )
```

So unless a `suffix` is provided, only `parquet-repartitioned` is added. In our calls to the function in this file we do add a `uuid4()` suffix, but the function itself does not a suffix unless one is provided.

thvasilo: > Can you move the examples in L176 here?

I'm trying to be consistent with Numpy docstring defined order that always has `Examples` as the last section of the docstring: https://numpydoc.readthedocs.io/en/latest/format.html#sections

docs/source/gs-processing/usage/row-count-alignment.rst
classicsong: I c.

docs/source/gs-processing/usage/row-count-alignment.rst
classicsong: Do we have anywhere explicitly tells users this information?

graphstorm-processing/graphstorm_processing/repartition_files.py
classicsong: I c.
But can you extend the docstring to include an example like what you commented here?

graphstorm-processing/graphstorm_processing/repartition_files.py
classicsong: Now I c what you man regarding to the suffix.

graphstorm-processing/graphstorm_processing/repartition_files.py
classicsong:
```suggestion
        """Changes the file index in the `original_relative_path` to `repartitioned_file_index`.
```

classicsong (approved):

docs/source/gs-processing/usage/row-count-alignment.rst
thvasilo: We have this in the SageMaker guide: https://graphstorm.readthedocs.io/en/latest/gs-processing/usage/amazon-sagemaker.html#the-num-output-files-parameter

Maybe I should make it more prominent?

graphstorm-processing/graphstorm_processing/repartition_files.py
thvasilo: Already changed in code, resolving.

PR #633, opened 2023-11-09
Update GraphStorm SageMaker solution to support Node ID remapping.

classicsong:
*Issue #, if available:*
#508 

*Description of changes:*
Update SageMaker support to support automatic node ID remapping.


sagemaker/launch/common_parser.py
```diff
@@ -5,6 +5,10 @@
 from ast import literal_eval
 import argparse
 
+# Npte: Sync with graphstorm.config SUPPORTED_TASKS
```
jalencato, line 8: Note：

jalencato: Btw, why not directly import from graphstorm.config

sagemaker/run/infer_entry.py
```diff
@@ -50,6 +50,8 @@ def parse_train_args():
     parser.add_argument("--custom-script", type=str, default=None,
         help="Custom training script provided by a customer to run customer training logic. \
             Please provide the path of the script within the docker image")
+    parser.add_argument("--output-chunk-size", type=int, default=100000,
+        help="Number of rows per chunked prediction result or node embedding file.")
```
jalencato, line 54: Do we need to do something simlar for do-nid-remap？

sagemaker/launch/common_parser.py
classicsong: The launch script will not rely on graphstorm package.

sagemaker/run/infer_entry.py
classicsong: No. We have predefined behavior: Training will with --do-nid-remap False and Inference will with --do-nid-remap True.

jalencato (approved):
LGTM, will add to regression test later.

zheng-da (approved):

PR #635, opened 2023-11-10
[GSProcessing] Don't produce masks when no labels present, other small fixes

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Change default behavior when no labels are given, from creating LP masks for all edge types, to creating no masks.
* Fixes to docs formatting.
* Fix broken re-partitioning launch argument.
* Add scipy dependency to SageMaker Dockerfile
* Allow custom tags for image names

### DIFF
--- a/graphstorm-processing/docker/build_gsprocessing_image.sh
+++ b/graphstorm-processing/docker/build_gsprocessing_image.sh
@@ -22,6 +22,7 @@ Available options:
 -p, --path          Path to graphstorm-processing directory, default is the current directory.
 -i, --image         Docker image name, default is 'graphstorm-processing'.
 -v, --version       Docker version tag, default is the library's current version (`poetry version --short`)
+-s, --suffix        Suffix for the image tag, can be used to push custom image tags. Default is "".
 -b, --build         Docker build directory, default is '/tmp/`
 EOF
   exit
@@ -46,6 +47,7 @@ parse_params() {
   BUILD_DIR='/tmp'
   TARGET='test'
   ARCH='x86_64'
+  SUFFIX=""
 
   while :; do
     case "${1-}" in
@@ -78,6 +80,10 @@ parse_params() {
       ;;
     -v | --version)
       VERSION="${2-}"
+      shift
+      ;;
+    -s | --suffix)
+      SUFFIX="${2-}"
       shift
       ;;
     -?*) die "Unknown option: $1" ;;
@@ -128,6 +134,7 @@ msg "- TARGET: ${TARGET}"
 msg "- GSP_HOME: ${GSP_HOME}"
 msg "- IMAGE_NAME: ${IMAGE_NAME}"
 msg "- VERSION: ${VERSION}"
+msg "- SUFFIX: ${SUFFIX}"
 
 # Prepare Docker build directory
 rm -rf "${BUILD_DIR}/docker/code"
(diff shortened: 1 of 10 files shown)


graphstorm-processing/docker/build_gsprocessing_image.sh
```diff
@@ -22,6 +22,7 @@ Available options:
 -p, --path          Path to graphstorm-processing directory, default is the current directory.
 -i, --image         Docker image name, default is 'graphstorm-processing'.
 -v, --version       Docker version tag, default is the library's current version (`poetry version --short`)
+-s, --suffix        Suffix for the image tag, can be used to push custom image tags. Default is "".
```
classicsong, line 25: Why we need a suffix?
Users can customize IMAGE_NAME, right?

graphstorm-processing/docker/push_gsprocessing_image.sh
```diff
@@ -45,6 +46,7 @@ parse_params() {
   REGION=${REGION:-us-west-2}
   ACCOUNT=$(aws sts get-caller-identity --query Account --output text)
   ARCH='x86_64'
+  SUFFIX=""
```
classicsong, line 49: ditto

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -325,6 +308,8 @@ def _initialize_metadata_dict(
         return metadata_dict
 
     def _finalize_graphinfo_dict(self, metadata_dict: Dict) -> Dict:
+        if self.skip_train_masks:
+            self.graph_info["task_type"] = "link_prediction"
```
classicsong, line 312: Can we add a warning here to tell people we automatically set the task type to link prediction?

graphstorm-processing/tests/test_dist_heterogenous_loader.py
```diff
@@ -264,15 +263,8 @@ def test_load_dist_hgl_without_labels(dghl_loader_no_label: DistHeterogeneousGra
...
-        "task_type": "link_predict",
-        "etype_label": [
-            "movie:included_in:genre",
-            "genre:included_in-rev:movie",
-            "user:rated:movie",
-            "movie:rated-rev:user",
-            "director:directed:movie",
-            "movie:directed-rev:director",
-        ],
+        "task_type": "link_prediction",
```
classicsong, line 266: I think we should keep the original test and add a new test with `"etype_label": [],`

graphstorm-processing/docker/build_gsprocessing_image.sh
thvasilo: They can customize the name, but that would lead to the creation of another ECR repository. We want to give users the option to customize just the tag.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: We have a warning printed in line 165, I can add that the registered task will be LP.

graphstorm-processing/tests/test_dist_heterogenous_loader.py
thvasilo: Could you clarify what each test would do? The test (`test_load_dist_hgl_without_labels`) checks the default behavior when no label is provided. Before we would generate masks for all edge types, leading to all types having "labels". Since we changed this default behavior we needed to change the test as well.

classicsong: I c

classicsong (approved):

PR #637, opened 2023-11-12
Fix bug in constructing node features

zheng-da:
*Description of changes:*
* add the support of constructing node features on featureless nodes in GLEM.
* fix a bug in sampling neighbors only with relations that connect to neighbors with features.


python/graphstorm/model/gnn_with_reconstruct.py
```diff
@@ -61,7 +61,10 @@ def construct_node_feat(g, rel_names, input_gnn, get_input_embeds, batch_size,
                                             partition_book=g.get_partition_book(),
                                             ntype=ntype, force_even=False)
         # We use all neighbors to reconstruct node features.
-        sampler = dgl.dataloading.MultiLayerNeighborSampler([-1], mask=edge_mask)
+        fanout = {rel_name: 0 for rel_name in g.canonical_etypes}
+        for rel_name in rel_names:
+            fanout[rel_name] = -1
```
classicsong, line 66: I am a little bit confused, in GSgnnNodeSemiSupDataLoader, we have construct_feat_fanout, but here we use full neighbors. Why not use construct_feat_fanout here?

tests/unit-tests/data_utils.py
```diff
@@ -248,6 +248,8 @@ def generate_dummy_hetero_graph_reconstruct(size='tiny', gen_mask=True):
                              th.randint(data_size, (data_size,))),
         ("n4", "r3", "n2"): (th.randint(data_size, (data_size,)),
                              th.randint(data_size, (data_size,))),
+        ("n1", "r5", "n2"): (th.randint(data_size, (data_size,)),
```
classicsong, line 251: This can trigger the bug? Why?

zheng-da: we want to generate features on n2. the bug happens when the target nodes have two relation types. one relation connects neighbors with features and the other doesn't.

python/graphstorm/model/gnn_with_reconstruct.py
zheng-da: this function is for inference. you can consider they are two bug fixes. we can split the PR into two PRs.

classicsong (approved):

PR #638, opened 2023-11-12
In GLEM LM training phase, GNN use cached LM embeddings instead of hot ones

wangz10:
*Issue #, if available:*

*Description of changes:*
GLEM iteratively trains LM and GNN. GNN depends on LM's embedding for its prediction, making it difficult to distill GNN into LM with predictions from GNN on pseudolabels. In this PR, LM training phase still maintain the LM cache for GNN to generate pseudolabels. 

- extend the `lm_embed.forward` with `use_cache` arg to override the forward computation even if cache is available
- changed the LM cache control in `GLEM.toggle`
- changed `GLEM._embed_nodes` to always use cached LM embeddings for GNN

### DIFF
--- a/python/graphstorm/model/node_glem.py
+++ b/python/graphstorm/model/node_glem.py
@@ -264,9 +264,11 @@ class GLEM(GSgnnNodeModelBase):
             self.training_lm = True
             self.toggle_params('gnn', True)
             self.toggle_params('lm', False)
-            # when training lm, do not use the cached LM
-            self.lm.unfreeze_input_encoder()
+            # when training lm: gnn still uses the cached LM embeddings from previous iteration
+            # hence we don't unfreeze nor update the cache
         elif part == 'gnn':
+            # clear previous cache
+            self.lm.unfreeze_input_encoder()
             self.training_lm = False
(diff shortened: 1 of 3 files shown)


python/graphstorm/model/node_glem.py
```diff
@@ -264,9 +264,11 @@ def toggle(self, part='lm', data=None):
             self.training_lm = True
             self.toggle_params('gnn', True)
             self.toggle_params('lm', False)
-            # when training lm, do not use the cached LM
-            self.lm.unfreeze_input_encoder()
+            # when training lm: gnn still uses the cached LM embeddings from previous iteration
+            # hence we don't unfreeze nor update the cache
         elif part == 'gnn':
+            # clear previous cache
+            self.lm.unfreeze_input_encoder()
```
zheng-da, line 271: if we train GNN part, why do we unfreeze the input encoder?

wangz10: It is later frozen (line 276). This unfreeze call is to clear the outdated LM cache that was used for training LM from the previous iteration.

PR #639, opened 2023-11-13
Use external-memory arrays in the worker processes of gconstruct

zheng-da:
*Description of changes:*
When using external memory to store the node/edge features, we should keep the output data from the worker processes in the external memory as well.


python/graphstorm/gconstruct/utils.py
```diff
@@ -36,6 +36,37 @@
...
+EXT_MEMORY_STORAGE = "ext_memory"
+
+def _to_ext_memory(name, data, path):
+    if isinstance(data, np.ndarray):
+        assert name is not None
+        path = os.path.join(path, f"{name}.npy")
+        ext_mem_arr = convert_to_ext_mem_numpy(path, data)
+        # We need to pass the array to another process. We don't want it
+        # to reference to data in the file.
+        ext_mem_arr.cleanup()
```
classicsong, line 48: Should we just return the path to the file instead of the object?

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -216,25 +217,30 @@ def _process_data(user_pre_parser, user_parser,
...
+                                                  ext_mem_workspace)
         update_two_phase_feat_ops(phase_one_ret, two_phase_feat_ops)
 
         dur = time.time() - pre_parse_start
         logging.debug("Preprocessing data files for %s takes %.3f seconds.",
                       task_info, dur)
 
     start = time.time()
-    return_dict = multiprocessing_data_read(in_files, num_proc, user_parser)
+    return_dict = multiprocessing_data_read(in_files, num_proc, user_parser,
```
classicsong, line 234: Have you implemented any logic to load the ext_mem objects as you flushed the ExtMemArrayWrapper._arr and set it into None.

python/graphstorm/gconstruct/utils.py
zheng-da: ext_mem_arr is a ExtNumpyWrapper. after cleaning up, the main thing it references is the path of the file and related metadata (shape, dtype).

python/graphstorm/gconstruct/construct_graph.py
zheng-da: ExtNumpyWrapper loads the data from the file whenever data are accessed.

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -313,14 +319,17 @@ def process_node_data(process_confs, arr_merger, remap_id, ext_mem=None, num_pro
                               label_ops=label_ops,
                               node_id_col=node_id_col,
                               read_file=read_file,
-                              ext_mem=ext_mem)
+                              ext_mem=ext_mem_workspace)
 
+        ext_mem_workspace_type = os.path.join(ext_mem_workspace, node_type) \
```
classicsong, line 324:
```suggestion
        ext_mem_workspace_ntype = os.path.join(ext_mem_workspace, node_type) \
```

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -493,14 +502,17 @@ def process_edge_data(process_confs, node_id_map, arr_merger,
                               read_file=read_file,
                               conf=process_conf,
                               skip_nonexist_edges=skip_nonexist_edges,
-                              ext_mem=ext_mem)
+                              ext_mem=ext_mem_workspace)
 
+        ext_mem_workspace_type = os.path.join(ext_mem_workspace, "_".join(edge_type)) \
```
classicsong, line 507:
```suggestion
        ext_mem_workspace_etype = os.path.join(ext_mem_workspace, "_".join(edge_type)) \
```

classicsong (approved):
LGTM

classicsong:
Can you check whether https://github.com/awslabs/graphstorm/pull/654 is related?

PR #641, opened 2023-11-14
Refactor the way we save and load ID mappings for GConstruct.

thvasilo:
*Issue #, if available:*

*Description of changes:*

* This PR aligns the way node id mapping files are produced and loaded by GSF to that of GSProcessing
* This is needed in order to ensure that downstream tasks can load mappings that were created either by GSProcessing or GConstruct.

Specifically:

* Mappings are created with one directory per-node-type, under a `raw_mappings_id` directory. 
* Each node mappings consists of one or more Parquet files. We try to split the files s.t. no individual file is larger than 1GB.

Breaking changes:

* Previously, GConstruct created the ID mappings as a single Parquet file, with its filename prefixed by the node type. With this change we create partitioned Parquet files, each node type under its own directory. The difference in behavior is exemplified below:

image


python/graphstorm/gconstruct/file_io.py
```diff
@@ -200,15 +200,15 @@ def read_data_parquet(data_file, data_fields=None):
     """
     table = pq.read_table(data_file)
     data = {}
-    df_table = table.to_pandas()
+    df_table: pd.DataFrame = table.to_pandas()
```
classicsong, line 203: Let keep it as it was. 
We do not apply strong typing in graphstorm code right now.

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -742,11 +742,11 @@ def process_graph(args):
     if len(edge_label_stats) > 0:
         save_edge_label_stats(args.output_dir, edge_label_stats)
 
-    for ntype in node_id_map:
-        map_file = os.path.join(args.output_dir, ntype + "_id_remap.parquet")
-        if node_id_map[ntype].save(map_file):
-            logging.info("Graph construction generates new node IDs for '%s'. " + \
-                    "The ID map is saved in %s.", ntype, map_file)
+    for ntype, node_id_map in node_id_map.items():
+        map_prefix = os.path.join(args.output_dir, "node_id_mappings", ntype)
```
classicsong, line 746:
```suggestion
        map_prefix = os.path.join(args.output_dir, "node_id_remappings", ntype)
```

python/graphstorm/gconstruct/remap_result.py
```diff
@@ -504,7 +504,7 @@ def _parse_gs_config(config):
         list of str: etypes that have prediction results
     """
     part_config = config.part_config
-    node_id_mapping = os.path.dirname(part_config)
+    node_id_mapping = os.path.join(os.path.dirname(part_config), "node_id_mappings")
```
classicsong, line 507:
```suggestion
    node_id_mapping = os.path.join(os.path.dirname(part_config), "node_id_remappings")
```

python/graphstorm/sagemaker/utils.py
```diff
@@ -242,6 +245,18 @@ def download_graph(graph_data_s3, graph_name, part_id, local_path, sagemaker_ses
 
     graph_data_s3 = graph_data_s3[:-1] if graph_data_s3.endswith('/') else graph_data_s3
 
+    # By default we assume the node mappings exist
+    # under the same path as the rest of the graph data
+    if not node_mapping_prefix_s3:
+        node_mapping_prefix_s3 = f"{graph_data_s3}/node_id_mappings"
```
classicsong, line 251:
```suggestion
        node_mapping_prefix_s3 = f"{graph_data_s3}/node_id_remappings"
```

python/graphstorm/sagemaker/utils.py
```diff
@@ -242,6 +245,18 @@ def download_graph(graph_data_s3, graph_name, part_id, local_path, sagemaker_ses
...
+    else:
+        node_mapping_prefix_s3 = (
+            node_mapping_prefix_s3[:-1] if node_mapping_prefix_s3.endswith('/')
+            else node_mapping_prefix_s3)
+        assert node_mapping_prefix_s3.endswith("node_id_mappings"), \
```
classicsong, line 256:
```suggestion
        assert node_mapping_prefix_s3.endswith("node_id_remappings"), \
```

python/graphstorm/gconstruct/construct_graph.py
classicsong: Let's keep the name id_remap

python/graphstorm/sagemaker/utils.py
```diff
@@ -242,6 +245,18 @@ def download_graph(graph_data_s3, graph_name, part_id, local_path, sagemaker_ses
...
+            "node_mapping_prefix_s3 must end with 'node_id_mappings'"
```
classicsong, line 257:
```suggestion
            "node_remapping_prefix_s3 must end with 'node_id_remappings'"
```

python/graphstorm/sagemaker/utils.py
```diff
@@ -225,6 +226,8 @@ def download_graph(graph_data_s3, graph_name, part_id, local_path, sagemaker_ses
         Path to store graph data
     sagemaker_session: sagemaker.session.Session
         sagemaker_session to run download
+    node_mapping_prefix_s3: str, optional
```
classicsong, line 229:
```suggestion
    node_remapping_prefix_s3: str, optional
```
Let's call it remapping.

tests/end2end-tests/graphstorm-ec/mgpu_test.sh
```diff
@@ -89,7 +89,15 @@ then
 fi
 
 echo "**************dataset: Generated multilabel MovieLens EC, do inference on saved model"
-python3 -m graphstorm.run.gs_edge_classification --inference --workspace $GS_HOME/inference_scripts/ep_infer --num-trainers $NUM_INFO_TRAINERS --num-servers 1 --num-samplers 0 --part-config /data/movielen_100k_multi_label_ec/movie-lens-100k.json --ip-config ip_list.txt --ssh-port 2222 --cf ml_ec_infer.yaml  --multilabel true --num-classes 6 --node-feat-name movie:title user:feat --use-mini-batch-infer false --save-embed-path /data/gsgnn_ec/infer-emb/ --restore-model-path /data/gsgnn_ec/epoch-$best_epoch/ --save-prediction-path /data/gsgnn_ec/prediction/ --logging-file /tmp/log.txt  --logging-level debug --preserve-input True
+python3 -m graphstorm.run.gs_edge_classification --inference --workspace $GS_HOME/inference_scripts/ep_infer \
```
classicsong, line 92: Can you revert the change? It is very hard to see the difference.

thvasilo: It's only introducing line breaks to make it easier to tell apart changes in the future, but I'll revert this for now. We could do this in a formatting PR.

classicsong (approved):

sagemaker/launch/launch_train.py
```diff
@@ -128,13 +131,14 @@ def get_train_parser():
...
     training_args.add_argument("--graph-name", type=str, help="Graph name",
         required=True)
-    training_args.add_argument("--graph-data-s3", type=str,
-        help="S3 location of input training graph", required=True)
     training_args.add_argument("--task-type", type=str,
         help=f"Task type in {SUPPORTED_TASKS}", required=True)
     training_args.add_argument("--yaml-s3", type=str,
         help="S3 location of training yaml file. "
              "Do not store it with partitioned graph", required=True)
+    training_args.add_argument("--node-mappings-s3", type=str,
```
classicsong, line 139: In train_infer you use --node-id-mappings-s3, but here you use --node-mappings-s3.
BTW, I don't think we need to update launch_train.py here.
In SageMaker, save-embedding and save-prediction are always set to False.

sagemaker/launch/launch_infer.py
```diff
@@ -154,8 +146,11 @@ def get_inference_parser():
              "Do not store it with partitioned graph",
              required=True)
     inference_args.add_argument("--model-artifact-s3", type=str,
-        help="S3 bucket to load the saved model artifacts",
+        help="S3 location to load the saved model artifacts from",
         required=True)
+    inference_args.add_argument("--node-id-mappings-s3", type=str,
```
classicsong, line 151:
```suggestion
    inference_args.add_argument("--node-mappings-s3", type=str,
```

sagemaker/run/train_entry.py
```diff
@@ -39,6 +39,8 @@ def parse_train_args():
              "Do not store it with partitioned graph")
     parser.add_argument("--model-artifact-s3", type=str,
         help="S3 location to store the model artifacts.")
+    parser.add_argument("--node-mappings-s3", type=str, required=False,
```
classicsong, line 42: We should add `--node-mappings-s3` to infer_entry.py instead of train_entry.py

python/graphstorm/sagemaker/sagemaker_train.py
```diff
@@ -141,6 +141,8 @@ def run_train(args, unknownargs):
             S3 location of training yaml file.
         model_artifact_s3: str
             S3 location to store the model artifacts.
+        node_mappings_s3: str
```
classicsong, line 144: We do not generate node embeddings or do prediction in training.
So we do not need to change sagemaker_train.py

sagemaker/launch/launch_infer.py
```diff
@@ -68,24 +68,19 @@ def run_job(input_args, image, unknownargs):
 
     prefix = f"gs-infer-{graph_name}"
 
+    params = {
+        "graph-data-s3": graph_data_s3,
+        "graph-name": graph_name,
+        "infer-yaml-s3": infer_yaml_s3,
+        "model-artifact-s3": model_artifact_s3,
+        "raw-node-mappings-s3": input_args.raw_node_mappings_s3,
```
jalencato, line 76: If we directly add ```raw-node-mappings-s3``` argument here, when it is empty. ```launch/launch_infer.py``` will call ```run/infer_entry.py``` like ```python3 run/infer_entry.py --raw-node-mappings-s3 ``` which will cause an error. Suggest to do what we do here: https://github.com/awslabs/graphstorm/blob/8380f8c7fe35dede91315fa7223eb13baf1d0010/sagemaker/launch/launch_train.py#L72-L75

python/graphstorm/gconstruct/id_map.py
```diff
@@ -74,13 +73,19 @@ class IdReverseMap:
...
-        assert os.path.exists(id_map_path), \
-            f"{id_map_path} does not exits."
-        data = read_data_parquet(id_map_path, ["new", "orig"])
+    def __init__(self, id_map_prefix):
+        assert os.path.exists(id_map_prefix), \
+            f"{id_map_prefix} does not exist."
+        try:
+            data = read_data_parquet(id_map_prefix, ["orig", "new"])
+        except AssertionError:
+            data = read_data_parquet(id_map_prefix, ["node_str_id", "node_int_id"])
```
classicsong, line 85: Can you add a comment here:
```# for backward compatible with graphstorm v0.2.1 . ```

File: python/graphstorm/gconstruct/id_map.py [hunk 15]
```diff
@@ -192,23 +197,32 @@ def map_id(self, ids):
...
-        vals = list(self._ids.values())
-        table = pa.Table.from_pandas(pd.DataFrame({'orig': keys, 'new': vals}))
-        pq.write_table(table, file_path)
-        return True
+        os.makedirs(file_prefix, exist_ok=True)
+        table = pa.Table.from_arrays([pa.array(self._ids.keys()), self._ids.values()],
+                                     names=["orig", "new"])
+        bytes_per_row = table.nbytes // table.num_rows
+        # Split table in parts, such that the max expected file size is ~1GB
+        max_rows_per_file = (1024**3) // bytes_per_row
```
classicsong, line 217: Let's make 1024**3 a constant variable.

tests/unit-tests/gconstruct/test_construct_graph.py
```diff
@@ -1701,10 +1701,63 @@ def test_feature_wrapper():
...
-        "Directory /tmp_featurewrapper should not exist after gc"
-    assert not os.path.isdir("/tmp_featurewrapper2"), \
-        "Directory /tmp_featurewrapper2 should not exist after gc"
+    assert not os.path.isdir("/tmp/featurewrapper"), \
+        "Directory /tmp/featurewrapper should not exist after gc"
+    assert not os.path.isdir("/tmp/featurewrapper2"), \
+        "Directory /tmp/featurewrapper2 should not exist after gc"
+
+
+def test_homogeneous():
```
classicsong, line 1710: What is the purpose this this test?
Does the test belong to this PR?

thvasilo: I think was added during the conflict resolution, I'll remove it.

python/graphstorm/sagemaker/utils.py
```diff
@@ -243,6 +246,18 @@ def download_graph(graph_data_s3, graph_name, part_id, world_size,
...
+    # By default we assume the node mappings exist
+    # under the same path as the rest of the graph data
+    if not raw_node_mapping_prefix_s3:
+        raw_node_mapping_prefix_s3 = f"{graph_data_s3}/raw_id_mappings"
+    else:
+        raw_node_mapping_prefix_s3 = (
+            raw_node_mapping_prefix_s3[:-1] if raw_node_mapping_prefix_s3.endswith('/')
+            else raw_node_mapping_prefix_s3)
+        assert raw_node_mapping_prefix_s3.endswith("raw_id_mappings"), \
+            "node_mapping_prefix_s3 must end with 'raw_id_mappings'"
```
thvasilo, line 258: @classicsong do you think this requirement is valid? I think I'd rather remove it to allow for more flexibility of how/where mappings are created.

classicsong: We can remove this assert, it is not necessary that the s3 prefix end with raw_id_mappings/

tests/unit-tests/gconstruct/test_construct_graph.py
classicsong: @jalencato Do you know what test is it?

PR #645, opened 2023-11-15
Support output data in csv format in remap_result.py

classicsong:
CSV a very common data format that accepted by many structured data storage systems. When outputing results (prediction results or node embeddings), we should support csv format.


python/graphstorm/gconstruct/remap_result.py
```diff
@@ -49,8 +54,51 @@
...
+    output_fname = f"{file_prefix}.csv"
+    csv_data = {}
+    for key, vals in data.items():
+        if len(vals.shape) == 1:
+            csv_data[key] = vals.tolist()
+        elif len(vals.shape) == 2:
+            csv_data[key] = [";".join([str(v) for v in val]) \
+                             for val in vals.tolist()]
+    data_frame = pd.DataFrame(data)
+    data_frame.to_csv(output_fname, index=False, sep=delimiter)
```
sojiadeshina, line 98: does this only save the vector data or does it save the id column too?

classicsong: Yes.

python/graphstorm/gconstruct/remap_result.py
```diff
@@ -49,8 +66,62 @@
 # id_maps to each worker process.
 id_maps = {}
 
+def write_data_parquet_file(data, file_prefix, col_name_map=None):
```
zheng-da, line 69: i think i have these functions in gconstruct. should we merge them?

python/graphstorm/gconstruct/remap_result.py
```diff
@@ -786,9 +888,17 @@ def generate_parser():
...
-                       choices=['parquet'],
+                       default=GS_OUTPUT_FORMAT_PARQUET,
+                       choices=[GS_OUTPUT_FORMAT_PARQUET, GS_OUTPUT_FORMAT_CSV],
                        help="The format of the output.")
+    group.add_argument("--output-delimiter", type=str, default=",",
+                       help="The delimiter used when saving data in CSV format.")
+    group.add_argument("--column-names", type=str, nargs="+", default=None,
+                       help="Defines how to rename default column names to new names."
+                       f"For example, given --column-names {GS_REMAP_NID_COL}:~id "
+                       f"{GS_REMAP_EMBED_COL}:embedding. The column "
```
sojiadeshina, line 900: will it be clear what GS_REMAP_NID_COL and GS_REMAP_EMBED_COL variable names mean?

python/graphstorm/gconstruct/remap_result.py
classicsong: This function calls the write_data_parquet in gconstruct.

python/graphstorm/gconstruct/remap_result.py
classicsong: We will create build a doc to introduce the output of save embeddings.

python/graphstorm/gconstruct/remap_result.py
```diff
@@ -49,8 +66,62 @@
...
+        data = {col_name_map[key]: val for key, val in data.items()}
+
+    output_fname = f"{file_prefix}.csv"
+    csv_data = {}
+    for key, vals in data.items():
+        if len(vals.shape) == 1:
+            csv_data[key] = vals.tolist()
+        elif len(vals.shape) == 2:
+            csv_data[key] = [";".join([str(v) for v in val]) \
+                             for val in vals.tolist()]
```
zheng-da, line 119: what is this? `val` is a float?

classicsong: val can be any primitive. If write_data_csv_file is called for saving embeddings, val will be a float.
If write_data_csv_file is called for saving prediction results, it can be integer.

zheng-da (approved):

Issue #642, opened 2023-11-14
[Feature request] support inductive link-prediction splits in `partition_graph_lp.py`


wangz10:

classicsong:
Can you be more specific of what does inductive setting mean?

wangz10:
@classicsong In inference time of inductive setting, there are new/unseen nodes from the training graph. The link-prediction model is tasked to predict links 1) between an unseen node and a seen node; or 2) between two unseen nodes.

PR #648, opened 2023-11-15
Implemented inductive splits for link prediction

wangz10:
*Issue #, if available:* #642

*Description of changes:*
 - Added a flag `--inductive-split` in `partition_graph_lp.py` to support split links in inductive setting.

### DIFF
--- a/tools/partition_graph_lp.py
+++ b/tools/partition_graph_lp.py
@@ -18,6 +18,7 @@
 
 import os
 import dgl
+import numpy as np
 import torch as th
 import argparse
 import time
@@ -44,6 +45,9 @@ if __name__ == '__main__':
                            help='The pct of train nodes/edges. Should be > 0 and < 1.')
     argparser.add_argument('--val-pct', type=float, default=0.1,
                            help='The pct of validation nodes/edges. Should be > 0 and < 1.')
+    argparser.add_argument('--inductive-split', action='store_true',
+                           help='split links for inductive settings: no overlapping nodes across '
+                           + 'splits.')
     # graph modification arguments
     argparser.add_argument('--add-reverse-edges', action='store_true',
                            help='turn the graph into an undirected graph.')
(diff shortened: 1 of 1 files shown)


tools/partition_graph_lp.py
```diff
@@ -149,15 +156,43 @@
...
-                                               int(num_edges * (args.train_pct + args.val_pct))] = True
-            g.edges[target_e].data['test_mask'][int(num_edges * (args.train_pct + args.val_pct)): ] = True
+            if not args.inductive_split:
+                # Randomly split links
+                g.edges[target_e].data['train_mask'][: int(num_edges * args.train_pct)] = True
+                g.edges[target_e].data['val_mask'][int(num_edges * args.train_pct): \
+                                                int(num_edges * (args.train_pct + args.val_pct))] = True
+                g.edges[target_e].data['test_mask'][int(num_edges * (args.train_pct + args.val_pct)): ] = True
+            else:
+                # Inductive split for link prediction
```
classicsong, line 172: Does this mean some head nodes are not available to the training graph?
Can you add some details here?

tools/partition_graph_lp.py
```diff
@@ -149,15 +156,43 @@
...
+                                        int(num_nodes * (args.train_pct + args.val_pct))]
+                test_u = shuffled_index[int(num_nodes * (args.train_pct + args.val_pct)): ]
+                # 2. find all out-edges for the sets of head nodes:
+                _, train_v, train_eids = g.out_edges(train_u, form='all', etype=target_e)
+                val_eids = g.out_edges(val_u, form='eid', etype=target_e)
+                test_eids = g.out_edges(test_u, form='eid', etype=target_e)
+                if utype == vtype:
+                    # we remove edges with tail nodes outside of the training set
+                    # this isn't necessary if head and tail are different types
+                    train_eids = train_eids[np.in1d(train_v, train_u)]
```
classicsong, line 191: What if there are some overlap between val and test?

tools/partition_graph_lp.py
wangz10: > Does this mean some head nodes are not available to the training graph?
Yes. 
I'll add more comments.

tools/partition_graph_lp.py
wangz10: You're right. Removed val/test overlap in the next commit.

tools/partition_graph_lp.py
```diff
@@ -149,15 +156,47 @@
...
-            g.edges[target_e].data['val_mask'][int(num_edges * args.train_pct): \
-                                               int(num_edges * (args.train_pct + args.val_pct))] = True
-            g.edges[target_e].data['test_mask'][int(num_edges * (args.train_pct + args.val_pct)): ] = True
+            if not args.inductive_split:
+                # Randomly split links
+                g.edges[target_e].data['train_mask'][: int(num_edges * args.train_pct)] = True
+                g.edges[target_e].data['val_mask'][int(num_edges * args.train_pct): \
+                                                int(num_edges * (args.train_pct + args.val_pct))] = True
+                g.edges[target_e].data['test_mask'][int(num_edges * (args.train_pct + args.val_pct)): ] = True
+            else:
```
zheng-da, line 171: i'm trying to understand why this is useful. this split is to avoid the endpoint nodes of the training/validation/test edges are not overlapped. However, this cannot guarantee that the nodes in the message passing in training and validation have overlap. do we need to avoid the overlap of nodes in the message passing?

wangz10: > However, this cannot guarantee that the nodes in the message passing in training and validation have overlap

This is true. However the tail nodes overlap is one-way: val set can contain tail nodes from training set, but training set doesn't contain tail nodes from val set. Therefore, the message passing in training set will not leak to unseen nodes in val/test sets. This also allows semi-inductive retrieval: given a new head nodes, retrieve a seen tail nodes from training set.

tools/partition_graph_lp.py
```diff
@@ -149,15 +156,47 @@
...
+                # Inductive split for link prediction
+                # 1. split the head nodes u into three disjoint sets (train/val/test)
+                # such that model will be evaluted to predict links for unseen nodes
```
zheng-da, line 174: please add a comment to explain how this can make the link prediction training and evaluation completely inductive.

classicsong (approved):

PR #653, opened 2023-11-16
[Bug Fix] Fix bug when loading ID mapping generated by dgl convert_partition.py

classicsong:
For Graph ID <-> Partition ID mapping files generated by dgl distributed partition script, each orig_nids.dgl file only stores the mapping of partition local IDs to Graph ID. It is required to concatenate all the orig_nids.dgl files from all the partitions to build the global Graph ID <-> Partition ID mapping .


python/graphstorm/model/utils.py
```diff
@@ -402,11 +423,12 @@ def distribute_nid_map(embeddings, rank, world_size,
...
-            if node_id_mapping_file.endswith("dgl"):
-                # node id mapping file from dgl tools/distpartitioning/convert_partition.py.
-                ori_node_id_mapping = dgl.data.utils.load_tensors(node_id_mapping_file)
-            else: # endswith "pt"
+            if node_id_mapping_file.endswith("pt"):
                 ori_node_id_mapping = th.load(node_id_mapping_file)
+            else:
+                # Homogeneous graph
+                # node id mapping file from dgl tools/distpartitioning/convert_partition.py.
+                ori_node_id_mapping = _load_dist_nid_map(node_id_mapping_file, ["_N"])["_N"]
```
thvasilo, line 431: Are we sure the node types are always going to be that?

thvasilo (approved):
LGTM, one small comment.

python/graphstorm/model/utils.py
classicsong: For distDGL, yes.

PR #654, opened 2023-11-16
[Bug Fix] GC for ExtFeatureWrapper

jalencato:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/python/graphstorm/gconstruct/utils.py
+++ b/python/graphstorm/gconstruct/utils.py
@@ -579,7 +579,8 @@ class ExtNumpyWrapper(ExtMemArrayWrapper):
     def cleanup(self):
         """ Clean up the array.
         """
-        self._arr.flush()
+        if self._arr is not None:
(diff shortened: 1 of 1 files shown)


python/graphstorm/gconstruct/utils.py
```diff
@@ -636,7 +636,8 @@ def cleanup(self):
         # Expected file structure:
         # merged_file file_feature1 file_feature2
         # rmtree will clean up all single feature files as ExtNumpyWrapper does not clean them up
-        self._arr.flush()
+        if self._arr is not None:
```
classicsong, line 639: I remember there are two places calling flush

jalencato: ExtNumpyWrapper called it as well. Let me check if this works with all our current test first. I will add it to the ExtNumpyWrapper as well. And add some test to it

classicsong (approved):

PR #659, opened 2023-11-18
[Bugfix] save_prediction_path should not use save_emb_path as its default value

classicsong:
Originally, if save_prediction_path is not set, it will use save_emb_path. The prediction results and emb results will be saved in the save path and the nid files of prediction results and embeddings will be mixed up.

Changing: Set the default value of save_prediction_path to None.

### DIFF
--- a/python/graphstorm/config/argument.py
+++ b/python/graphstorm/config/argument.py
@@ -272,7 +272,6 @@ class GSConfig:
         _ = self.eval_batch_size
         _ = self.eval_frequency
         _ = self.no_validation
-        _ = self.save_prediction_path
         _ = self.eval_etype
         if self.task_type is not None:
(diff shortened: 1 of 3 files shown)


python/graphstorm/config/argument.py
```diff
@@ -1564,8 +1566,9 @@ def save_prediction_path(self):
             return self._save_prediction_path
 
         # if save_prediction_path is not specified in inference
-        # use save_embed_path
-        return self.save_embed_path
+        logging.warning("save_prediction_path is not set. Prediction results" \
```
zheng-da, line 1569: i think we shouldn't print this warning message. we usually print warning messages for problems. but not setting prediction path is a normal thing to do.

Issue #541, opened 2023-10-10
[Bug] AUC_ROC and Pecision_Recall metric do not work in Binary Classification


zhjwy9343:
In Binary Classification case, use auc_roc and precision_recall as the evaluation metrics, there will an error raised by scklearn library. 
`ERROR:root:Failure found during evaluation of the auc metric returning -1: y should be a 1d array, got an array of shape (*****, 2) instead.`

Reason:
scklearn' auc, precision_recall function require 1d predictions in binary case, but GS return 2d prediction.

Suggested solutions:
Because scklearn also asks for 2d prediction for multiple classes, where the number of classes is larger than 2, GS need to check the prediction dimension.
- If pred.shape[1] == 2, take the 2nd column to form a 1d prediction;
- If pred.shape[1] > 2, no change of prediction.

PR #664, opened 2023-11-21
Add condition for binary classification

GentleZhu:
*Issue #, if available:*
https://github.com/awslabs/graphstorm/issues/541

*Description of changes:*
Add conditions on ROC_AUC and PRAUC calculation to check whether pred_proba only has two classes.

### DIFF
--- a/python/graphstorm/eval/eval_func.py
+++ b/python/graphstorm/eval/eval_func.py
@@ -195,13 +195,17 @@ def eval_roc_auc(logits,labels):
     # The roc_auc_score function computes the area under the receiver operating characteristic
     # (ROC) curve, which is also denoted by AUC or AUROC. The following returns the average AUC.
     rocauc_list = []
-    labels=labels_to_one_hot(labels, predicted_labels.shape[1])
-    for i in range(labels.shape[1]):
-        # AUC is only defined when there is at least one positive data.
-        if np.sum(labels[:, i] == 1) > 0 and np.sum(labels[:, i] == 0) > 0:
-            is_labeled = labels[:, i] == labels[:, i]
-            rocauc_list.append(roc_auc_score(labels[is_labeled, i],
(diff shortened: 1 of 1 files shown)


python/graphstorm/eval/eval_func.py
```diff
@@ -195,13 +195,17 @@ def eval_roc_auc(logits,labels):
     # The roc_auc_score function computes the area under the receiver operating characteristic
     # (ROC) curve, which is also denoted by AUC or AUROC. The following returns the average AUC.
     rocauc_list = []
-    labels=labels_to_one_hot(labels, predicted_labels.shape[1])
```
classicsong, line 198: Can you add some unit test for this function?

python/graphstorm/eval/eval_func.py
```diff
@@ -195,13 +195,17 @@ def eval_roc_auc(logits,labels):
...
-    for i in range(labels.shape[1]):
-        # AUC is only defined when there is at least one positive data.
-        if np.sum(labels[:, i] == 1) > 0 and np.sum(labels[:, i] == 0) > 0:
-            is_labeled = labels[:, i] == labels[:, i]
-            rocauc_list.append(roc_auc_score(labels[is_labeled, i],
-                                             predicted_labels[is_labeled, i]))
+    if predicted_labels.shape[1] == 2:
```
classicsong, line 198: Can you add comments about in which case `predicted_labels.shape[1] == 2` ?
Do you mean this is binary classification?

python/graphstorm/eval/eval_func.py
GentleZhu: Do we have any existing unit test for binary classification?

zhjwy9343:
In the original eval_func.py line 290, there is another use of auc_roc_score() function. Please also check if this will be impacted by binary classifcation.

PR #666, opened 2023-11-27
[BugFix] Avoid mistakenly printing keep alive failure error message in SageMaker

classicsong:
*Issue #, if available:*
SageMaker training/inference script may mistakenly print an error message about the failure of keep alive checking. This is due to then race between the keep_alive thread and the main process during cluster termination.

*Description of changes:*
Add a check to avoid sending keep alive message when the cluster has finished its job.


python/graphstorm/sagemaker/utils.py
```diff
@@ -115,7 +115,7 @@ def keep_alive(client_list, world_size, task_end):
 
     logging.info("keepalive thread exiting...")
 
-def terminate_workers(client_list, world_size, task_end):
+def terminate_workers(client_list, world_size):
     """ termiate all worker deamons.
```
thvasilo, line 119:
```suggestion
    """ terminate all worker daemons.
```

thvasilo (approved):
LGTM, minor spelling corrections.

PR #667, opened 2023-11-27
Implemented evaluator for link-prediction in retrieval setting

wangz10:
*Issue #, if available:*

*Description of changes:*  
- This PR enables `lp_infer.py` to evaluate a trained link-prediction model in retrieval setting, where all nodes of the target type are used as negative samples. To perform this evaluation:
  + need to pass `--eval-negative-sampler full` to `python3 -m graphstorm.run.gs_link_prediction --inference`
  + need to provide trained model (`--restore-model-path`) and saved GNN embeddings (`--restore-embed-path`)
- implementation details:
  + `GSgnnLinkPredictionRetrievalDataLoader`: a data loader yielding pos pairs only
  + `load_gsgnn_embeddings` util function to load GNN embeddings from disk
  + `LinkPredictDotDecoder.calc_retrieval_scores`: method to compute dot product scores with pos pairs and all embs as neg pairs


python/graphstorm/model/utils.py
```diff
@@ -872,6 +872,33 @@ def save_full_node_embeddings(g, save_embed_path,
 
     save_shuffled_node_embeddings(shuffled_embs, save_embed_path, save_embed_format)
 
+def load_gsgnn_embeddings(emb_path, g):
```
zheng-da, line 875: we didn't have a function to load GNN embeddings saved by GraphStorm? @classicsong

PR #675, opened 2023-12-05
Implemented hits@k metric for link-prediction

wangz10:
*Issue #, if available:*

*Description of changes:*
- Implemented hits@k metric for link-prediction
- Refactored `GSgnnMrrLPEvaluator` and `GSgnnPerEtypeMrrLPEvaluator` to work with both mrr and hits@k


python/graphstorm/config/argument.py
```diff
@@ -2076,6 +2076,14 @@ def eval_metric(self):
 
         return eval_metric
 
+    @property
+    def eval_k(self):
```
classicsong, line 2080: Is it possible to have a list of ks instead of just 1 k?

python/graphstorm/eval/evaluator.py
```diff
@@ -802,14 +804,16 @@ class GSgnnMrrLPEvaluator(GSgnnLPEvaluator):
...
-    def __init__(self, eval_frequency, data,
+    def __init__(self, eval_frequency, eval_metric, data,
                  num_negative_edges_eval, lp_decoder_type,
                  use_early_stop=False,
                  early_stop_burnin_rounds=0,
                  early_stop_rounds=3,
-                 early_stop_strategy=EARLY_STOP_AVERAGE_INCREASE_STRATEGY):
-        eval_metric = ["mrr"]
+                 early_stop_strategy=EARLY_STOP_AVERAGE_INCREASE_STRATEGY,
+                 k=100):
```
classicsong, line 816: It is better to define a hit@k evaluator.

python/graphstorm/config/argument.py
wangz10: I think most of the LP/retrieval datasets only select one specific k.

python/graphstorm/eval/evaluator.py
classicsong: We have two choices:
1) define a GSgnnRankingLPEvaluator class and both GSgnnMrrLPEvaluator and GSgnnHitKLPEvaluator will inherate from GSgnnRankingLPEvaluator.
2) define a GSgnnRankingLPEvaluator class which implements mrr() and hit(), and rename it to GSgnnMrrLPEvaluator and  GSgnnRankingLPEvaluator.

I would prefer the first one.

PR #676, opened 2023-12-05
[Bug fix] Fix ShageMaker S3 path bug

classicsong:
*Issue #, if available:*
`S3Downloader.download` use S3 path prefix match to find all matched file to download. This is different from folder match.

*Description of changes:*
Add "/" to the S3 path when downloading partitions

### DIFF
--- a/python/graphstorm/sagemaker/utils.py
+++ b/python/graphstorm/sagemaker/utils.py
@@ -274,7 +274,8 @@ def download_graph(graph_data_s3, graph_name, part_id, world_size,
         logging.info("Download graph from %s to %s",
                      os.path.join(graph_data_s3, graph_part),
                      graph_part_path)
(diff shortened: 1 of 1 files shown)


thvasilo (approved):
LGTM this should help avoid downloading all files under same prefix.

python/graphstorm/sagemaker/utils.py
```diff
@@ -272,9 +272,10 @@ def download_graph(graph_data_s3, graph_name, part_id, world_size,
             graph_path, sagemaker_session=sagemaker_session)
     try:
         logging.info("Download graph from %s to %s",
-                     os.path.join(graph_data_s3, graph_part),
+                     os.path.join(os.path.join(graph_data_s3, graph_part), ""),
```
zheng-da, line 275: can you provide comments why we have to generate the path like this?

PR #677, opened 2023-12-06
Add the support of using WholeGraph distributed embedding to store/update sparse_emb

chang-l:
Update on 12/11/2023:

Most added code is covered by unit tests, except the sparse optimizer step function in graphstorm. To compensate it and gain more confident for the integrated feature, I also added an independent unit test `test_wg_sparse_opt.py`, which has no dependency on graphstorm, to show the integrated wholegraph sparse optimizer can generate close enough embs vs. torch sparse adam optimizer.

Current code should be compatible now, ie., all unit tests should pass (except the fails due to https://github.com/awslabs/graphstorm/issues/685). This PR is ready for review.

Updated todos:
- ~Test/debug for `sparse_optimizer.step()` to show the updates are consistent between WholeGraph vs. DistDGL vs. PyTorch~
- Need to validate/benchmark an end2end test with learnable embeddings using WholeGraph
- Misc. questions/TODOs in the comments

---
Just like using WholeGraph distributed feature store for gathering, this PR tries to add the support of using WholeGraph to manage `sparse_embs` to unblock learnable embeddings with WholeGraph.

For now, it is still a draft PR.  The working test only covers save/load sparse embeddings using WholeGraph.

Todos:
- Test/debug for `sparse_optimizer.step()` to show the updates are consistent between WholeGraph vs. DistDGL vs. PyTorch
- Still need to work on adding end2end test with learnable embeddings using WholeGraph (any missing pieces?)
- Misc. questions/TODOs in the comments

@zheng-da @classicsong @isratnisa @nvcastet @TristonC


python/graphstorm/model/embed.py
```diff
@@ -213,57 +222,102 @@ def __init__(self,
...
-                                                               part_policy)
+                    if is_distributed() and is_wholegraph_sparse_emb():
+                        if get_rank() == 0:
+                            logging.debug(
+                                "Use WholeGraph to host additional sparse embeddings on node %s",
+                                ntype,
+                            )
+
+                        # Here, _sparse_embeds[ntype] holds WholeMemoryEmbeddingModule
+                        # Question(chang-l): cannot pass the embed name into WG embedding, would
```
classicsong, line 272: Can we create a WholeGraphEmbedding class under model module? Then we can store the embed name in its object.

python/graphstorm/model/embed.py
```diff
@@ -290,6 +344,8 @@ def forward(self, input_feats, input_nodes):
         assert isinstance(input_nodes, dict), 'The input node IDs should be in a dict.'
         embs = {}
         for ntype in input_nodes:
+            if isinstance(input_nodes[ntype], np.ndarray):
```
classicsong, line 347: Why input_nodes is an numpy ndarray?

python/graphstorm/model/embed.py
```diff
@@ -299,20 +355,38 @@ def forward(self, input_feats, input_nodes):
...
                 device = self.proj_matrix[ntype].device
-                if len(input_nodes[ntype]) == 0:
-                    dtype = self.sparse_embeds[ntype].weight.dtype
-                    embs[ntype] = th.zeros((0, self.sparse_embeds[ntype].embedding_dim),
-                                           device=device, dtype=dtype)
-                    continue
-                emb = self.sparse_embeds[ntype](input_nodes[ntype], device)
+                if is_wholegraph_embedding_module(self.sparse_embeds[ntype]):
+                    # Need all procs pass here due to nccl all2lallv in wholegraph
+                    node_emb = self.sparse_embeds[ntype](input_nodes[ntype].cuda())
```
classicsong, line 373: What is this? are you forcing every trainer to call wholegraph here?

python/graphstorm/model/embed.py
```diff
@@ -340,6 +414,75 @@ def require_cache_embed(self):
         """
         return self.cache_embed
 
+    def create_wholememory_optimizer(self, optimizer_type: str, param_dict: dict):
```
classicsong, line 417: We can create a new file called wholegraph.py

python/graphstorm/model/embed.py
```diff
@@ -340,6 +414,75 @@ def require_cache_embed(self):
...
+        param_dict: dict
+            parameters of the optimizer
+
+        Returns
+        -------
+        WholeMemoryOptimizer obj (pylibwholegraph.torch.WholeMemoryOptimizer)
+        """
+        import pylibwholegraph.torch as wgth
+
+        return wgth.create_wholememory_optimizer(optimizer_type, param_dict)
```
classicsong, line 437: WholeMemoryOptimizer is cool!

python/graphstorm/model/utils.py
```diff
@@ -186,27 +195,53 @@ def save_sparse_emb(model_path, sparse_emb, ntype):
...
     emb_path = os.path.join(model_path, ntype)
     os.makedirs(emb_path, exist_ok=True)
-    emb_file_path = os.path.join(emb_path, f'sparse_emb_{pad_file_index(rank)}.pt')
-    th.save(embs, emb_file_path)
+
+    if is_wholegraph_sparse_emb() and is_wholegraph_embedding_module(sparse_emb):
+        # Using WholeGraph will save sparse emb in npy format, we need np.load(mmap_mode)
+        # to load the file concurrently by all ranks, which is not supported by torch.load
+        emb_file_path = os.path.join(emb_path, f"sparse_emb_{pad_file_index(rank)}.npy")
+        (
```
classicsong, line 216:
```suggestion
        (local_tensor, _) \
            = sparse_emb.wm_embedding.get_embedding_tensor().get_local_tensor(host_view=True)
```

python/graphstorm/model/utils.py
```diff
@@ -1359,29 +1394,118 @@ def load_sparse_emb(target_sparse_emb, ntype_emb_path):
...
     world_size = get_world_size()
     num_files = len(os.listdir(ntype_emb_path))
-    num_embs = target_sparse_emb.num_embeddings
+    if is_wholegraph_sparse_emb() and is_wholegraph_embedding_module(target_sparse_emb):
+        num_embs = target_sparse_emb.wm_embedding.shape[0]
+    else:
+        num_embs = target_sparse_emb.num_embeddings
+
+    def _wholegraph_load_scatter(num_embs, file_id, num_files, file_path):
+        def _get_sparse_emb_lowerbound_range(file_size, rank, world_size):
```
classicsong, line 1403: I think tools/convert_feat_to_wholegraph.py uses a similar logic. Shall we unify the implementation and move this _get_sparse_emb_lowerbound_range to wholegraph.py as a helper function?

python/graphstorm/utils.py
```diff
@@ -91,11 +92,61 @@ def use_wholegraph(part_config):
...
     return USE_WHOLEGRAPH
 
+def use_wholegraph_sparse_emb():
+    """ Use wholegraph for feature fetching directly
+    """
+    global USE_WHOLEGRAPH_SPARSEEMB
+    try:
+        # Question(chang-l): Do we need to have an option in gs config file?
+        USE_WHOLEGRAPH_SPARSEEMB = (
+            "USE_WG_SPARSE_EMB" in os.environ and os.environ.get("USE_WG_SPARSE_EMB") != "0"
```
classicsong, line 102: For GraphStorm, we prefer to use parameters to control the execution logic.

python/graphstorm/model/embed.py
chang-l: It could be a ndarray, regarding an existing test: https://github.com/awslabs/graphstorm/blob/5e7cbb8b1e781fb90959f280c47bd84b4c429508/tests/unit-tests/test_embed.py#L67
It's fine if it's a DGL distTensor/PyT tensor (indexing array is a ndarray) but it wouldn't work for wholegraph gather, which requires to be a tensor obj... I can move this line under the condition `if is_wholegraph_embedding_module(..):` if you prefer.

python/graphstorm/model/embed.py
chang-l: Yes, in case `input_nodes[ntype]` could be an empty tensor and  skip this iteration (`continue` within the following lines); some of the training procs will not call nccl all2all, which is undesirable for wholegraph. Here, I move the wholegraph gather call, i.e., `self.sparse_embeds[ntype](input_nodes[ntype].cuda())` before `continue` to avoid nccl segfault.

python/graphstorm/model/embed.py
chang-l: Btw, your comment reminds me that there is a bug in WholeMemoryOptimizer that affect the accuracy when embed_dim is not multiple of 32 (https://github.com/rapidsai/wholegraph/pull/108).  I patch fixed it when I perform all the unit tests and e2e tests.

python/graphstorm/model/utils.py
chang-l: Agree.

chang-l: I simplified logic to consistent with tools/convert_feat_to_wholegraph.py.
The downside is: the saved sparse embs are in binary format, i.e., meta info, like shapes, dtype etc.  will not be available. It might not be an issue, as when we load it via `load_sparse_emb()`, the meta info should be available in `target_sparse_emb`, right?

python/graphstorm/model/embed.py
chang-l: Created a such class, `WholeGraphSparseEmbedding`, under wholegraph.py

chang-l:
@classicsong  Here is an example of yaml file change, w.r.t this PR, to add wholegraph support for learnable embeddings: https://github.com/awslabs/graphstorm/commit/44e4643c720ed0a74672cd35cda8cdfbc44be2cd

Note I only updated file [gsgnn_lp.py](https://github.com/awslabs/graphstorm/blob/main/python/graphstorm/run/gsgnn_lp/gsgnn_lp.py) for link pred tasks.

python/graphstorm/model/embed.py
```diff
@@ -201,69 +215,123 @@ def __init__(self,
...
 
         self.activation = activation
         self.cache_embed = cache_embed
 
-        if dgl.__version__ <= "1.1.2" and is_distributed() and get_backend() == "nccl":
+        if (
+            dgl.__version__ <= "1.1.2"
+            and is_distributed()
+            and get_backend() == "nccl"
+            and not self.use_wholegraph_sparse_emb
```
classicsong, line 236: Let's make use_wholegraph_sparse_emb a property of GSNodeEncoderInputLayer

python/graphstorm/model/embed.py
```diff
@@ -290,29 +358,46 @@ def forward(self, input_feats, input_nodes):
...
-                    dtype = self.sparse_embeds[ntype].weight.dtype
-                    embs[ntype] = th.zeros((0, self.sparse_embeds[ntype].embedding_dim),
-                                           device=device, dtype=dtype)
-                    continue
-                emb = self.sparse_embeds[ntype](input_nodes[ntype], device)
+                # If DistEmbedding supports 0-size input, we can remove this if statement.
+                if isinstance(self.sparse_embeds[ntype], WholeGraphSparseEmbedding):
+                    # Need all procs pass the following due to nccl all2lallv in wholegraph
+                    emb = self.sparse_embeds[ntype](input_nodes[ntype], device)
+                    if len(input_nodes[ntype]) == 0:
```
classicsong, line 386: Do we really need L386-L390? Will wholegraph sparse_embeds() return an empty tensor exactly same as L388?

python/graphstorm/model/utils.py
```diff
@@ -190,24 +196,36 @@ def save_sparse_emb(model_path, sparse_emb, ntype):
...
     os.makedirs(emb_path, exist_ok=True)
-    emb_file_path = os.path.join(emb_path, f'sparse_emb_{pad_file_index(rank)}.pt')
-    th.save(embs, emb_file_path)
+
+    if isinstance(sparse_emb, WholeGraphSparseEmbedding):
+        (local_tensor, _) = sparse_emb.weight.get_local_tensor(host_view=True)
+        # Using WholeGraph will save sparse emb in binary format (evenly distributed)
+        # Example: wg_sparse_emb_part_1_of_2, wg_sparse_emb_part_2_of_2
+        assert (
+            local_tensor.shape[0] == end - start
```
classicsong, line 212: Is _get_sparse_emb_range compatible with WholeGraph tensor split?

python/graphstorm/model/utils.py
```diff
@@ -1361,28 +1379,36 @@ def load_sparse_emb(target_sparse_emb, ntype_emb_path):
...
-            for idx in idxs:
-                # TODO: dgl.distributed.DistEmbedding should allow some basic tensor ops
-                target_sparse_emb._tensor[start+idx] = emb[idx]
+
+    if isinstance(target_sparse_emb, WholeGraphSparseEmbedding):
+        # Using WholeGraph will load sparse emb in binary format, let's assume
+        # the sparse emb is saved by WholeGraphSparseEmbedding.save_to_file(), i.e.,
+        # the meta info remains the same.
+        # Example: wg_sparse_emb_part_1_of_2, wg_sparse_emb_part_2_of_2
+        target_sparse_emb.load_from_file(ntype_emb_path, "wg_sparse_emb", num_files)
```
classicsong, line 1388: Do we allow save wholegraph sparse tensor into 8 parts with 8 GPUs but load them using only 4 GPUs?
If not, can we add some check here?

python/graphstorm/wholegraph/wholegraph.py
```diff
@@ -283,3 +288,241 @@ def load_wg_feat(part_config_path, num_parts, type_name, name):
...
+):
+    """Create a wholegraph sparse embedding module.
+
+    This is to use wholegraph distributed host/device memory to store sparse embs.
+    To enable trainable embeddings, the created wholegraph embedding has to be wrapped
+    in a WholeMemoryEmbeddingModule to attach gradients during each pass.
+
+    Parameters
+    ----------
+    nnodes : int
```
classicsong, line 332: num_nodes: int

python/graphstorm/wholegraph/wholegraph.py
```diff
@@ -283,3 +288,241 @@ def load_wg_feat(part_config_path, num_parts, type_name, name):
...
+    optimizer_type: str
+        optimizer types: [ "sgd" | "adam" | "adagrad" | "rmsprop" ]
+    param_dict: dict
+        parameters of the optimizer
+
+    Returns
+    -------
+    WholeMemoryOptimizer : WholeGraph native optimizer (wgth.WholeMemoryOptimizer)
+    """
+    if wgth is None:
```
classicsong, line 313: Shall we check WHOLEGRAPH_INIT instead of wgth?

python/graphstorm/model/embed.py
chang-l: You are right, wholegraph should return a empty tensor with valid meta info.

python/graphstorm/model/utils.py
chang-l: Yes, it should. Added an assertion here for sanity check.

python/graphstorm/wholegraph/wholegraph.py
chang-l: Sounds good.

python/graphstorm/model/utils.py
chang-l: As long as we use all procs to load all files, it should be fine, i.e., num_files does not need to be the same as num_training_procs.
For example, say we save WG sparse emb saved in 8 files; then later, we can load them with arbitrary procs as long as all wholegraph processes are participated.

python/graphstorm/model/embed.py
```diff
@@ -201,69 +215,123 @@ def __init__(self,
...
         # create weight embeddings for each node for each relation
         self.proj_matrix = nn.ParameterDict()
         self.input_projs = nn.ParameterDict()
-        embed_name = 'embed'
+        embed_name = "embed"
+        if self.use_wholegraph_sparse_emb:
+            # WG sparse optimizer has to be created at first like below
+            # This is because WG embedding depends on WG sparse optimizer to track/trace
+            # the gradients for embeddings.
+            self.wg_sparse_embs_optimizer = create_wholememory_optimizer("adam", {})
```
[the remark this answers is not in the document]
chang-l: I postpone the tensor materialization step until the initialization of the optimizer, to make it consistent with the existing logic..

python/graphstorm/model/gnn.py
```diff
@@ -729,7 +750,23 @@ def init_optimizer(self, lr, sparse_optimizer_lr, weight_decay, lm_lr=None):
...
-            emb_optimizer = dgl.distributed.optim.SparseAdam(sparse_params, lr=sparse_optimizer_lr)
+            if self.use_wholegraph_sparse_emb():
+                # To use wholegraph sparse optimizer, optimizer needs to be created
+                # before sparse embeddings. Within attach_wg_optimizer, we materialize
+                # the WG distributed tensor and then attach the optimizer.
+                emb_optimizer = create_wholememory_optimizer("adam", {})
+                for params in sparse_params:
+                    for param in params:
+                        assert isinstance(param, WholeGraphDistTensor) and param.use_wg_optimizer, \
+                            "Please create params (WG tensor) with use_wg_optimizer=True."
```
classicsong, line 761:
```suggestion
                            "Please create params (WholeGraph tensor) with use_wg_optimizer=True."
```

python/graphstorm/model/utils.py
classicsong: Another question, does it require each machine participating in the training has all the wholegraph files under ntype_emb_path?

python/graphstorm/run/gsgnn_lp/gsgnn_lp.py
```diff
@@ -91,8 +96,9 @@ def main(config_args):
     config = GSConfig(config_args)
     config.verify_arguments(True)
 
+    use_wg_feats = use_wholegraph(config.part_config)
     gs.initialize(ip_config=config.ip_config, backend=config.backend,
-                  use_wholegraph=use_wholegraph(config.part_config))
+                  use_wholegraph=config.use_wholegraph_sparse_emb or use_wg_feats)
```
classicsong, line 101: Shall we also update gsgnn_np.py, gsgnn_ep.py and gsgnn_node_emb.py?

python/graphstorm/wholegraph/wholegraph.py
```diff
@@ -283,3 +292,336 @@ def load_wg_feat(part_config_path, num_parts, type_name, name):
...
+    ):
+        self._nnodes = shape[0]
+        self._embedding_dim = shape[1]
+        self._name = name
+        self._dtype = dtype
+        self._location = location
+        self._use_wg_optimizer = use_wg_optimizer
+        # Need the pylibwholegraph be at least 23.12.00 to support _tensor.scatter API.
+        assert pylibwholegraph.__version__ >= "23.12.00", \
+            "Please upgrade to the latest version of WholeGraph."
```
classicsong, line 393:
```suggestion
            "Please upgrade to WholeGraph 23.12.00 or higher."
```

python/graphstorm/wholegraph/wholegraph.py
```diff
@@ -283,3 +292,336 @@ def load_wg_feat(part_config_path, num_parts, type_name, name):
...
+        -------
+        None
+
+        """
+        assert self._use_wg_optimizer, \
+            "Please create WholeGraphDistTensor tensor with use_wg_optimizer=True."
+        if self.optimizer == wg_optimizer and self._module is not None:
+            # no-op if the optimizer is the same
+            return
+        # When attach a new optimizer, we have to purge the old _tensor/_module/_optimizer.
```
classicsong, line 423: If self.optimizer is not None, Can we just raise an error here?
Ideally, we do not want to initialize the optimizer twice.

python/graphstorm/wholegraph/wholegraph.py
```diff
@@ -437,17 +450,19 @@ def save_to_file(
         -------
         None
         """
+        assert self._tensor is not None, "Please create WholeGraph tensor first."
```
classicsong, line 453:
```suggestion
        assert self._tensor is not None, "Please create WholeGraph tensor first by either create WholeGraphDistTensor with use_wg_optimizer=False or by calling attach_wg_optimizer after WholeGraphDistTensor is initialized."
```
As the behavior of WholeGraphDistTensor is different from other DistTensors, we need to be more specific in the error message.

python/graphstorm/wholegraph/wholegraph.py
```diff
@@ -457,16 +472,88 @@ def load_from_file(
...
-
+        wg_optimizer : WholeMemoryOptimizer or None
+            The WholeGraph sparse optimizer to be attached to the WholeGraph embedding.
         Returns
         -------
         None
         """
+
+        if self._tensor is None and self._use_wg_optimizer:
+            self.attach_wg_optimizer(wg_optimizer)
```
classicsong, line 483: Shall we assert wg_optimizer is not None here?

python/graphstorm/model/utils.py
classicsong: In WhoelGraphDistTensor.load_from_file, wg_optimizer is required if we are loading a trainable sparse embedding. But his is not handled here. Shall we pass in a wg_optimizer here or we will do this later when initializing the optimizer.

python/graphstorm/model/embed.py
```diff
@@ -290,29 +354,46 @@ def forward(self, input_feats, input_nodes):
...
                 emb = input_feats[ntype].float() @ self.input_projs[ntype]
                 if self.use_node_embeddings:
                     assert ntype in self.sparse_embeds, \
-                            f"We need sparse embedding for node type {ntype}"
-                    node_emb = self.sparse_embeds[ntype](input_nodes[ntype], emb.device)
-                    concat_emb=th.cat((emb, node_emb),dim=1)
+                        f"We need sparse embedding for node type {ntype}"
+                    # emb.device: target device to put the gathered results
+                    if self._use_wholegraph_sparse_emb:
+                        node_emb = self.sparse_embeds[ntype].module(input_nodes[ntype].cuda())
```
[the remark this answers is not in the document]
chang-l: Right, the output tensor will be on the current cuda device. 

>> Shall we use to(emb.device) instead of cuda()? Can cuda move the tensor to the right GPU?
As long as we called torch.cuda.set_device, `cuda()` indicates the current cuda device. In fact, this is a constraint of WholeGraph that it requires the input indexes can only located at the current cuda device (here, `module()` function is very similar to wholegraph `gather()` call).

python/graphstorm/model/embed.py
```diff
@@ -290,29 +354,46 @@ def forward(self, input_feats, input_nodes):
...
-                if len(input_nodes[ntype]) == 0:
-                    dtype = self.sparse_embeds[ntype].weight.dtype
-                    embs[ntype] = th.zeros((0, self.sparse_embeds[ntype].embedding_dim),
-                                           device=device, dtype=dtype)
-                    continue
-                emb = self.sparse_embeds[ntype](input_nodes[ntype], device)
+                # If DistEmbedding supports 0-size input, we can remove this if statement.
+                if isinstance(self.sparse_embeds[ntype], WholeGraphDistTensor):
+                    # Need all procs pass the following due to nccl all2lallv in wholegraph
+                    emb = self.sparse_embeds[ntype].module(input_nodes[ntype].cuda())
```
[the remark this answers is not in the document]
chang-l: Ditto (input indexes and output tensor have to be on the current cuda device)

python/graphstorm/model/utils.py
chang-l: >Another question, does it require each machine participating in the training has all the wholegraph files under ntype_emb_path?

Unfortunately, for now, yes.

chang-l: > wg_optimizer is required if we are loading a trainable sparse embedding.

I see. When loading a trainable embeddings, do we assume `wg_sparse_optimizer` is already available at: target_sparse_emb._optimizer? i.e., would `target_sparse_emb` be attached a optimizer via `init_optimizer()` function?
If so (target_sparse_emb._optimizer is not None), then I think we don't need to pass an additional optimizer to the function.

python/graphstorm/wholegraph/wholegraph.py
chang-l: Good point. I am thinking of two scenarios:
(1) user specifies wg_optimizer when calling load_from_file:
    In this sense, we need to replace the existing optimizer/tensor/module by a new one before reading from files. So we need to call attach-wg_optimizer again
(2) user does not specify wg_optimizer
    In this sense, by default, we assume user has self.optimizer defined (if use_wg_optimizer is on) and all we need to do is update the _tensor with the one loaded from files.

python/graphstorm/run/gsgnn_lp/gsgnn_lp.py
chang-l: I updated gsgnn_np.py, gsgnn_ep.py and gsgnn_node_emb.py. Pls let me know if there are others.

python/graphstorm/model/utils.py
classicsong: By default optimizer is initialized before restoring model params.

python/graphstorm/wholegraph/wholegraph.py
classicsong: Can you put the description of these two scenarios in the function docstring?
Actually there are three scenarios:
1) learnable embedding with wg_optimizer
2) learnable embedding with wg_optimizer = None
3) regular embedding.

python/graphstorm/wholegraph/wholegraph.py
```diff
@@ -457,16 +477,100 @@ def load_from_file(
...
+        torch.Tensor
+            The requested node embeddings.
+        """
+        assert self._tensor is not None, "Please create WholeGraph tensor first."
+        idx = idx.cuda()
+        output_tensor = self._tensor.gather(idx)  # output_tensor is on cuda by default
+        return output_tensor
+
+    def get_local_tensor(self):
+        """ Get the local tensor and offset of the embedding tensor. """
```
classicsong, line 547: What is the type of local_tensor. Can you add a docstring for Return values?

tests/unit-tests/test_wg_sparse_opt.py
```diff
@@ -18,129 +18,296 @@
...
+        "DGL_NUM_CLIENT",
+        "DGL_DIST_MAX_TRY_TIMES",
+        "DGL_DIST_DEBUG",
+    ]:
+        if key in os.environ:
+            os.environ.pop(key)
+
+def _initialize(proc_id, nprocs, use_wholegraph=True):
     backend = "nccl"
+    #from dgl.distributed import role
```
classicsong, line 96: remove commented code.

python/graphstorm/model/embed.py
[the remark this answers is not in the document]
chang-l: Right, it's already in GPU. But I am not sure if `emb.device` must be current cuda device, or could it be `cpu`?

python/graphstorm/wholegraph/wholegraph.py
chang-l: Added docstring for Returns

python/graphstorm/model/embed.py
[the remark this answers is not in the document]
classicsong: I c. it is in GPU.

classicsong (approved):

Issue #394, opened 2023-08-28
Support hard negative for link prediction tasks.


classicsong:
In link prediction tasks, randomly sampling nodes as negative nodes for positive edges creates 'easy negative' examples in most of time. It is required to support hard negative sampling so that a user can provide negative nodes under a specific edge type for a specific node.

PR #678, opened 2023-12-08
Add edge hard negative support.

classicsong:
*Issue #, if available:*
#394 

*Description of changes:*

### DIFF
--- a/python/graphstorm/config/argument.py
+++ b/python/graphstorm/config/argument.py
@@ -1875,25 +1875,26 @@ class GSConfig:
         return None
 
     @property
-    def train_hard_edge_dstnode_negative(self):
-        """ The list of canonical etypes that have hard negative sets
+    def train_etypes_negative_dstnode(self):
+        """ The list of canonical etypes that have hard negative edges
+        constructed by corrupting destination nodes.
 
             The format of the arguement should be:
-            train_hard_edge_dstnode_negative:
+            train_etypes_negative_dstnode:
               - src_type,rel_type0,dst_type:negative_nid_field
               - src_type,rel_type1,dst_type:negative_nid_field
             Each edge type can have different fields storing the hard negatives.
 
             or
-            train_hard_edge_dstnode_negative:
+            train_etypes_negative_dstnode:
               - negative_nid_field
             All the edge types use the same filed storing the hard negatives.
         """
         # pylint: disable=no-member
-        if hasattr(self, "_train_hard_edge_dstnode_negative"):
+        if hasattr(self, "_train_etypes_negative_dstnode"):
             assert self.task_type == BUILTIN_TASK_LINK_PREDICTION, \
                 "Hard negative only works with link prediction"
-            hard_negatives = self._train_hard_edge_dstnode_negative
+            hard_negatives = self._train_etypes_negative_dstnode
             if len(hard_negatives) == 1 and \
                 ":" not in hard_negatives[0]:
                 # global feat_name
@@ -1903,7 +1904,14 @@ class GSConfig:
             hard_negative_dict = {}
             for hard_negative in hard_negatives:
                 negative_info = hard_negative.split(":")
+                assert len(negative_info) == 2, \
+                    "negative dstnode information must be provided in format of " \
+                    f"src,relation,dst:feature_name, but get {hard_negative}"
+
                 etype = tuple(negative_info[0].split(","))
+                assert len(etype) == 3, \
+                    f"Edge type must in format of (src,relation,dst), but get {etype}"
+
                 assert etype not in hard_negative_dict, \
                     f"You already specify the fixed negative of {etype} " \
                     f"as {hard_negative_dict[etype]}"
@@ -1915,25 +1923,25 @@ class GSConfig:
         return None
 
     @property
-    def num_hard_negatives(self):
+    def num_train_hard_negatives(self):
         """ Number of hard negatives per edge type
 
             The format of the arguement should be:
-            num_hard_negatives:
+            num_train_hard_negatives:
               - src_type,rel_type0,dst_type:num_negatives
               - src_type,rel_type1,dst_type:num_negatives
             Each edge type can have different number of hard negatives.
 
             or
-            num_hard_negatives:
+            num_train_hard_negatives:
               - num_negatives
             All the edge types use the same number of hard negatives.
         """
         # pylint: disable=no-member
-        if hasattr(self, "_num_hard_negatives"):
+        if hasattr(self, "_num_train_hard_negatives"):
             assert self.task_type == BUILTIN_TASK_LINK_PREDICTION, \
                 "Hard negative only works with link prediction"
-            num_negatives = self._num_hard_negatives
+            num_negatives = self._num_train_hard_negatives
             if len(num_negatives) == 1 and \
                 ":" not in num_negatives[0]:
                 # global feat_name
@@ -1954,25 +1962,26 @@ class GSConfig:
         return None
 
     @property
-    def eval_fixed_edge_dstnode_negative(self):
-        """ The list of canonical etypes that have predefined negative sets
+    def eval_etypes_negative_dstnode(self):
+        """ The list of canonical etypes that have predefined negative edges
+        constructed by corrupting destination nodes.
 
             The format of the arguement should be:
-            eval_fixed_edge_dstnode_negative:
+            eval_etypes_negative_dstnode:
               - src_type,rel_type0,dst_type:negative_nid_field
               - src_type,rel_type1,dst_type:negative_nid_field
             Each edge type can have different fields storing the fixed negatives.
 
             or
-            eval_fixed_edge_dstnode_negative:
+            eval_etypes_negative_dstnode:
               - negative_nid_field
             All the edge types use the same filed storing the fixed negatives.
         """
         # pylint: disable=no-member
-        if hasattr(self, "_eval_fixed_edge_dstnode_negative"):
+        if hasattr(self, "_eval_etypes_negative_dstnode"):
             assert self.task_type == BUILTIN_TASK_LINK_PREDICTION, \
                 "Fixed negative only works with link prediction"
-            fixed_negatives = self._eval_fixed_edge_dstnode_negative
+            fixed_negatives = self._eval_etypes_negative_dstnode
             if len(fixed_negatives) == 1 and \
                 ":" not in fixed_negatives[0]:
                 # global feat_name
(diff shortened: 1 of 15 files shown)


python/graphstorm/config/argument.py
```diff
@@ -1874,9 +1874,128 @@ def lp_edge_weight_for_loss(self):
 
         return None
 
+    @property
+    def train_hard_edge_dstnode_negative(self):
```
zheng-da, line 1878:
```suggestion
    def train_hard_negative_edges_dstnode(self):
```

zheng-da: I think this is used for custom negative edges instead of only for hard negative edges.

python/graphstorm/config/argument.py
```diff
@@ -1874,9 +1874,128 @@ def lp_edge_weight_for_loss(self):
...
+        """ The list of canonical etypes that have hard negative sets
```
zheng-da, line 1879:
```suggestion
        """ The list of canonical etypes that have hard negative edges constructed by corrupting destination nodes.
```

python/graphstorm/config/argument.py
```diff
@@ -1874,9 +1874,128 @@ def lp_edge_weight_for_loss(self):
...
+            hard_negatives = self._train_hard_edge_dstnode_negative
+            if len(hard_negatives) == 1 and \
+                ":" not in hard_negatives[0]:
+                # global feat_name
+                return hard_negatives[0]
+
+            # per edge type feature
+            hard_negative_dict = {}
+            for hard_negative in hard_negatives:
+                negative_info = hard_negative.split(":")
```
zheng-da, line 1905: do we need to verify the correctness of the input format?

python/graphstorm/config/argument.py
```diff
@@ -1874,9 +1874,128 @@ def lp_edge_weight_for_loss(self):
...
+                    f"as {hard_negative_dict[etype]}"
+
+                hard_negative_dict[etype] = negative_info[1]
+            return hard_negative_dict
+
+        # By default fixed negative is not used
+        return None
+
+    @property
+    def num_hard_negatives(self):
```
zheng-da, line 1918: why this one doesn't have train in the field name and the one above has train in the field name?

zheng-da: the number of negative edges for training and evaluation has to be the same?

python/graphstorm/config/argument.py
```diff
@@ -1874,9 +1874,128 @@ def lp_edge_weight_for_loss(self):
...
+                    f"You already specify the fixed negative of {etype} " \
+                    f"as {num_hard_negative_dict[etype]}"
+
+                num_hard_negative_dict[etype] = int(negative_info[1])
+            return num_hard_negative_dict
+
+        return None
+
+    @property
+    def eval_fixed_edge_dstnode_negative(self):
```
zheng-da, line 1957: similar comments on the naming as the one above

python/graphstorm/dataloading/sampler.py
```diff
@@ -70,6 +72,190 @@ def _generate(self, g, eids, canonical_etype):
         dst = F.randint(shape, dtype, ctx, 0, self._local_neg_nids[vtype].shape[0])
         return src, self._local_neg_nids[vtype][dst]
 
+class GSHardEdgeDstNegative(object):
```
zheng-da, line 75: i think we should name this as hard negative sampler. it should be custom negative edge sampler.

python/graphstorm/dataloading/sampler.py
```diff
@@ -70,6 +72,190 @@ def _generate(self, g, eids, canonical_etype):
...
+        # value in `hard_negatives` will be integers representing the node ids.
+        # For others, they will be -1s meaning there are missing fixed negatives.
+        if th.sum(hard_negatives == -1) == 0:
+            # Fast track, there is no -1 in hard_negatives
+            max_num_hard_neg = hard_negatives.shape[1]
+            neg_idx = th.randperm(max_num_hard_neg)
+            # shuffle the hard negatives
+            hard_negatives = hard_negatives[:,neg_idx]
+
+            print(f"{required_num_hard_neg} {max_num_hard_neg} {self._k}")
```
zheng-da, line 130: remove this

python/graphstorm/dataloading/sampler.py
```diff
@@ -70,6 +72,190 @@ def _generate(self, g, eids, canonical_etype):
...
+                    # There is not enough hard negative to fill required_num_hard_neg
+                    num_hard_neg = max_num_hard_neg
+
+                # There is not enough negatives
+                src, neg = self._negative_sampler._generate(g, eids, canonical_etype)
+                # replace random negatives with fixed negatives
+                neg = neg.reshape(-1, self._k)
+                neg[:,:num_hard_neg] = hard_negatives[:,:num_hard_neg]
+                return src, neg.reshape((-1,))
+        else:
```
zheng-da, line 153: does this work for evaluation? if we want to use fixed negative edges for evaluation, we shouldn't allow generating random negative edges. i think accepting various numbers of hard negative edges makes the code complex.

python/graphstorm/dataloading/sampler.py
```diff
@@ -70,6 +72,190 @@ def _generate(self, g, eids, canonical_etype):
...
+
+        Returns
+        -------
+        tuple[Tensor, Tensor, Tensor, Tensor] or
+        dict[etype, tuple(Tensor, Tensor Tensor, Tensor)
+            The returned [positive source, negative source,
+            postive destination, negatve destination]
+            tuples as pos-neg examples.
+        """
+        def _gen_neg_pair(pos_pair, canonical_etype):
```
zheng-da, line 197: it's better to move this function out. it's complex enough.

python/graphstorm/dataloading/sampler.py
```diff
@@ -70,6 +72,190 @@ def _generate(self, g, eids, canonical_etype):
...
+    """
+    def __init__(self, k, dst_negative_field, negative_sampler, num_hard_negs=None):
+        assert is_wholegraph() is False, \
+                "Hard negative is not supported for WholeGraph."
+        self._dst_negative_field = dst_negative_field
+        self._k = k
+        self._negative_sampler = negative_sampler
+        self._num_hard_negs = num_hard_negs
+
+    def _generate(self, g, eids, canonical_etype):
```
zheng-da, line 99: where is this function called?

python/graphstorm/config/argument.py
classicsong: Rename it to num_train_hard_negatives

python/graphstorm/config/argument.py
classicsong: Changed to train_etypes_negative_dstnode

python/graphstorm/config/argument.py
classicsong: Add some.

python/graphstorm/dataloading/sampler.py
classicsong: Evaluation is defined in gen_neg_pairs.

python/graphstorm/dataloading/sampler.py
classicsong: I will implemented a GSFixedEdgeDstNegativeSampler class

python/graphstorm/dataloading/sampler.py
classicsong: This is called by DGL.

python/graphstorm/dataloading/sampler.py
classicsong: I created a new class: GSFixedEdgeDstNegativeSampler

python/graphstorm/config/argument.py
```diff
@@ -1874,9 +1874,143 @@ def lp_edge_weight_for_loss(self):
...
+            num_negatives = self._num_train_hard_negatives
+            if len(num_negatives) == 1 and \
+                ":" not in num_negatives[0]:
+                # global feat_name
+                return int(num_negatives[0])
+
+            # per edge type feature
+            num_hard_negative_dict = {}
+            for num_negative in num_negatives:
+                negative_info = num_negative.split(":")
```
zheng-da, line 1953: you should add similar checks as above.

python/graphstorm/config/argument.py
```diff
@@ -1874,9 +1874,143 @@ def lp_edge_weight_for_loss(self):
...
+
+                etype = tuple(negative_info[0].split(","))
+                assert len(etype) == 3, \
+                    f"Edge type must in format of (src,relation,dst), but get {etype}"
+                assert etype not in fixed_negative_dict, \
+                    f"You already specify the fixed negative of {etype} " \
+                    f"as {fixed_negative_dict[etype]}"
+
+                fixed_negative_dict[etype] = negative_info[1]
+            return fixed_negative_dict
```
zheng-da, line 2006: it seems the code is almost the same as `train_etypes_negative_dstnode`. should we have one function for both arguments?

python/graphstorm/dataloading/sampler.py
```diff
@@ -70,6 +72,182 @@ def _generate(self, g, eids, canonical_etype):
...
+        ----------
+        dst_negative_field: str or dict of str
+            The field storing the hard negatives.
+    """
+    def __init__(self, dst_negative_field):
+        assert is_wholegraph() is False, \
+                "Hard negative is not supported for WholeGraph."
+        self._dst_negative_field = dst_negative_field
+
+    def gen_neg_pairs(self, g, pos_pairs):
```
zheng-da, line 191: who call this method?

python/graphstorm/dataloading/sampler.py
zheng-da: i think you should have some docstring to explain how is this method used.

python/graphstorm/dataloading/sampler.py
zheng-da: is this called by DGL as well?

zheng-da: why do you need to provide `gen_neg_pairs` here and provide `_generate` in the class above?

classicsong: This is called by DGL.
gen_neg_pairs() is used by test dataloader and _generate is used by training dataloader.

python/graphstorm/dataloading/sampler.py
classicsong: Added

python/graphstorm/dataloading/sampler.py
classicsong: Added some doc string

python/graphstorm/config/argument.py
classicsong: We can

zheng-da (approved):

PR #681, opened 2023-12-11
Update builtin train/inference entries to accept unknown arguments

classicsong:
*Issue #, if available:*
As we chain train/inference with node id remapping in the train and inference pipeline, both train/inference entries and remap entry should be more robust to unknown arguments.

Related issue: #674 

*Description of changes:*
Update builtin train/inference entries to accept unknown arguments.


zhjwy9343 (approved):
LGTM

PR #694, opened 2023-12-22
Fix a bug in gconstruct when saving intermediate data on disks

zheng-da:
*Description of changes:*
Numpy has a bug when storing a list of objects in numpy memmap array. We should avoid saving such a list to a memmap array.

### DIFF
--- a/python/graphstorm/gconstruct/utils.py
+++ b/python/graphstorm/gconstruct/utils.py
@@ -42,7 +42,10 @@ def _to_ext_memory(name, data, path):
     if isinstance(data, np.ndarray):
         assert name is not None
         path = os.path.join(path, f"{name}.npy")
-        if len(data) > 0:
+        # We only save a data matrix to the disk.
+        # This avoids the problem of saving an array of objects to disks.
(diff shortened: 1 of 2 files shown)


python/graphstorm/gconstruct/utils.py
```diff
@@ -38,11 +38,19 @@
...
+
 def _to_ext_memory(name, data, path):
     if isinstance(data, np.ndarray):
         assert name is not None
         path = os.path.join(path, f"{name}.npy")
-        if len(data) > 0:
+        # We only save a data array with numeric or boolean values to the disk.
+        # This avoids the problem of saving an array of objects to disks.
+        # Note: There is a bug in Numpy when saving an array of objects to disks.
+        if len(data) > 0 and _is_numeric(data):
```
classicsong, line 53: Why not check whether data has dtype numpy.object?

classicsong (approved):

python/graphstorm/gconstruct/utils.py
zheng-da: there are many other numpy dtypes, such as str, that may not work well with external memory.

PR #695, opened 2023-12-26
Fix a bug in loading cached LM embeddings.

zheng-da:
*Description of changes:*
The current code cannot load cached LM embeddings correctly.

### DIFF
--- a/python/graphstorm/model/lm_embed.py
+++ b/python/graphstorm/model/lm_embed.py
@@ -288,7 +288,9 @@ class LMCache:
             # correctly, we should have loaded the cached embeddings of the LM models.
             # We can save the hash code of all LM models.
             self._lm_hash = self._lm_models.get_all_lm_hashes()
-        else:
+        elif len(self._lm_emb_cache) > 0:
+            # In this case, some of the cached LM embeddings have been loaded.
+            # We need to clear up the cache.
             if get_rank() == 0:
                 logging.warning("Fail to load all embeddings from the cache.")
                 logging.warning("Not use the cached data.")
--- a/python/graphstorm/model/lm_embed.py
+++ b/python/graphstorm/model/lm_embed.py
@@ -267,9 +267,10 @@ class LMCache:
         return model_name + "-" + model_hash[0:10]
 
     def _load_embeddings(self):
-        """ Load LM embeddings from files.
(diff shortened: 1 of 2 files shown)


python/graphstorm/model/lm_embed.py
```diff
@@ -388,9 +404,12 @@ def update_cache(self, lm_infer_batch_size, use_fp16=True):
                 logging.info('Computing bert embedding on node %s takes %.3f seconds',
                              ntype, time.time() - start)
             lm_model.train()
+        # store/update the lm model hash used for the cache
+        self._lm_hash = self._lm_models.get_all_lm_hashes()
```
classicsong, line 408: Is this the cause of the bug?

python/graphstorm/model/lm_embed.py
```diff
@@ -327,25 +340,28 @@ def update_cache(self, lm_infer_batch_size, use_fp16=True):
...
+        # If all cached embeddings has been loaded to memory:
         if np.all([ntype in self._lm_emb_cache for ntype in self._lm_models.ntypes]):
-            # check if lm is updated, if lm is updated, clear the current cache
+            # check if lm is updated. if it is updated, clear the current cache
             if self._lm_hash != self._lm_models.get_all_lm_hashes():
                 self._clear_cache()
             else:
                 # if lm models was not updated, don't compute the embeddings again
-                return
+                return False
```
classicsong, line 361: Who is consuming this return value?

python/graphstorm/model/lm_embed.py
zheng-da: not really. This statement runs when we need to compute BERT embeddings from scratch. We need to save the hash number of the current BERT models.

What we need to make sure is that when the BERT embeddings are loaded from files, we need to assign the right hash number for the loaded embeddings.

python/graphstorm/model/lm_embed.py
zheng-da: i think this is mainly used for testing right now. otherwise, it's hard to tell if the embeddings are computed from scratch or are loaded from files.

classicsong: OK

classicsong (approved):

PR #696, opened 2023-12-27
Fix a bug in inference code

zheng-da:
*Description of changes:*
This PR reorganizes the code of performing inference of GNN and LM models.
Specifically, we split the nodes for inference based on locality. In this case, the embeddings are saved to local partitions via shared memory and we only need to run barrier before returning fron the inference function.
However, when computing LM embeddings, we split the nodes evenly to ensure all processes take roughly the same amount of time to compute LM embeddings. Otherwise, we will see a timeout in barrier in some processes. Because now the nodes are split evenly, we need to write data to remote memory. Before returning from the inference function, we need to call flush_data to ensure all data written to distributed memory can be read correctly.


python/graphstorm/model/gnn_encoder_base.py
```diff
@@ -337,7 +343,13 @@ def dist_inference_one_layer(layer_id, g, dataloader, target_ntypes, layer, get_
...
                 assert k in y, "All mini-batch outputs should have the same tensor names."
                 y[k][output_nodes[k]] = h[k].cpu()
-    flush_data()
+    # The nodes are split in such a way that all processes only need to compute
+    # the embeddings of the nodes in the local partition. Therefore, a barrier
+    # is enough to ensure that all data have been written to memory for distributed
+    # read after this function is returned.
+    # Note: there is a risk here. If the nodes for inference on each partition
+    # are very skewed, some of the processes may timeout in the barrier.
+    barrier()
```
classicsong, line 352: You assume the data are split with `force_even=False`.
If a custom inference code invoke this API with force_even=True will potentially trigger a bug. Are we going to force users to use `force_even=False`? Is there anyway to check it?

zheng-da: this is a problem. we don't have a way to effectively check if the data is split evenly or not.

classicsong (approved):

PR #699, opened 2024-01-09
[Enhancement] change the input argument of GSTaskTrackerAbc to be an integer

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
- This PR changes the input argument of `GSTaskTrackerAbc` from `GSConfig` object into an integer because the `GSTaskTrackerAbc` only needs an integer to set the `log_report_frequency` attribute. 
- Using the `GSConfig` object will prevent users from using task tracker to monitor running process because creating a GSConfig is NOT publicly open, and is very complex.
- Decouple the `GSTaskTracker` from using `GSConfig` could help users to construct task trackers and use them in the GraphStorm programming APIs.

### DIFF
--- a/examples/customized_models/HGT/hgt_nc.py
+++ b/examples/customized_models/HGT/hgt_nc.py
@@ -335,7 +335,7 @@ def main(args):
                                   config.early_stop_strategy)
     trainer.setup_evaluator(evaluator)
     # Optional: set up a task tracker to show the progress of training.
-    tracker = GSSageMakerTaskTracker(log_log_report_frequency=config.eval_frequency)
+    tracker = GSSageMakerTaskTracker(log_report_frequency=config.eval_frequency)
     trainer.setup_task_tracker(tracker)
 
     # Start the training process.
--- a/examples/customized_models/HGT/hgt_nc.py
+++ b/examples/customized_models/HGT/hgt_nc.py
@@ -335,7 +335,7 @@ def main(args):
                                   config.early_stop_strategy)
     trainer.setup_evaluator(evaluator)
     # Optional: set up a task tracker to show the progress of training.
-    tracker = GSSageMakerTaskTracker(log_report_frequency=config.eval_frequency)
+    tracker = GSSageMakerTaskTracker(config.eval_frequency)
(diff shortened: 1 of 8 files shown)


examples/customized_models/HGT/hgt_nc.py
```diff
@@ -335,7 +335,7 @@ def main(args):
                                   config.early_stop_strategy)
     trainer.setup_evaluator(evaluator)
     # Optional: set up a task tracker to show the progress of training.
-    tracker = GSSageMakerTaskTracker(config)
+    tracker = GSSageMakerTaskTracker(log_log_report_frequency=config.eval_frequency)
```
classicsong, line 338:
```suggestion
    tracker = GSSageMakerTaskTracker(log_report_frequency=config.eval_frequency)
```

classicsong (commented):
Can you provide why you make this change in the PR description?

zhjwy9343:
> Can you provide why you make this change in the PR description?

Sure. Update the descriptions and give the reasons of this change.

python/graphstorm/gsf.py
```diff
@@ -656,4 +656,4 @@ def check_homo(g):
 
 def create_builtin_task_tracker(config):
     tracker_class = get_task_tracker_class(config.task_tracker)
-    return tracker_class(config)
+    return tracker_class(config.log_report_frequency)
```
classicsong, line 659: We are using config.log_report_frequency here but in other places and in the own-models.rst, we recommend users to use config.eval_frequency. Any specific reasons?

classicsong: Shall we use eval_frequency here and deprecate log_report_frequency

zhjwy9343: Yes, agree to deprecate the log_report_frequency. This argument is confused. We's better to reduce the confusion.

I will remove the `log_report_frequency`, replacing it with eval_frequency.

zhjwy9343: changed. And also remove the `log_report_frequency` from docs. So later we can deprecate this argument.

classicsong (approved):

PR #700, opened 2024-01-10
[GSProcessing] BERT Tokenizer

jalencato:
*Issue #, if available:*

*Description of changes:*

------------------------------------------------------
Beside implementing the BERT Tokenizer feature
- Add etype_featsize dict in graph loader to fix a bug
- Add dependency for setuptools that is necessary for emr-serverless run
- Reformat the _process_node_feature and _process_edge feature part to process tokenize feature and also allow enough backward compatibilities. 

Result:
- Constructing full mag dataset with tokenize feature for 111 minutes. Refer to the example here: https://github.com/awslabs/graphstorm/blob/main/examples/mag/mag_v0.2.json


docs/source/gs-processing/developer/input-configuration.rst
```diff
@@ -447,6 +447,15 @@ arguments.
         will be considered as an array. For Parquet files, if the input type is ArrayType(StringType()), then the
         separator is ignored; if it is StringType(), it will apply same logic as in CSV.
 
+-  ``huggingface``
+
+   -  Transforms a text feature column to tokens or embeddings with different Hugging Face models, enabling nuanced understanding and processing of natural language data.
+   -  ``kwargs``:
+
+      - ``normalizer`` (String, required): It should be "tokenize_hf".
```
thvasilo, line 455: Where does the normalizer name come from? This is a text to either embedding or token transformation right? Why do we only accept one value here?

docs/source/gs-processing/developer/input-configuration.rst
```diff
@@ -447,6 +447,15 @@ arguments.
...
+      - ``bert_model`` (String, required): It should be the identifier of a pre-trained model available in the Hugging Face Model Hub.
```
thvasilo, line 456: Have we tested that this works with models downloaded from the internet for both EMR Serverless and SageMaker?

docs/source/gs-processing/developer/input-configuration.rst
```diff
@@ -447,6 +447,15 @@ arguments.
...
+      - ``max_seq_length`` (Integer, required): It specifies the maximum number of tokens of the input.
```
thvasilo, line 457: Why not have a default value here?

graphstorm-processing/graphstorm_processing/config/hf_configs.py
```diff
@@ -0,0 +1,47 @@
...
+limitations under the License.
+"""
+
+from typing import Mapping
+
+from .feature_config_base import FeatureConfig
+
+
+class HFConfig(FeatureConfig):
+    """Feature configuration for single-column numerical features.
```
thvasilo, line 23: Should be text features right?

graphstorm-processing/graphstorm_processing/config/hf_configs.py
```diff
@@ -0,0 +1,47 @@
...
+    bert_model: str, required
+        The name of the lm model.
+    max_seq_length: int, required
+        The maximal length of the tokenization results.
+    """
+
+    def __init__(self, config: Mapping):
+        super().__init__(config)
+        self.bert_model = self._transformation_kwargs.get("bert_model", "none")
+        self.max_seq_length = self._transformation_kwargs.get("max_seq_length", "none")
```
thvasilo, line 36: What happens when these are set to `"none"`?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hf_transformation.py
```diff
@@ -0,0 +1,123 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+from typing import Sequence
+from pyspark.sql import DataFrame
+from pyspark.sql.types import ArrayType, IntegerType, StructType, StructField
+from pyspark.sql.functions import udf
+
+import numpy as np
+from transformers import AutoTokenizer
```
thvasilo, line 22: Group third party imports together, if possible order alphabetically
```suggestion

import numpy as np
from pyspark.sql import DataFrame
from pyspark.sql.types import ArrayType, IntegerType, StructType, StructField
from pyspark.sql.functions import udf
from transformers import AutoTokenizer
```

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hf_transformation.py
```diff
@@ -0,0 +1,123 @@
...
+
+from graphstorm_processing.constants import HUGGINGFACE_TOKENIZE
+from .base_dist_transformation import DistributedTransformation
+
+
+def apply_norm(
```
thvasilo, line 28: As with the comment for the feature config, why do we call this a norm? That has a very specific meaning in the [context of numerical data](https://en.wikipedia.org/wiki/Norm_(mathematics)) which doesn't apply here.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hf_transformation.py
```diff
@@ -0,0 +1,123 @@
...
+    ----------
+    cols : Sequence[str]
+        List of column names to apply normalization to.
+    bert_norm : str
+        The type of normalization to use. Valid values is "tokenize"
+    bert_model : str
+        The name of huggingface model.
+    input_df : DataFrame
+        The input DataFrame to apply normalization to.
+    """
```
thvasilo, line 44: Update docstring accordligly

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hf_transformation.py
```diff
@@ -0,0 +1,123 @@
...
+
+    if bert_norm == HUGGINGFACE_TOKENIZE:
```
thvasilo, line 46: Can this be not true? What happens in that case?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hf_transformation.py
```diff
@@ -0,0 +1,123 @@
...
+        schema = StructType(
+            [
+                StructField("input_ids", ArrayType(IntegerType())),
+                StructField("attention_mask", ArrayType(IntegerType())),
+                StructField("token_type_ids", ArrayType(IntegerType())),
+            ]
+        )
+
+        # Define UDF
+        @udf(returnType=schema)
```
thvasilo, line 60: Can we determine Spark version somehow and use arrow UDFs when available?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hf_transformation.py
```diff
@@ -0,0 +1,123 @@
...
+    bert_norm : str
+        The type of normalization to use. Valid values is "tokenize"
+    """
+
+    def __init__(
+        self, cols: Sequence[str], normalizer: str, bert_model: str, max_seq_length: int
+    ) -> None:
+        super().__init__(cols)
+        self.cols = cols
+        assert len(self.cols) == 1, "Bert transformation only supports single column"
```
thvasilo, line 109: Shouldn't we say HuggingFace transformation?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hf_transformation.py
```diff
@@ -0,0 +1,123 @@
...
+        self.bert_norm = normalizer
+        self.bert_model = bert_model
+        self.max_seq_length = max_seq_length
+
+    def apply(self, input_df: DataFrame) -> DataFrame:
+        scaled_df = apply_norm(
```
thvasilo, line 115: There's no scaler applied here so the var name doesn't make sense. Let's use naming that corresponds to the use-case, I think we're overindexing on the numerical transform and re-using names from there.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -960,19 +957,30 @@ def _process_node_features(
                 }
                 node_type_feature_metadata[feat_name] = node_feature_metadata_dict
 
+                feat_val = single_feature_df.take(1)[0].asDict().get(feat_name, None)
+                nfeat_size = 1 if isinstance(feat_val, (int, float)) else len(feat_val)
+                ntype_feat_sizes.update({feat_name: nfeat_size})
+
+                # pylint: disable=cell-var-from-loop
```
thvasilo, line 964: Can we take a second look at https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/cell-var-from-loop.html

Why is the error being raised here, let's ensure we're not falling into the error described in the docs

graphstorm-processing/tests/test_dist_huggingface_transformation.py
```diff
@@ -0,0 +1,72 @@
...
+            max_length=max_seq_length,
+            truncation=True,
+            padding="max_length",
+            return_tensors="np",
+        )
+
+        # Compare the Spark DataFrame output with the expected tokenizer output
+        expected_output = tokenized_data[feature]
+        for idx, row in enumerate(output_data):
+            assert_array_equal(
```
thvasilo, line 70: Is this output guaranteed to be exactly identical or do we need to add some error tolerance?

thvasilo (changes requested):
Overall the implementation looks solid. Let's work on some changes before merging.

docs/source/gs-processing/developer/input-configuration.rst
jalencato: It is used to differ the embedding or token transformation, but currently we do not implement embedding transformation for this PR, so we only accept token transformation for now.

docs/source/gs-processing/developer/input-configuration.rst
jalencato: For SageMaker, the internet access works well, nothing else needs to be deployed. For EMR Serverless, either NAT or local cache in the docker image is necessary. Either of these two options can work.

docs/source/gs-processing/developer/input-configuration.rst
jalencato: It matches the behavior of gconstruct.

graphstorm-processing/tests/test_dist_huggingface_transformation.py
jalencato: It should be exactly identical, as the result are all tokenized result with integer output, there should be no accuracy problem here.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hf_transformation.py
jalencato: With the low Spark version, defining a UDF with @udf(returnType=schema, useArrow=True) is even illegal and throws error during compiler stage. So I would suggest to move to arrow UDFs once Spark 3.5 is all available.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hf_transformation.py
jalencato: Add an if statement in config.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hf_transformation.py
thvasilo: Yes, I was talking about something like:

```python
import pyspark
from packaging.version import parse as version_parse

if version_parse(pyspark.__version__) >= version_parse("3.5.0")
    # Use arrow UDF
else:
    # Use plain UDF
```

Might be too complex though, so perhaps we'll leave arrow support for the next release.

graphstorm-processing/tests/test_converter.py
```diff
@@ -351,4 +362,4 @@ def test_convert_gsprocessing(converter: GConstructConfigConverter):
             "type": "classification",
             "split_rate": {"train": 0.9, "val": 0.1, "test": 0.0},
         },
-    ]
+    ]
```
thvasilo, line 365: Let's add the final newline here.

docs/source/gs-processing/developer/input-configuration.rst
thvasilo: Let's include the necessary changes to the 0.2.2 EMR-S Dockerfile in this PR then.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hf_transformation.py
thvasilo: Let's add an else clause that fails with a `ValueError` to be safe.

docs/source/gs-processing/developer/input-configuration.rst
thvasilo: I would prefer this to be more descriptive then, let's use `action` instead of `normalizer`. 

So for now the only supported `action` will be `tokenize_hf`, and in the future we'll add `embedding_hf`.

thvasilo:
```suggestion
      - ``action`` (String, required): The action to perform on the text data. Currently we only support text tokenization through HuggingFace models, so the only accepted value here is "tokenize_hf".
```

docs/source/gs-processing/developer/input-configuration.rst
thvasilo: I'm also a bit confused here, if we don't support embedding generation how is the HF model relevant, are there different tokenization models?

jalencato: Cause you need the pretrained models to do the tokenization: https://huggingface.co/transformers/v3.0.2/model_doc/auto.html#autotokenizer

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hf_transformation.py
jalencato: Agree. I would suggest to do it later, and we can apply it to all kinds of feature transformation.

graphstorm-processing/docker/0.2.1/emr-serverless/Dockerfile.cpu
thvasilo, line 1: 0.2.1 Dockerfiles are now in maintenance mode, let's create new Dockerfiles for v0.2.2 for both SageMaker and EMR-S.

docs/source/gs-processing/usage/emr-serverless.rst
```diff
@@ -98,6 +98,9 @@ Here you will need to replace ``<aws-account-id>``, ``<arch>`` (``x86_64`` or ``
 from the image you just created. GSProcessing version ``0.2.1`` uses ``emr-6.13.0`` as its
 base image, so we need to ensure our application uses the same release.
 
+Additionally, if it is required to use text feature transformation, it is suggested to setup VPC and NAT route for the emr cluster:
+`Create a VPC on emr-serverless
+<https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/vpc-access.html>`_
```
thvasilo, line 103: Let's mention the option of including the model in the image.

docs/source/gs-processing/usage/distributed-processing-setup.rst
```diff
@@ -111,6 +111,16 @@ The script also supports other arguments to customize the image name,
...
 
+For EMR Serverless images, setting up a VPC and NAT route is a necessary step when using text data feature transformation.
+You can find detailed instructions on creating a VPC for EMR Serverless in the AWS documentation: `Create a VPC on emr-serverless
+<https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/vpc-access.html>`_.
+Alternatively, there is one easier way to do that, you can opt to include the huggingface model cache directly in your Docker image.
+The build_gsprocessing_image.sh script provides an option to embed the huggingface bert model cache within the Docker image.
+
+.. code-block:: bash
+
+    bash docker/build_gsprocessing_image.sh --environment emr-serverless --model-name bert-base-uncased
```
thvasilo, line 122: Actually I think it should be OK to pre-install the models on SageMaker images as well right? We should change/move this section to reflect that.

graphstorm-processing/tests/test_converter.py
```diff
@@ -56,7 +56,7 @@ def test_try_read_unsupported_feature(converter: GConstructConfigConverter, node
     node_dict["nodes"][0]["features"] = [
         {
             "feature_col": ["paper_title"],
-            "transform": {"name": "tokenize_hf"},
+            "transform": {"name": "bert_hf"},
```
thvasilo, line 59: This change is aimed to trigger an expected error that we don't support this transformation yet?

thvasilo (changes requested):
Small changes requested.

graphstorm-processing/tests/test_converter.py
jalencato: Yes. As currently we have already accepte tokenize_hf, we have to change it to bert_hf

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hf_transformation.py
```diff
@@ -0,0 +1,131 @@
...
+) -> DataFrame:
+    """Applies a single normalizer to the imputed dataframe, individually to each of the columns
+    provided in the cols argument.
+
+    Parameters
+    ----------
+    cols : Sequence[str]
+        List of column names to apply normalization to.
+    action : str
+        The type of normalization to use. Valid values is "tokenize"
```
thvasilo, line 38:
```suggestion
        The type of normalization to use. Currently we only accept the `tokenize_hf` action.
```

docs/source/gs-processing/developer/input-configuration.rst
```diff
@@ -447,6 +447,15 @@ arguments.
         will be considered as an array. For Parquet files, if the input type is ArrayType(StringType()), then the
         separator is ignored; if it is StringType(), it will apply same logic as in CSV.
 
+-  ``huggingface``
+
+   -  Transforms a text feature column to tokens or embeddings with different Hugging Face models, enabling nuanced understanding and processing of natural language data.
+   -  ``kwargs``:
+
+      - ``action`` (String, required): The action to perform on the text data. Currently we only support text tokenization through HuggingFace models, so the only accepted value here is "tokenize_hf".
```
thvasilo, line 455: Can we add some more detail about what the `tokenize_hf` action does in a sub-bullet here? Imagine it's a use seeing this transformation for the first time, what do they need to know to use it (and what output to expect).

graphstorm-processing/docker/build_gsprocessing_image.sh
```diff
@@ -23,7 +23,9 @@ Available options:
 -i, --image         Docker image name, default is 'graphstorm-processing'.
 -v, --version       Docker version tag, default is the library's current version (`poetry version --short`)
 -s, --suffix        Suffix for the image tag, can be used to push custom image tags. Default is "".
--b, --build         Docker build directory, default is '/tmp/`
+-b, --build         Docker build directory, default is '/tmp/'.
+-m, --model         Huggingface Model name that needs to be packed into the docker image. Default is "".
```
thvasilo, line 27:
```suggestion
-m, --hf-model         Huggingface Model name that needs to be packed into the docker image. Default is "".
```

graphstorm-processing/docker/build_gsprocessing_image.sh
```diff
@@ -86,6 +89,10 @@ parse_params() {
       SUFFIX="${2-}"
       shift
       ;;
+    -m | --MODEL)
```
thvasilo, line 92:
```suggestion
    -m | --hf-model)
```

docs/source/gs-processing/usage/distributed-processing-setup.rst
```diff
@@ -111,6 +111,18 @@ The script also supports other arguments to customize the image name,
 tag and other aspects of the build. See ``bash docker/build_gsprocessing_image.sh --help``
 for more information.
 
+For EMR Serverless images, setting up a VPC and NAT route is a necessary step when using text data feature transformation.
+You can find detailed instructions on creating a VPC for EMR Serverless in the AWS documentation: `Create a VPC on emr-serverless
+<https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/vpc-access.html>`_.
+Alternatively, there is one easier way to do that, you can opt to include the huggingface model cache directly in your Docker image.
+It is available for both SageMaker docker image and EMR-serverless docker image. It is a good way to save cost as it avoids downloading when launching the clusters.
+The build_gsprocessing_image.sh script provides an option to embed the huggingface bert model cache within the Docker image.
```
thvasilo, line 119: Let's lead the text with the easier option.
```suggestion
If you plan to use text transformations that utilize Huggingface model, you can opt to include the Huggingface model cache directly in your Docker image.
The build_gsprocessing_image.sh script provides an option to embed the huggingface bert model cache within the Docker image, using the `--hf-model` argument.
You can do this for both the SageMaker docker image and EMR Serverless docker image. It is a good way to save cost as it avoids downloading models after launching the job.
If you'd rather download the Huggingface models at runtime, for EMR Serverless images, setting up a VPC and NAT route is a necessary.
You can find detailed instructions on creating a VPC for EMR Serverless in the AWS documentation: `Create a VPC on emr-serverless
<https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/vpc-access.html>`_.

```

graphstorm-processing/pyproject.toml
```diff
@@ -19,6 +19,8 @@ pandas = "^1.3.5"
 psutil = "^5.9.5"
 sagemaker = "^2.83.0"
 scipy = "^1.10.1"
+transformers = "^4.37.1"
+setuptools = "59.6.0"
```
thvasilo, line 23: This would only be a requirement for EMR-S 7.0.0, which we're not using currently, let's remove it.

thvasilo: When we release an EMR 7 Dockerfile we'll handle this particular dependency using a virtualenv on the image instead of modifying the project's dependencies.

docs/source/gs-processing/usage/emr-serverless.rst
```diff
@@ -98,6 +98,10 @@ Here you will need to replace ``<aws-account-id>``, ``<arch>`` (``x86_64`` or ``
 from the image you just created. GSProcessing version ``0.2.1`` uses ``emr-6.13.0`` as its
 base image, so we need to ensure our application uses the same release.
 
+Additionally, if it is required to use text feature transformation, it is suggested to setup VPC and NAT route for the emr cluster:
+`Create a VPC on emr-serverless
+<https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/vpc-access.html>`_. If it is preferred to save model cache
+inside the docker image, check out how we do that in: :doc:`distributed-processing-setup`.
```
thvasilo, line 104: Let's mention the easier build time option here, but mention that "at the time of this writing the size limit for EMR Serverless images is 5GB (https://docs.aws.amazon.com/emr/latest/EMR-Serverless-UserGuide/application-custom-image.html#considerations)"

thvasilo (changes requested):
A few changes left.

Let's ensure the following:

* run `black` on the codebase to ensure formatting is correct.
* Increment the version in pyproject.toml to `0.2.2`.
* Ensure both images build and can run a text transformation using the new build argument on both EMR-S and SM.

thvasilo (approved):
LGTM, we should run some benchmarks to see if we have any bottlenecks with the implementation.

jalencato:
> LGTM, we should run some benchmarks to see if we have any bottlenecks with the implementation.

Update our result on the mag with the tokenized feature.

PR #707, opened 2024-01-18
[GSProcessing] Fix support for large integer values for noop transfomation

### LINKED ISSUE #706: Large integer fix for GSProcessing

thvasilo:
For now we also parse all numerical CSV values as float64. In the future we'll allow the use the option to choose, defaulting at float32.

*Issue #, if available:* Fixes #706

*Description of changes:*

* For CSV numerical feature input always parse as float64. We plan to allow this to be user-defined in the future.
* For CSV regression label input always parse as float64. We plan to allow this to be user-defined in the future.
* For Parquet numerical feature input for noop transformation, use the input's data type, to avoid precision loss by converting to float32.
* For CSV numerical input for noop transformation (which is parsed as string), output to float64.

*How this was tested*:

* Added relevant unit test
* Tested on generated graph dataset with integer features with values up to 80M, results shown below

image

### DIFF
--- a/graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_noop_transformation.py
+++ b/graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_noop_transformation.py
@@ -17,18 +17,18 @@ from typing import List, Optional
 
 from pyspark.sql import DataFrame
 from pyspark.sql import functions as F
-from pyspark.sql.types import ArrayType, FloatType, NumericType
+from pyspark.sql.types import ArrayType, DoubleType, NumericType
 
 from graphstorm_processing.constants import SPECIAL_CHARACTERS
 from .base_dist_transformation import DistributedTransformation
 
 
 class NoopTransformation(DistributedTransformation):
-    """A no-op transformation that parses data as floats or lists of floats
-    and forwards the result withouth any processing.
+    """A no-op transformation that parses data as numerical values
+    and forwards the result without any processing.
 
     For CSV input that contains numerical array rows that use a separator character,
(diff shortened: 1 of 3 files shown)


graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_noop_transformation.py
```diff
@@ -66,14 +66,18 @@ def apply(self, input_df: DataFrame) -> DataFrame:
                     f"for column {self.cols[0]}"
                 )
             return input_df
+        elif isinstance(col_datatype, NumericType):
```
jalencato, line 69: Is this for parquet input to keep the same data format?

zhjwy9343:
A higher level question: because the current change will read in data as Double, do we allow users to specify the output data type to be Float? Or you will make the output type change in the next PR?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_noop_transformation.py
thvasilo: Yes, if the incoming data has a numeric type, we just preserve it.

thvasilo:
> A higher level question: because the current change will read in data as Double, do we allow users to specify the output data type to be Float? Or you will make the output type change in the next PR?

The data are read in as Double only when the source is CSV, and only for numerical features and regression labels. Right now the output type is determined by Spark, for example scaling+imputation produces Doubles, and we don't recast those to avoid precision issues.

In the follow up PR the user will be able to optionally define the input/output data type (they should always be the same).

zhjwy9343 (approved):
LGTM

PR #708, opened 2024-01-18
BugFix: Fix the bug when fixed_test_size is set while the number of testing edges are smaller than fixed_test_size

classicsong:
*What is the bug:*
When fixed_test_size is set while the number of testing edges are smaller than fixed_test_size, GraphStorm only print a warning of setting the fixed_test_size to the number of testing edges, but it does not take the real action. In certain cases, the link prediction test dataloader will generate empty mini-batches which will crash the testing process.

*Description of changes:*
Fix the bug and update the unit tests.

### DIFF
--- a/python/graphstorm/dataloading/dataloading.py
+++ b/python/graphstorm/dataloading/dataloading.py
@@ -991,6 +991,8 @@ class GSgnnLinkPredictionTestDataLoader():
                                 "is %d, which is smaller than the expected"
                                 "test size %d, force it to %d",
                                 etype, len(t_idx), self._fixed_test_size[etype], len(t_idx))
+                self._fixed_test_size[etype] = len(t_idx)
+
         self._negative_sampler = self._prepare_negative_sampler(num_negative_edges)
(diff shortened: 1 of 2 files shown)


python/graphstorm/dataloading/dataloading.py
```diff
@@ -991,6 +991,8 @@ def __init__(self, dataset, target_idx, batch_size, num_negative_edges, fanout=N
                                 "is %d, which is smaller than the expected"
                                 "test size %d, force it to %d",
                                 etype, len(t_idx), self._fixed_test_size[etype], len(t_idx))
+                self._fixed_test_size[etype] = len(t_idx)
```
zhjwy9343, line 994: If the `t_idx` is empty, will `len(t_idx)` be 0?

classicsong: t_idx will be an empty tensor. and len(t_idx) will be 0.

zhjwy9343 (approved):
LGTM

PR #709, opened 2024-01-19
Add a tool to do graph data sanity check

classicsong:
*Issue #, if available:*
In some cases, a user may provide a graph with invalid features. For example, some node features are NaN or Inf. 

*Description of changes:*
The script introduced in this PR can be used by users to do sanity checks to see whether the node/edge features and node/edge masks are valid.

### DIFF
--- a/tools/graph_sanity_check.py
+++ b/tools/graph_sanity_check.py
@@ -183,21 +183,25 @@ def verify_feats(feat, error_prefix):
     """
     # check whether feat has Nan
     has_nan = th.isnan(feat).any().item()
+    err_msg = None
+    warn_msg = None
     if has_nan:
-        return f"ERROR: {error_prefix} There are NaN values in the feature, please check.", None
+        err_msg = f"ERROR: {error_prefix} There are NaN values in the feature, please check."
 
     # check whether feat has inf
     has_inf = th.isinf(feat).any().item()
     if has_inf:
-        return f"ERROR: {error_prefix} There are infinite values in the feature, please check.", None
+        inf_err_msg = f"ERROR: {error_prefix} There are infinite values in the feature, please check."
+
+        err_msg = f"{err_msg}\n{inf_err_msg}" if err_msg is not None else inf_err_msg
 
     # check whether the value is normalized between -1 to 1
     larger_1 = (feat > 1.0).any().item()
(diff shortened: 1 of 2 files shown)


tools/graph_sanity_check.py
```diff
@@ -0,0 +1,288 @@
...
+    has_nan = th.isnan(feat).any().item()
+    if has_nan:
+        return f"ERROR: {error_prefix} There are NaN values in the feature, please check.", None
+
+    # check whether feat has inf
+    has_inf = th.isinf(feat).any().item()
+    if has_inf:
+        return f"ERROR: {error_prefix} There are infinite values in the feature, please check.", None
+
+    # check whether the value is normalized between -1 to 1
```
jalencato, line 194: Is it necessary to customized the normalized range to an input range?

tools/graph_sanity_check.py
```diff
@@ -0,0 +1,288 @@
...
+        Return
+        ------
+            str: the error message if the mask is invalid otherwise None
+    """
+    try:
+        # DGL use nonzero_1d to translate a mask into indices of nodes/edges.
+        # so we check the mask using dgl.backend.nonzero_1d
+        mask = F.nonzero_1d(mask)
+    except:
+        return f"ERROR: {error_prefix} The mask is not loadable for loadable. " \
```
zhjwy9343, line 165: "for loadable" -> "for loading"?

tools/graph_sanity_check.py
```diff
@@ -0,0 +1,288 @@
...
+                 "Ideally, the mask should be a 1D bool tensor."
+    return None
+
+def verify_feats(feat, error_prefix):
```
zhjwy9343, line 169: In case when there are multiple errors in node/edge features, can we group these errors together and return them. Otherwise, users need to run this tool a few times to find out all of these errors.

tools/graph_sanity_check.py
classicsong: No. But it is recommended.

tools/graph_sanity_check.py
classicsong: Updated

zhjwy9343 (approved):
LGTM

PR #710, opened 2024-01-19
[Bugfix] Fix a bug in launch.py when it is lauching remap commands.

classicsong:
*Issue #, if available:*
When launching remap commands in a distributed env, launch.py continuously appends new arguments into the launch command. For example:

```
PYTHONPATH=/opt/conda/bin:/root/dgl/tools:/opt/ml/code/graphstorm/python; /opt/conda/bin/python3 -m graphstorm.gconstruct.remap_result --cf config.yaml --restore-model-path /opt/ml/gsgnn_model --save-embed-path /tmp/infer_output/embs --save-embed-path /tmp/infer_output/embs --ip-config ip_list.txt --part-config metadata.json --rank 0 --world-size 16 --with-shared-fs False --num-processes 1 --output-chunk-size 100000 --preserve-input False --rank 1 --world-size 16 --with-shared-fs False --num-processes 1 --output-chunk-size 100000 --preserve-input False

PYTHONPATH=/opt/conda/bin:/root/dgl/tools:/opt/ml/code/graphstorm/python; /opt/conda/bin/python3 -m graphstorm.gconstruct.remap_result --cf config.yaml --restore-model-path /opt/ml/gsgnn_model --save-embed-path /tmp/infer_output/embs --save-embed-path /tmp/infer_output/embs --ip-config ip_list.txt --part-config metadata.json --rank 0 --world-size 16 --with-shared-fs False --num-processes 1 --output-chunk-size 100000 --preserve-input False --rank 1 --world-size 16 --with-shared-fs False --num-processes 1 --output-chunk-size 100000 --preserve-input False --rank 2 --world-size 16 --with-shared-fs False --num-processes 1 --output-chunk-size 100000 --preserve-input False
```

*Description of changes:*
Use `copy.deepcopy(udf_command)` to avoid appending args to the original `udf_command` in the `wrap_dist_remap_command` function.

### DIFF
--- a/python/graphstorm/run/launch.py
+++ b/python/graphstorm/run/launch.py
@@ -331,7 +331,7 @@ def wrap_dist_remap_command(
         with_shared_fs: bool,
         num_trainers: int,
         output_chunk_size: int = 100000,
-        preserve_input: bool = False):
+        preserve_input: bool = False) -> str:
     """ Wrap distributed remap command
 
         Parameters
--- a/python/graphstorm/run/launch.py
+++ b/python/graphstorm/run/launch.py
@@ -30,6 +30,7 @@ import signal
 import subprocess
 import sys
 import time
+import copy
 from functools import partial
 from threading import Thread
 from typing import Optional
@@ -324,7 +325,7 @@ def construct_torch_dist_launcher_cmd(
     )
 
 def wrap_dist_remap_command(
-        udf_command: str,
+        udf_command: list,
         rank: int,
         world_size: int,
         with_shared_fs: bool,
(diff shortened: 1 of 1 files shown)


python/graphstorm/run/launch.py
```diff
@@ -30,6 +30,7 @@
 import subprocess
 import sys
 import time
+import copy
 from functools import partial
 from threading import Thread
 from typing import Optional
```
thvasilo, line 36:
```suggestion
from typing import List, Optional
```
See comment below.

python/graphstorm/run/launch.py
```diff
@@ -324,7 +325,7 @@ def construct_torch_dist_launcher_cmd(
     )
 
 def wrap_dist_remap_command(
-        udf_command: str,
+        udf_command: list,
```
thvasilo, line 328: Typing hinting generics for standard collections (e.g. `list`, `dict`) were introduced in Python 3.9 ([PEP-585](https://peps.python.org/pep-0585/)), to support Python>=3.7 use:

```suggestion
        udf_command: List[str],
```

python/graphstorm/run/launch.py
```diff
@@ -324,7 +325,7 @@ def construct_torch_dist_launcher_cmd(
     )
 
 def wrap_dist_remap_command(
```
thvasilo, line 327: Can we take the chance to add a `Returns` docstring?

e.g.

```
        Returns
        -------
        A string that is the original command list, joined and modified to
        include the provided arguments.
```

We can also annotate the return type as string

```
 preserve_input: bool = False) -> str:
```

thvasilo (changes requested):
Some changes need to ensure Python >=3.7 support.

thvasilo (approved):
LGTM now. Can we add a return type annotation for `wrap_dist_remap_command`? 

```
def wrap_dist_remap_command(
        udf_command: List[str],
        rank: int,
        world_size: int,
        with_shared_fs: bool,
        num_trainers: int,
        output_chunk_size: int = 100000,
        preserve_input: bool = False) -> str:
```

python/graphstorm/run/launch.py
```diff
@@ -349,29 +350,35 @@ def wrap_dist_remap_command(
...
-    udf_command += ["--with-shared-fs", "True" if with_shared_fs else "False"]
-    udf_command += ["--num-processes", str(num_trainers)]
-    udf_command += ["--output-chunk-size", str(output_chunk_size)]
-    udf_command += ["--preserve-input", "True" if preserve_input else "False"]
+    new_udf_command += ["--rank", str(rank)]
+    new_udf_command += ["--world-size", str(world_size)]
+    new_udf_command += ["--with-shared-fs", "True" if with_shared_fs else "False"]
+    new_udf_command += ["--num-processes", str(num_trainers)]
+    new_udf_command += ["--output-chunk-size", str(output_chunk_size)]
+    new_udf_command += ["--preserve-input", "True" if preserve_input else "False"]
```
thvasilo, line 374: Actually maybe we just write:

```suggestion
    new_udf_command.extend([
        "--rank", str(rank),
        "--world-size", str(world_size),
        "--with-shared-fs", "True" if with_shared_fs else "False",
        "--num-processes", str(os.cpu_count()),
        "--output-chunk-size", str(output_chunk_size),
        "--preserve-input", "True" if preserve_input else "False"]
    )
```

Seems more readable to me at least. Not that we also change the default num-processes

classicsong: I would suggest we create another PR for setting num-processes to os.cpu_count(). And we need to do some profiling to see whether os.cpu_count() is the right default number.

PR #712, opened 2024-01-20
Run training jobs without ssh in the local machine.

zheng-da:
*Description of changes:*
When running a training job on the local machine, we can avoid using ssh.

### DIFF
--- a/docs/source/install/env-setup.rst
+++ b/docs/source/install/env-setup.rst
@@ -49,7 +49,10 @@ Configure SSH No-password login
 ................................
 To perform distributed training in a cluster of machines, please use the following commands
 to configure a local SSH no-password login that GraphStorm relies on.
-*Note*: this is not needed for the standalone mode.
+
+.. note::
+
+    This is not needed for the standalone mode.
 
 .. code-block:: bash
 
--- a/docs/source/install/env-setup.rst
+++ b/docs/source/install/env-setup.rst
@@ -47,7 +47,9 @@ For CPU environment:
 
 Configure SSH No-password login
 ................................
-Use the following commands to configure a local SSH no-password login that GraphStorm relies on.
+To perform distributed training in a cluster of machines, please use the following commands
+to configure a local SSH no-password login that GraphStorm relies on.
+*Note*: this is not needed for the standalone mode.
 
 .. code-block:: bash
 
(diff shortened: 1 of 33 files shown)


zhjwy9343:
After we change this behavior, should we update any documentations?

python/graphstorm/run/launch.py
```diff
@@ -733,8 +913,56 @@ def signal_handler(sig, frame): # pylint: disable=unused-argument
...
+                    ip = result[0]
+                    port = get_available_port(ip)
+                    hosts.append((ip, port))
+                else:
+                    raise RuntimeError("Format error of ip_config.")
+    else:
+        # The user doesn't provide an IP config file. This means we are going to
+        # run the training job in the local machine. We should create a temporary
+        # IP config file.
+        with tempfile.NamedTemporaryFile(delete=False) as fp:
```
classicsong, line 957: Who will remove this file?

zhjwy9343: And why do we still need this temp file as we already know the ip address?

zheng-da: DistDGL needs this file.

zheng-da: right now it's not deleted. this is a temporary file. should we delete it?

docs/source/install/env-setup.rst
```diff
@@ -47,7 +47,9 @@ For CPU environment:
 
 Configure SSH No-password login
 ................................
-Use the following commands to configure a local SSH no-password login that GraphStorm relies on.
+To perform distributed training in a cluster of machines, please use the following commands
+to configure a local SSH no-password login that GraphStorm relies on.
+*Note*: this is not needed for the standalone mode.
```
zhjwy9343, line 52: Can you use the blow rst code to highlight this Node.

.. note::

    This operation is not needed for the standalone mode.

docs/source/tutorials/quick-start.rst
```diff
@@ -12,12 +12,6 @@ GraphStorm is designed for easy-to-use GML models, particularly the graph neural
 
 This tutorial will use GraphStorm's built-in OGB-arxiv dataset for a node classification task to demonstrate these three steps.
 
-.. warning::
```
zhjwy9343, line 15: Could you also help to remove the same warning section in the `own-data.rst` line 11-15?

zhjwy9343 (approved):
LGTM

Issue #704, opened 2024-01-16
[Bug] Large integer features will lost precision after graph construction


zhjwy9343:
## Phenomena:

If using large integer numbers (~67M) as node features, the graph construction process will convert them as float32 and store as float32 tensor. But the float32 losses precision, and when cast the float32 back to int, it will generate duplicated integer. 

## Impact:

Users want to add original node ids as one of the node features, and use them for customized purposes. The lost of precision will cause ID duplications, hence resulting errors.

## How to reproduce

- Create a graph data with more than 60M nodes, and set integer node id as one of the node features. 
- Run gconstruction.
- In the partition folder, e.g., part0, read in the `node_feat.dgl` file with `dgl.data.util.load_tensors`.
- Extract the node feature with node ids, compare the feature shape against its unique values shape.
- The two shapes are different.
```[tasklist]
### Tasks
- [ ] https://github.com/awslabs/graphstorm/issues/706
- [ ] Fix for GConstruct
```

classicsong:
Can we close this issue? @zhjwy9343

PR #717, opened 2024-01-29
[Enhance]  Support float64 as output type in graph construction

zhjwy9343:
*Issue #704, if available:*

*Description of changes:*
This PR add the float64 as one of the output dtypes for Noop transformer. This will fix the issue #704 for large integer numbers in features.


thvasilo (approved):
LGTM small change. One suggestion: in GSProcessing for Parquet files that are typed we respect the input type by default.

zhjwy9343:
> LGTM small change. One suggestion: in GSProcessing for Parquet files that are typed we respect the input type by default.

Thanks. The change is for output type, so if the users give float64 we will use float64. But the implementation before this PR converted to float32 by default.

PR #726, opened 2024-02-03
[Enhancement] Add default task tracker in trainers and inferrers

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
This PR add a default task tracker in Trainers and Inferrers. And then Trainers and Inferrers use the task tracker to initialize an Evaluator. The new logic is that :
1. In constructor, set default task_tracker  to be None;
2. Users can setup task tracker by using setup_task_tracker();
3. When setup the Evaluator in Trainers and Inferrers, check 
    - if they have a task tracker preset, and the Evaluator does not have task tracker set, use the task tracker to set the Evaluator;
    - if they have no a task tracker, and the Evaluator either, create a default SageMakerTaskTracker using the eval_frequency of the Evaluator, and then setup the Evaluator with the newly created task tracker.
    - if the Evaluator already has a task tracker, then just setup the Evaluator without any addition operations.
4. Because trainers require evaluator, whenever setup a evaluator, a task_tracker will be added to trainers.


python/graphstorm/trainer/gsgnn_trainer.py
```diff
@@ -89,15 +88,23 @@ def setup_task_tracker(self, task_tracker):
...
+        frequency.
         """
-        if self.task_tracker is not None:
-            evaluator.setup_task_tracker(self.task_tracker)
+        if evaluator.task_tracker is None:
+            if self.task_tracker is not None:
+                task_tracker = GSSageMakerTaskTracker(evaluator.eval_frequency)
+                evaluator.setup_task_tracker(task_tracker)
+            else:
+                evaluator.setup_task_tracker(self.task_tracker)
```
classicsong, line 106: ditto

classicsong (commented):
Can you also add some unit test for "If the evaluator has its own task tracker, just setup the evaluator. But if the evaluator has no task tracker, will use this Trainer's task tracker to setup the evaluator."

python/graphstorm/eval/evaluator.py
```diff
@@ -338,6 +338,18 @@ def history(self):
...
+        """
+        return self._eval_frequency
+
+    @property
+    def task_tracker(self):
+        """ Task tracker of this evaluator
+        """
+        return self.tracker
+
 class GSgnnRegressionEvaluator(GSgnnInstanceEvaluator):
```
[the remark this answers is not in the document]
zhjwy9343: They both extend the `GSgnnInstanceEvaluator` so they also extend the `setup_task_tracker()` method.

python/graphstorm/trainer/gsgnn_trainer.py
zhjwy9343: The self.task_tracker belongs to the trainer, and setup_task_tracker will need this argument.

zhjwy9343:
> Can you also add some unit test for "If the evaluator has its own task tracker, just setup the evaluator. But if the evaluator has no task tracker, will use this Trainer's task tracker to setup the evaluator."

I add two unit-test files for trainer and inferrer. So far just add test cases of setup task tracker.

classicsong (approved):

PR #729, opened 2024-02-07
[Bug Fix] Local Docker Build

jalencato:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docker/Dockerfile.local
+++ b/docker/Dockerfile.local
@@ -9,6 +9,7 @@ RUN apt-get install -y python3-pip git wget psmisc
 RUN apt-get install -y cmake
 
 # Install Pytorch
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.local
```diff
@@ -9,6 +9,7 @@ RUN apt-get install -y python3-pip git wget psmisc
 RUN apt-get install -y cmake
 
 # Install Pytorch
+RUN pip3 install networkx==3.1
```
classicsong, line 12: Which component needs this?

jalencato: Torch needs this. Currently the version torch automatically download is only compatible with python 3.9 while the base image only have 3.8.

classicsong (approved):

PR #731, opened 2024-02-09
[GSProcessing] Make SageMaker image build faster and smaller.

thvasilo:
*Issue #, if available:*

*Description of changes:*
* We switch away from using `pipenv` in the SageMaker image, relying instead on on the poetry-generated requirements.txt file to install dependencies. 
* Also switch to images having `prod` target by default (although this requires `poetry` to be installed on the image building environment).
* Update out-of-date README files.

# Testing

Unit tests+Ran a SageMaker job with ml-25m data.

### DIFF
--- a/graphstorm-processing/docker/0.2.2/emr-serverless/Dockerfile.cpu
+++ b/graphstorm-processing/docker/0.2.2/emr-serverless/Dockerfile.cpu
@@ -31,7 +31,7 @@ RUN yum erase -y openssl-devel && \
         sudo \
         xz-devel && \
         rm -rf /var/cache/yum
-RUN git clone https://github.com/pyenv/pyenv.git ${PYENV_ROOT} && \
+RUN git clone https://github.com/pyenv/pyenv.git ${PYENV_ROOT} --single-branch && \
     pyenv install ${PYTHON_VERSION} && \
     pyenv global ${PYTHON_VERSION}
 
@@ -41,7 +41,7 @@ WORKDIR /usr/lib/spark/code/
 COPY requirements.txt requirements.txt
 # Use --mount=type=cache,target=/root/.cache when Buildkit CI issue is fixed:
 # https://github.com/moby/buildkit/issues/1512
-RUN pip install -r /usr/lib/spark/code/requirements.txt \
+RUN pip install --no-cache-dir -r /usr/lib/spark/code/requirements.txt \
     && rm -rf /root/.cache
 
 # GSProcessing codebase
@@ -63,7 +63,8 @@ RUN python -m pip install --no-deps /usr/lib/spark/code/graphstorm_processing-*.
     rm /usr/lib/spark/code/graphstorm_processing-*.whl && rm -rf /root/.cache
 
 FROM runtime AS test
-RUN python -m pip install --no-deps /usr/lib/spark/code/graphstorm-processing/ && rm -rf /root/.cache
+RUN python -m pip install --no-deps /usr/lib/spark/code/graphstorm-processing/ mock && \
+    rm -rf /root/.cache
 
 USER hadoop:hadoop
 WORKDIR /home/hadoop
--- a/graphstorm-processing/docker/0.2.2/sagemaker/Dockerfile.cpu
+++ b/graphstorm-processing/docker/0.2.2/sagemaker/Dockerfile.cpu
@@ -11,32 +11,17 @@ ENV LC_ALL=C.UTF-8
 ENV LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/usr/local/lib"
 ENV LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/opt/conda/lib"
 ENV PATH=/opt/conda/bin:$PATH
+ENV PIP_NO_CACHE_DIR=1
 
-# Install GSProcessing requirements to pipenv Python
-RUN pipenv install \
-    boto3==1.28.38 \
-    joblib==1.3.1 \
-    mock==5.1.0 \
-    pandas==1.3.5 \
-    pip==23.1.2 \
-    protobuf==3.20.3 \
-    psutil==5.9.5 \
-    pyarrow==13.0.0 \
-    pyspark==3.4.1 \
-    scipy==1.11.3 \
-    setuptools \
-    transformers==4.37.1 \
-    spacy==3.6.0 \
-    torch==2.1.0 \
-    wheel \
-    && rm -rf /root/.cache
-# Do a pipenv sync so our base libs are independent from our editable code, making them cacheable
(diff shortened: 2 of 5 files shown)


graphstorm-processing/docker/README.md
```diff
@@ -12,34 +12,48 @@ with Amazon SageMaker see docs/source/usage/distributed-processing-setup.rst.
 ## Building the image
 
 To build the image you will run `bash build_gsprocessing_image.sh`
-script that has one required parameter, `--target` that can take
-one of two values, `prod` and `test` that determine whether we
-include the source and tests on the image (when `test` is used),
-or just install the libary on the image (when `prod` is used).
```
jalencato, line 18: Why do we delete the target description here? Although we currently enable the prod as the default one. We still provide option to let user choose the target for prod and test.

graphstorm-processing/docker/0.2.2/emr-serverless/Dockerfile.cpu
```diff
@@ -63,7 +63,8 @@ RUN python -m pip install --no-deps /usr/lib/spark/code/graphstorm_processing-*.
     rm /usr/lib/spark/code/graphstorm_processing-*.whl && rm -rf /root/.cache
 
 FROM runtime AS test
-RUN python -m pip install --no-deps /usr/lib/spark/code/graphstorm-processing/ && rm -rf /root/.cache
+RUN python -m pip install --no-deps /usr/lib/spark/code/graphstorm-processing/ mock && \
```
jalencato, line 66: Do we need to install pytest here? I check our pyproject.toml does not include pytest package here.

graphstorm-processing/docker/0.2.2/sagemaker/Dockerfile.cpu
```diff
@@ -60,10 +45,11 @@ fi
...
-RUN python3 -m pip install /usr/lib/spark/code/graphstorm_processing-*.whl && \
-    rm /usr/lib/spark/code/graphstorm_processing-*.whl
+RUN python3 -m pip install --no-deps /usr/lib/spark/code/graphstorm_processing-*.whl && \
+    rm /usr/lib/spark/code/graphstorm_processing-*.whl && rm -rf /root/.cache
 CMD ["gs-processing"]
 
 FROM base AS test
-RUN python3 -m pip install /usr/lib/spark/code/graphstorm-processing/
-CMD ["sh", "-c", "pytest ./code/tests/"]
+RUN python3 -m pip install --no-deps /usr/lib/spark/code/graphstorm-processing/ mock && \
```
jalencato, line 53: Same here, do we need to install pytest?

graphstorm-processing/docker/0.2.2/sagemaker/Dockerfile.cpu
```diff
@@ -11,32 +11,17 @@ ENV LC_ALL=C.UTF-8
...
-    torch==2.1.0 \
-    wheel \
-    && rm -rf /root/.cache
-# Do a pipenv sync so our base libs are independent from our editable code, making them cacheable
-RUN pipenv sync --system && python3 -m spacy download en_core_web_lg \
+WORKDIR /usr/lib/spark/code/
+
+# Install GSProcessing dependencies to system Python 3.9
+COPY requirements.txt requirements.txt
+RUN /usr/local/bin/python3.9 -m pip install --no-cache-dir -r /usr/lib/spark/code/requirements.txt \
```
classicsong, line 20: We may need transformers as we are going to support BERT.

jalencato: We have transformers here, as we already have the transformers installed in the requirements.txt. It is generated by poetry.

graphstorm-processing/docker/README.md
thvasilo: The main text here used to explain the only _required_ argument that was previously target (when this was written). The only required argument now is environment.

We add the description of target as one of the arguments below.

graphstorm-processing/docker/0.2.2/sagemaker/Dockerfile.cpu
thvasilo: Pytest is already installed on the image, but `mock` is missing.

graphstorm-processing/docker/0.2.2/sagemaker/Dockerfile.cpu
thvasilo: Yes to clarify this `requirements.txt` file is [generated at build-time by poetry,](https://github.com/thvasilo/graphstorm/blob/c85b20404ae3bb564636f6bb3bcfc1a76290278b/graphstorm-processing/docker/build_gsprocessing_image.sh#L150-L151) based on the dependencies set in pyproject.toml (which include transformers).

graphstorm-processing/docker/0.2.2/emr-serverless/Dockerfile.cpu
thvasilo: Testing in the SageMaker image at least it seems like the env already has pytest. We can rely on that for now, or have an explicit optional dependency group for it (like we currently have the `dev` dependency group).

graphstorm-processing/docker/0.2.2/sagemaker/Dockerfile.cpu
```diff
@@ -11,32 +11,17 @@ ENV LC_ALL=C.UTF-8
 ENV LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/usr/local/lib"
 ENV LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:/opt/conda/lib"
 ENV PATH=/opt/conda/bin:$PATH
+ENV PIP_NO_CACHE_DIR=1
 
-# Install GSProcessing requirements to pipenv Python
-RUN pipenv install \
```
jalencato, line 16: Compared with the previous pipenv and current pip install options, the current one ran 1 hour on amazon review, but previous one ran 40 minutes on the amazon review data processing. Will pipenv here do something different to have that influence?

thvasilo: Very unlikely, but not impossible (at this point I've seen everything). Let me try with some reference datasets for which I know the expected runtime and see if it makes a consistent difference.

thvasilo: Test by doing 2 runs for 2 jobs, comparing runtimes with these changes vs. previous version of the image. The runtimes are very close (+/- 5% which is expected for SM). 

We can take a look at replicating this particular issue for this dataset and see if the difference is persistent (could be for example including MKL or other env difference). For this PR though I'd consider this a non-issue.

jalencato (approved):
LGTM, the performance number matches our expectation as well.

PR #736, opened 2024-02-14
[Bug fix] Add NCCL test for save embedding during training and fix device bugs.

classicsong:
*Issue #, if available:*
There is a bug in all the training scripts, when `save_embed_path` is set to a path and the backend is set to `nccl`. The restored model (used for computing the embedding with the best model) is not moved to GPU and will cause a `nccl` all_reduce error.

*Description of changes:*
Fix the bug.

### DIFF
--- a/tests/end2end-tests/graphstorm-ec/mgpu_test.sh
+++ b/tests/end2end-tests/graphstorm-ec/mgpu_test.sh
@@ -24,7 +24,7 @@ error_and_exit () {
 }
 
 echo "**************dataset: Generated multilabel MovieLens EC, RGCN layer: 1, node feat: generated feature, inference: full graph, exclude-training-targets: True"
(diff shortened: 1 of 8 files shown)


tests/end2end-tests/graphstorm-ec/mgpu_test.sh
```diff
@@ -89,7 +89,7 @@ then
...
+python3 -m graphstorm.run.gs_edge_classification --inference --workspace $GS_HOME/inference_scripts/ep_infer --num-trainers $NUM_INFO_TRAINERS --num-servers 1 --num-samplers 0 --part-config /data/movielen_100k_multi_label_ec/movie-lens-100k.json --ip-config ip_list.txt --ssh-port 2222 --cf ml_ec_infer.yaml  --multilabel true --num-classes 5 --node-feat-name movie:title user:feat --use-mini-batch-infer false --save-embed-path /data/gsgnn_ec/infer-emb/ --restore-model-path /data/gsgnn_ec/epoch-$best_epoch/ --save-prediction-path /data/gsgnn_ec/prediction/ --logging-file /tmp/log.txt  --logging-level debug --preserve-input True --backend nccl
```
jalencato, line 92: shall we keep original test? It may be necessary to keep test on both nccl and gloo. Or we can just add a couple of nccl tests .

jalencato (approved):
LGTM

zheng-da (approved):

PR #405, opened 2023-08-31
Wholegraph integration

isratnisa:
This PR leverages the [WholeGraph](https://github.com/rapidsai/wholegraph) framework to accelerate node feature transfer across multiple machines.  Users can enable `--use-wholegraph` flag when using EFA-enabled devices with the NCCL backend to take full advantage of this capability.

Node features are expected to be stored in a WholeGraph-compatible format on disk. Users can utilize the `convert_feat_to_wholegraph.py` script to transform the features (`'feat'`) from the `node_feat.dgl` file into a format compatible with WholeGraph. These transformed features are then saved as `.../wholegraph/<ntype>~<feat>`. The node_feat.dgl file is subsequently trimmed to remove the feat attribute, leaving only other attributes such as `train_mask`, `test_mask`, and `val_mask` intact. The original file is stored as `node_feat.dgl.bak`.

PR #737, opened 2024-02-14
[WholeGraph] Add support of using WholeGraph to store/load cache_lm_emb

chang-l:
Draft PR (todo: add some unit tests)
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/python/graphstorm/config/argument.py
+++ b/python/graphstorm/config/argument.py
@@ -307,6 +307,7 @@ class GSConfig:
         if self.node_lm_configs:
             _ = self.lm_infer_batch_size
             _ = self.freeze_lm_encoder_epochs
+        _ = self.use_wholegraph_cache_lm_embed
 
         if self.distill_lm_configs:
             _ = self.textual_data_path
@@ -687,6 +688,18 @@ class GSConfig:
         """
         if hasattr(self, "_cache_lm_embed"):
             return self._cache_lm_embed
+        else:
+            return None
+
+    @property
+    def use_wholegraph_cache_lm_embed(self):
+        """ Whether to cache the LM embeddings on files by using WholeGraph.
+        """
+        if hasattr(self, "_use_wholegraph_cache_lm_embed"):
+            if self._use_wholegraph_cache_lm_embed:
+                assert self.cache_lm_embed, "You must turn on cache_lm_embed " \
+                    "to use wholegraph cache lm embeddings."
+            return self._use_wholegraph_cache_lm_embed
         else:
             return None
 
@@ -2479,6 +2492,12 @@ def _add_lm_model_args(parser):
             type=lambda x: (str(x).lower() in ['true', '1']),
             default=argparse.SUPPRESS,
             help="Whether to cache the LM embeddings in files. " + \
+                    "If the LM embeddings have been saved before, load the saved embeddings " + \
+                    "instead of computing the LM embeddings again.")
+    group.add_argument("--use-wholegraph-cache-lm-embed",
+            type=lambda x: (str(x).lower() in ['true', '1']),
+            default=argparse.SUPPRESS,
+            help="Whether to use WholeGraph to cache the LM embeddings in files. " + \
                     "If the LM embeddings have been saved before, load the saved embeddings " + \
                     "instead of computing the LM embeddings again.")
     return parser
--- a/python/graphstorm/model/lm_embed.py
+++ b/python/graphstorm/model/lm_embed.py
@@ -31,7 +31,12 @@ from .embed import GSNodeInputLayer
 from .embed import GSNodeEncoderInputLayer
 from .lm_model import init_lm_model
 from .lm_model import get_lm_node_feats
-from .utils import load_pytorch_embedding, save_pytorch_embedding
+from .utils import (
+    load_pytorch_embedding,
+    save_pytorch_embedding,
+    load_wholegraph_embedding,
+    save_wholegraph_embedding
+)
 from ..utils import get_rank, get_world_size, create_dist_tensor
 from ..distributed import flush_data
 
@@ -239,11 +244,12 @@ class LMCache:
     embed_path : str
         The path where the embedding files are stored.
     """
-    def __init__(self, g, lm_models, embed_path=None):
+    def __init__(self, g, lm_models, embed_path=None, use_wg=False):
         self._g = g
         self._lm_models = lm_models
         self._lm_emb_cache = {}
         self._embed_path = embed_path
+        self._use_wg = use_wg
         self._lm_hash = ''
 
     def _get_model_hash(self, ntype):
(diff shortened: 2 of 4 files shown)


python/graphstorm/config/argument.py
```diff
@@ -2481,6 +2494,12 @@ def _add_lm_model_args(parser):
             help="Whether to cache the LM embeddings in files. " + \
                     "If the LM embeddings have been saved before, load the saved embeddings " + \
                     "instead of computing the LM embeddings again.")
+    group.add_argument("--use-wholegraph-cache-lm-embed",
```
zheng-da, line 2497: do we need an argument to turn on this? if a user uses wholegraph, why is it not turned on automatically?

chang-l: >if a user uses wholegraph, why is it not turned on automatically?

To clarify, do you mean as long as WholeGraph is installed/available on user-end, we turn on this feature automatically? Note that WholeGraph has limitations and might not work in certain scenarios (for now), e.g., (1) needs all procs alive/participate in its operations (dist_tensor creation/gather/scatter) if using the default NCCL backend; (2) needs all files accessible by all machines/procs when loading into WholeGraph memory. However, distDGL (default) is more flexible.

On the other hand, there are many wholegraph-related options/functionalities currently in gs (`feature store`, `sparse embedding`, this PR, etc). At some point, I think we should have a unified flag to turn on them all to improve usability.

zheng-da: there is a flag that can turn on wholegraph support in graphstorm. i don't think we need a specific flag to turn on wholegraph for LM cache.

zheng-da: if wholegraph support is turned on, we should enable wholegraph for every possible component in graphstorm.

zheng-da:
everything else looks good to me.

python/graphstorm/config/argument.py
chang-l: > there is a flag that can turn on wholegraph support in graphstorm. 

Which flag/argument? I traced back to the first PR of adding WholeGraph for feature store https://github.com/awslabs/graphstorm/pull/405. It seems it turned on WholeGraph implicitly by file name pattern matching, i.e.,:
https://github.com/awslabs/graphstorm/blob/main/python/graphstorm/utils.py#L102-L108 

And this PR: https://github.com/awslabs/graphstorm/pull/677  uses an argument `--use-wholegraph-sparse-emb` to turn on WholeGraph for sparse embedding.

chang-l: >if wholegraph support is turned on, we should enable wholegraph for every possible component in graphstorm.

I agree. So, maybe we can (1) keep the way of using WholeGraph for inital feature store  as https://github.com/awslabs/graphstorm/pull/405 ; and (2) combine all other WholeGraph componetns for intermediate distributed store in one flag/argument.  Does it sound reasonable? @zheng-da @classicsong 

If so, I will change the flag name in this PR and have another PR to consolidate the wholegraph-related flags.

python/graphstorm/model/utils.py
```diff
@@ -565,6 +565,99 @@ def save_pytorch_embedding(emb_path, embedding, rank, world_size):
...
+    name : str
+        The name of the created distributed tensor.
+
+    Returns
+    -------
+    WholeGraphDistTensor : the loaded embeddings in WholeGraph.
+    """
+    with open(os.path.join(emb_path, "emb_info.json"), 'r', encoding='utf-8') as f:
+        emb_info = json.load(f)
+
```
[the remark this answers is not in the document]
chang-l: Sure.

python/graphstorm/config/argument.py
chang-l: Done the part in this PR. I will open another PR to let sparse_emb use the same argument.

python/graphstorm/model/lm_embed.py
```diff
@@ -305,10 +316,17 @@ def _save_embeddings(self):
...
-            save_pytorch_embedding(embed_path,
-                                   self._lm_emb_cache[ntype],
-                                   get_rank(),
-                                   get_world_size())
+            if self.use_wg:
+                save_wholegraph_embedding(embed_path,
+                                        self._lm_emb_cache[ntype],
+                                        get_rank(),
+                                        get_world_size(),
+                                        fmt="binary")
```
chang-l, line 324: Currently, I coded the embedding saved in `fmt="binary"` format. Let me know if you want it to be "pytorch" as default. @classicsong

classicsong: I think "binary" works.

python/graphstorm/config/argument.py
```diff
@@ -537,6 +538,18 @@ def verbose(self):
 
         return False
 
+    @property
+    def use_wholegrap(self):
```
classicsong, line 542: How about use_wholegraph_embed?
As use_wholegraph for feature store is automatically enabled if features are stored in wholegraph format, If we use `use_wholegraph` here, it is a little bit confusing.

python/graphstorm/model/utils.py
```diff
@@ -565,6 +565,137 @@ def save_pytorch_embedding(emb_path, embedding, rank, world_size):
...
+            "World_size when save the embedding is different than the current world_size. " \
+            "Please switch to the binary format."
+        wg_rank = dist_emb.get_comm().get_rank()
+        file_path = os.path.join(emb_path, f'embed-{pad_file_index(wg_rank)}.pt')
+        assert os.path.exists(file_path), f"Embedding file {file_path} of \
+            my rank {wg_rank} doesn't exist."
+        emb = th.load(file_path)
+        local_emb = dist_emb.get_local_tensor()[0]
+        assert emb.shape[0] == local_emb.shape[0] and emb.shape[1] == local_emb.shape[1], \
+            "Embedding shape does not match!"
```
classicsong, line 683:
```suggestion
            "Embedding shape of {Feature name} does not match! Expect xx, but get xx"
```

classicsong (commented):
Can you add 2 end2end tests? (one for create wg LMCache another for loading it.) You can add it in the lp mgpu test. There should be one for DistDGLTensor LMCache example.

python/graphstorm/config/argument.py
chang-l: Sounds good. I will update it accordingly

tests/end2end-tests/graphstorm-nc/mgpu_test.sh
```diff
@@ -249,6 +249,32 @@ error_and_exit $?
...
+error_and_exit $?
+
+cnt=$(ls -l /data/gsgnn_nc_ml_text/ | grep epoch | wc -l)
+if test $cnt != 1
+then
+    echo "The number of save models $cnt is not equal to the specified topk 1"
+    exit -1
+fi
+
+best_epoch=$(grep "successfully save the model to" /tmp/train_log.txt | tail -1 | tr -d '\n' | tail -c 1)
```
classicsong, line 282: Can you also check the whether the cached Bert embeddings are saved under /data/movielen_100k_lm_encoder_train_val_1p_4t/? And can you check the meta.json?

classicsong: I mean whether the `wg-embed_part_0_of_2`,  `wg-embed_part_1_of_2`, ..., are under `/data/movielen_100k_lm_encoder_train_val_1p_4t/`

classicsong (approved):
LGTM

PR #739, opened 2024-02-15
[GSProcessing] Allow custom out_dtype in feature transformation

jalencato:
*Issue #, if available:*

*Description of changes:*

* Add out_dtype support for no-op, numerical, and multi-numerical transformation for float32 for floatType and float64 for doubleType.

* Allow transformation between gconstruct and gsprocessing.


graphstorm-processing/graphstorm_processing/constants.py
```diff
@@ -44,6 +46,8 @@
 ################# Numerical transformations  ################
 VALID_IMPUTERS = ["none", "mean", "median", "most_frequent"]
 VALID_NORMALIZERS = ["none", "min-max", "standard", "rank-gauss"]
+VALID_OUTDTYPE = ["float32", "float64"]
```
thvasilo, line 49: Maybe we can make the names of types constants as well?

```
TYPE_FLOAT32 = "float32"
TYPE_FLOAT64 = "float64"
VALID_OUTDTYPE = [TYPE_FLOAT32, TYPE_FLOAT64]
```

The change all raw string refs to the constants.

graphstorm-processing/tests/test_dist_numerical_transformation.py
```diff
@@ -114,6 +122,37 @@ def test_numerical_transformation_without_transformation(input_df: DataFrame, ch
...
 
 
+@pytest.mark.parametrize("norm", ["min-max", "standard", "rank-gauss"])
+@pytest.mark.parametrize("out_dtype", ["float32", "float64"])
+def test_numerical_min_max_transformation_precision(
+    spark: SparkSession, check_df_schema, out_dtype, norm
+):
+    """Test numerical transformation without any transformation applied"""
+    # Adjust the number to be an integer
+    high_precision_integer = 1.2345678901234562
```
thvasilo, line 132: This is a float though? Are we trying to encode a large int perhaps?

graphstorm-processing/tests/test_dist_numerical_transformation.py
```diff
@@ -132,7 +171,7 @@ def test_numerical_transformation_with_median_imputer_and_std_norm(
     expected_imputed_std_ages = [0.2, 0.2, 0.1, 0.3, 0.2]
 
     for row, expected_val in zip(transformed_rows, expected_imputed_std_ages):
-        assert row["age"] == expected_val
+        assert_almost_equal(row["age"], expected_val, decimal=7)
```
thvasilo, line 174: I guess this test was unstable because of float accuracy.

thvasilo (approved):
LGTM minor comments.

graphstorm-processing/tests/test_dist_numerical_transformation.py
jalencato: Yep. But we are testing are between float32 and float64. Large int is better for LongType though, so we choose a float number here.

graphstorm-processing/tests/test_dist_numerical_transformation.py
jalencato: Certainly. We may need to add decimal for every assert_almost_equal test on float number in the future.

PR #758, opened 2024-02-29
add infer_batch_size to get a small batch size for LLM.

GentleZhu:
*Issue #, if available:*
#759 
*Description of changes:*
The default inference batch size 1024 is too large for LLM inference. Add an additional parameter to allow the infer_batch_size configuration.

### DIFF
--- a/python/graphstorm/inference/emb_infer.py
+++ b/python/graphstorm/inference/emb_infer.py
@@ -40,9 +40,9 @@ class GSgnnEmbGenInferer(GSInferrer):
     """
     def infer(self, data, task_type, save_embed_path, eval_fanout,
             use_mini_batch_infer=False,
-            infer_batch_size=1024,
             node_id_mapping_file=None,
-            save_embed_format="pytorch"):
+            save_embed_format="pytorch",
+            infer_batch_size=1024):
         """ Do Embedding Generating
 
         Generate node embeddings and save into disk.
@@ -64,6 +64,8 @@ class GSgnnEmbGenInferer(GSInferrer):
             graph partition algorithm.
         save_embed_format : str
             Specify the format of saved embeddings.
+        infer_batch_size: int
+            Specify the inference batch size of embeddings.
         """
         assert save_embed_path is not None, \
             "It requires save embed path for gs_gen_node_embedding"
--- a/python/graphstorm/inference/emb_infer.py
+++ b/python/graphstorm/inference/emb_infer.py
@@ -40,6 +40,7 @@ class GSgnnEmbGenInferer(GSInferrer):
     """
     def infer(self, data, task_type, save_embed_path, eval_fanout,
             use_mini_batch_infer=False,
+            infer_batch_size=1024,
             node_id_mapping_file=None,
             save_embed_format="pytorch"):
         """ Do Embedding Generating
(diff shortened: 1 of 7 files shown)


python/graphstorm/inference/lp_infer.py
```diff
@@ -42,6 +42,7 @@ class GSgnnLinkPredictionInferrer(GSInferrer):
     def infer(self, data, loader, save_embed_path,
             edge_mask_for_gnn_embeddings='train_mask',
             use_mini_batch_infer=False,
+            infer_batch_size=1024,
```
classicsong, line 45: You need to add this new argument at the end of the argument list to make it backward compatible.

classicsong: Also you need to add a docstring

python/graphstorm/inference/emb_infer.py
```diff
@@ -40,6 +40,7 @@ class GSgnnEmbGenInferer(GSInferrer):
     """
     def infer(self, data, task_type, save_embed_path, eval_fanout,
             use_mini_batch_infer=False,
+            infer_batch_size=1024,
```
classicsong, line 43: Same as GSgnnLinkPredictionInferrer

classicsong (commented):
You also need to update np_infer.py and ep_infer.py

jalencato (commented):
The implementation is simple but some general questions:

*  Do we want the infer_batch_size as a new input argument? As the description, the infer_batch_size is a new input, instead of set it in the code, we can make it as a new argument, and set the default as 1024. Refer to the argument code: https://github.com/awslabs/graphstorm/blob/968371b893afe0c2b2b09e9248fba9f11b75cb3a/python/graphstorm/config/argument.py#L582
* If we want to apply that to the lp task, we may need to apply this to all task.

GentleZhu:
> The implementation is simple but some general questions:
> 
> * Do we want the infer_batch_size as a new input argument? As the description, the infer_batch_size is a new input, instead of set it in the code, we can make it as a new argument, and set the default as 1024. Refer to the argument code: https://github.com/awslabs/graphstorm/blob/968371b893afe0c2b2b09e9248fba9f11b75cb3a/python/graphstorm/config/argument.py#L582
> * If we want to apply that to the lp task, we may need to apply this to all task.
> * Meanwhile, do not forget to add ready label to the PR to trigger the test.



> The implementation is simple but some general questions:
> 
> * Do we want the infer_batch_size as a new input argument? As the description, the infer_batch_size is a new input, instead of set it in the code, we can make it as a new argument, and set the default as 1024. Refer to the argument code: https://github.com/awslabs/graphstorm/blob/968371b893afe0c2b2b09e9248fba9f11b75cb3a/python/graphstorm/config/argument.py#L582
> * If we want to apply that to the lp task, we may need to apply this to all task.
> * Meanwhile, do not forget to add ready label to the PR to trigger the test.

I think it's for lp task only, for np and ep the dataloader is passed into the infer function. I think it only matters when we use co-train LM+GNN. Even that, during inference of LM+GNN, I suppose lm features are cached and no updated, so the infer_batch_size doesn't matter and it will not be OOM.

python/graphstorm/inference/emb_infer.py
```diff
@@ -63,6 +64,8 @@ def infer(self, data, task_type, save_embed_path, eval_fanout,
             graph partition algorithm.
         save_embed_format : str
             Specify the format of saved embeddings.
+        infer_batch_size: int
+            Specify the inference batch size of embeddings.
```
classicsong, line 68:
```suggestion
            Specify the inference batch size of for computing input embeddings.
```

GentleZhu: I don't quite understand word "input embeddings" here. I thought we are computing node embeddings in this function.

python/graphstorm/inference/emb_infer.py
```diff
@@ -63,6 +64,8 @@ def infer(self, data, task_type, save_embed_path, eval_fanout,
             graph partition algorithm.
         save_embed_format : str
             Specify the format of saved embeddings.
+        infer_batch_size: int
+            Specify the inference batch size of node embeddings.
```
classicsong, line 68:
```suggestion
            Specify the inference batch size when computing node embeddings with mini batch inference.
```

classicsong (approved):
LGTM

PR #760, opened 2024-02-29
Link Prediction using graph-aware fine-tuning of LLMs

GentleZhu:
*Issue #, if available:*

*Description of changes:*
This PR supports link prediction using GPEFT with random negatives. It relies on PR #758 


examples/peft_llm_gnn/llm_gnn_model.py
```diff
@@ -59,9 +61,34 @@ def get_lm_node_feats(g, lm_feat_names, lm_ntypes):
...
 
-class LLMGraphModel(gsmodel.GSgnnNodeModelBase):
-    """ LLMGNN Model Class
+    def require_cache_embed(self):
+        return False
+    
+    def forward(self, input_feats, input_nodes):
+        return input_feats
+
+class CosineRandomNegative(nn.Module):
```
zheng-da, line 74: why does it have cosine in the name?

examples/peft_llm_gnn/llm_gnn_model.py
```diff
@@ -59,9 +61,34 @@ def get_lm_node_feats(g, lm_feat_names, lm_ntypes):
...
+    def __init__(self):
+        super(CosineRandomNegative, self).__init__()
+        self.margin = 0.5
+
+    def forward(self, pos_score, neg_score):
+        pos_distance = []
+        neg_distance = []
+        for k in pos_score:
```
zheng-da, line 82: do you iterate over each element here? why not implement it in a vector way?

examples/peft_llm_gnn/llm_gnn_model.py
```diff
@@ -89,7 +116,7 @@ def __init__(self, g, node_lm_configs, h_dim, out_dim, num_layers,
                  use_norm = True,   # use normalization or not, default is True
                  alpha_l2norm = 0,
                  lr = 0.001):
-        super(LLMGraphModel, self).__init__()
+        super(GNNLLM_NC, self).__init__()
```
zheng-da, line 119: GPEFT should also be used here.

examples/peft_llm_gnn/llm_gnn_model.py
```diff
@@ -192,10 +221,169 @@ def create_optimizer(self):
...
+                 alpha_l2norm = 0,
+                 lr = 0.001):
+        super(GNNLLM_LP, self).__init__()
+        
+        self.target_etype=target_etype
+        self.alpha_l2norm = alpha_l2norm
+        self.lr = lr
+        self.gnn_encoder = GPEFT(g, h_dim, num_layers, target_ntype, node_lm_configs)
+        # dummy node input encoder to enable LP Infer
+        self.node_input_encoder = DummyNodeInputLayer()
```
zheng-da, line 349: do you have to set up this field?

examples/peft_llm_gnn/llm_gnn_model.py
GentleZhu: Originally, the normalization is done in the loss function. Now it's in the model computation.

examples/peft_llm_gnn/llm_gnn_model.py
GentleZhu: It's a dict of {node_type: scores} as output by the LinkPredictDotDecoder

examples/peft_llm_gnn/llm_gnn_model.py
GentleZhu: I intend to do this in a seperate PR that makes the previous node classification uses GPEFT component. Is this okay?

examples/peft_llm_gnn/llm_gnn_model.py
GentleZhu: I added a assert to ensure this

zheng-da: i don't know what you mean here. do you mean you normalize the embeddings? so it's equivalent to compute cosine similarity?

examples/peft_llm_gnn/llm_gnn_model.py
zheng-da: if you can remember to do so. it'll be better to do it in the same PR. it'll be more manageable.

examples/peft_llm_gnn/llm_gnn_model.py
zheng-da: where is the field used? i searched for it. it seems this is the only place where it is used.

examples/peft_llm_gnn/llm_gnn_model.py
GentleZhu: yes, cosine similarity is equivalent as dot-product with normalized vectors. And we use LinkPredictDotDecoder here.

examples/peft_llm_gnn/llm_gnn_model.py
GentleZhu: It's here. If we don't have model.node_input_encoder, it will report exception. 
https://github.com/awslabs/graphstorm/blob/b38259de64877455f8dc965582a5dc180eda807d/python/graphstorm/model/gnn.py#L985C35-L985C54

zheng-da (approved):

PR #767, opened 2024-03-13
[GSProcessing] Allow re-partitioning to run on the Spark leader

thvasilo:
*Issue #, if available:*

*Description of changes:*

With this change we allow the re-partition step to run on the leader instance, removing the need to run a separate job for it. If no files need re-partitioning, and only Parquet modification is needed for labels/masks we _always_ apply that on the leader.

To accomplish this we make the following changes:

* Add new `--do-repartition` argument
* Refactor repartition_files.py to allow separate execution of just Parquet metadata modification, and full repartition that runs both the row count alignment and the Parquet metadata modification.
    * As part of this refactoring we also split the previously verbose `main` function into multiple smaller functions to allow easier maintenance. Each step is now performed by a separate function, namely `_repartition_edge_files`, `_repartition_node_files` and `modify_flat_array_metadata`. A combined public function `repartition_files` calls these three functions as needed in sequence. All functions take the graph metadata dict as input, and a ParquetRepartitioner instance.
* Introduce checks for node data and Parquet metadata modification in our unit tests. Test coverage for `repartition_files.py` is now increased to 89% (from 81%).

## Testing

Unit testing, and ran jobs on SageMaker and EMR-S with ml25M. For large-scale random 100M-edges-10M-nodes graph, the repartition step completed successfully on SageMaker (large memory instances) and failed (gracefully) on EMR-S. When the repart step fails we print an error in the logs to indicate to the user they'll need to run the additional gs-repartition job separately.

### DIFF
--- a/docs/source/gs-processing/usage/amazon-sagemaker.rst
+++ b/docs/source/gs-processing/usage/amazon-sagemaker.rst
@@ -70,7 +70,9 @@ Launch the gs-repartition job on Amazon SageMaker
 
 
 In the above we have set `--do-repartition True` to perform the re-partition step on the Spark
-leader instance. The re-partitioning job runs on a single instance, so for large graphs you will
+leader instance, since we know each individual feature and the edge structure are small
+enough to fit in the memory of the Spark leader.
+For large graphs you will
 want to launch that step as a separate job on an instance with more memory to avoid memory errors.
 `ml.r5` instances should allow you to re-partition graph data with billions of nodes and edges.
 For more details on the re-partitioning step see :doc:`row-count-alignment`.
@@ -79,7 +81,7 @@ To run the re-partition job as a separate job use:
 
 .. code-block:: bash
 
-    # Ensure the bash variables are as set as above
+    # Ensure the bash variables are as set as above.
     # This will only run the follow-up re-partitioning job on a single instance
     python scripts/run_repartitioning.py --s3-input-prefix ${OUTPUT_PREFIX} \
         --role ${ROLE} --image ${IMAGE_URI}  --config-filename "metadata.json" \
--- a/docs/source/gs-processing/usage/amazon-sagemaker.rst
+++ b/docs/source/gs-processing/usage/amazon-sagemaker.rst
@@ -62,20 +62,31 @@ job, followed by the re-partitioning job, both on SageMaker:
         --instance-type ${INSTANCE_TYPE} \
         --job-name "${GRAPH_NAME}-${INSTANCE_COUNT}x-${INSTANCE_TYPE//./-}-${NUM_FILES}files" \
         --num-output-files ${NUM_FILES} \
+        --do-repartition True \
         --wait-for-job
 
-    # This will run the follow-up re-partitioning job
+Launch the gs-repartition job on Amazon SageMaker
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+In the above we have set `--do-repartition True` to perform the re-partition step on the Spark
+leader instance. The re-partitioning job runs on a single instance, so for large graphs you will
+want to launch that step as a separate job on an instance with more memory to avoid memory errors.
+`ml.r5` instances should allow you to re-partition graph data with billions of nodes and edges.
+For more details on the re-partitioning step see :doc:`row-count-alignment`.
+
+To run the re-partition job as a separate job use:
+
+.. code-block:: bash
+
+    # Ensure the bash variables are as set as above
+    # This will only run the follow-up re-partitioning job on a single instance
     python scripts/run_repartitioning.py --s3-input-prefix ${OUTPUT_PREFIX} \
         --role ${ROLE} --image ${IMAGE_URI}  --config-filename "metadata.json" \
         --instance-type ${INSTANCE_TYPE} --wait-for-job
 
 
-.. note::
 
-    The re-partitioning job runs on a single instance, so for large graphs you will
-    want to scale up to an instance with more memory to avoid memory errors. `ml.r5` instances
-    should allow you to re-partition graph data with billions of nodes and edges.
-    For more details on the re-partitioning step see ::doc:`row-count-alignment`.
 
 The ``--num-output-files`` parameter
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/source/gs-processing/usage/amazon-sagemaker.rst
+++ b/docs/source/gs-processing/usage/amazon-sagemaker.rst
@@ -86,8 +86,6 @@ To run the re-partition job as a separate job use:
         --instance-type ${INSTANCE_TYPE} --wait-for-job
 
 
-
-
 The ``--num-output-files`` parameter
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
--- a/docs/source/gs-processing/usage/row-count-alignment.rst
+++ b/docs/source/gs-processing/usage/row-count-alignment.rst
@@ -67,7 +67,7 @@ Local repartitioning
 --------------------
 
 The simplest way to apply the re-partitioning step is to do so during the `gs-processing` step,
-by passing the additional `--repartition-on-leader True` argument to our launch script.
+by passing the additional `--do-repartition True` argument to our launch script.
 
 Alternatively, we can run a local re-partitioning job using a local
 installation of GSProcessing:
--- a/docs/source/gs-processing/usage/row-count-alignment.rst
+++ b/docs/source/gs-processing/usage/row-count-alignment.rst
@@ -66,7 +66,10 @@ after the processing step. This step performs two functions:
 Local repartitioning
 --------------------
 
-The simplest way to apply the re-partitioning step is to do so using a local
+The simplest way to apply the re-partitioning step is to do so during the `gs-processing` step,
+by passing the additional `--repartition-on-leader True` argument to our launch script.
+
+Alternatively, we can run a local re-partitioning job using a local
 installation of GSProcessing:
 
 .. code-block:: bash
@@ -101,7 +104,7 @@ on SageMaker:
     INSTANCE_TYPE="ml.t3.xlarge"
 
     python scripts/run_repartitioning.py --s3-input-prefix ${PROCESSED_OUTPUT} \
-        --role ${ROLE} --image ${IMAGE_URI}  --config-filename "metadata.json" \
+        --role ${ROLE} --image ${IMAGE_URI} \
         --instance-type ${INSTANCE_TYPE} --wait-for-job
 
 File streaming repartitioning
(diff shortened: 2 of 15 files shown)


graphstorm-processing/tests/test_repartition_files.py
```diff
@@ -396,6 +410,33 @@ def mock_json_load(file):
...
+                    filemeta = pq.read_metadata(absolute_feature_filepath)
+                    file_rows = filemeta.num_rows
                     assert (
-                        expected_count == pq.read_metadata(absolute_feature_filepath).num_rows
+                        expected_count == file_rows
                     ), f"Count mismatch for {feature_name}, {absolute_feature_filepath}"
+                    # Ensure the flat array has the correct metadata embedded
+                    if feature_name == "label" and task_type not in {
+                        "link_predict",
+                        "link_prediction",
```
jalencato, line 421: Why do we have two kinds of labels for link prediction here? ( Feel free to correct me if there is anything wrong

jalencato (commented):
Overall, it looks good to me. Some high-level comments:

*  Do we need to use flag name with --repartition-on-leader? I think for users, they do not need to know something like "on-leader", it maybe confused for them to understand "on-leader" here. We can make it with just "--do-repartition"
* Is there any need to still keep the repartition process separate? Since most of users suffer from repartition problem, maybe we can just make it a must-do thing inside the GSProcessing and the repartition will become a black-box thing.

graphstorm-processing/tests/test_repartition_files.py
thvasilo: Needed for legacy with NeptuneML

graphstorm-processing/graphstorm_processing/graph_loaders/row_count_utils.py
```diff
@@ -85,25 +85,6 @@ def add_row_counts_to_metadata(self, metadata_dict: dict) -> dict:
         )
         self._add_counts_for_features(top_level_key="node_data", edge_or_node_type_key="node_type")
```
classicsong, line 87: shall we call verify_metadata_match here?

graphstorm-processing/graphstorm_processing/repartition_files.py
```diff
@@ -90,11 +94,14 @@ def __init__(
...
             self.bucket = self.input_prefix.split("/")[1]
-            self.pyarrow_fs = fs.S3FileSystem(region=region)
+            self.pyarrow_fs = fs.S3FileSystem(
+                region=region, retry_strategy=fs.AwsDefaultS3RetryStrategy(max_attempts=10)
+            )
         else:
             self.pyarrow_fs = fs.LocalFileSystem()
         self.verify_outputs = verify_outputs
         self.streaming_repartitioning = streaming_repartitioning
+        self.verbosity = 1 if logging.getLogger().getEffectiveLevel() <= logging.INFO else 0
```
classicsong, line 104: Why not use logging to control?

graphstorm-processing/graphstorm_processing/repartition_files.py
```diff
@@ -764,95 +771,66 @@ def parse_args(args):
...
-    # We first collect the most frequent row counts, to minimize the number of
-    # repartitions we have to perform.
-    if edge_data_exist:
-        edge_data_meta = metadata_dict["edge_data"]
-        # TODO: Frequencies are important, but so is data size,
-        # we would like to avoid downloading large feature files if possible
-        edge_row_counts_frequencies = collect_frequencies_for_data_counts(edge_data_meta)
+    if edge_data_meta:
+        task_type = metadata_dict["graph_info"].get("task_type", "link_prediction")  # type: str
+        print(f"{task_type=}")
```
classicsong, line 833: logging

graphstorm-processing/graphstorm_processing/repartition_files.py
```diff
@@ -861,197 +839,280 @@ def main():
...
                         etype_label_property,
                         type_name,
                     )
-                    label_files = type_data_dict[etype_label_property]["data"]  # type: List[str]
-                    # TODO: These ops are probably safe to parallelize, since the tables are likely
-                    # easy to fit in memory (one float/bool per row)
                     repartitioner.modify_metadata_for_flat_arrays(label_files)
                 for mask in ["train_mask", "test_mask", "val_mask"]:
-                    logging.info("Modifying metadata for '%s' of edge type '%s'", mask, type_name)
                     if mask in type_data_dict:
```
classicsong, line 887: What is the difference between re-partitioning edge features and re-partitioning edge masks?

classicsong: Is this because, masks are usually 1D?

classicsong: Can you also add a TODO here: when supporting multi-task training, it is possible that the masks may have other names like train_mask_a, train_mask_b, etc.

classicsong: Same for labels.

graphstorm-processing/scripts/run_distributed_processing.py
```diff
@@ -45,8 +45,11 @@
     Number of EC2 instances to use for the processing job. (Default: 2)
 --instance-type: str
     EC2 instance type for the processing job. (Default: 'ml.r5.4xlarge')
---add-reverse-edges: bool
-    When set to True, will create reverse edges for every edge type. (Default: True)
+--add-reverse-edges: str
+    When set to "True", will create reverse edges for every edge type. (Default: "True")
```
classicsong, line 49: Why not use bool?

thvasilo: Just consistency, the entry point scripts for SageMaker need to use String values, so we do the same here. 

I can change to to a flag for the launcher if you think that makes more sense.

docs/source/gs-processing/usage/row-count-alignment.rst
```diff
@@ -101,7 +104,7 @@ on SageMaker:
     INSTANCE_TYPE="ml.t3.xlarge"
 
     python scripts/run_repartitioning.py --s3-input-prefix ${PROCESSED_OUTPUT} \
-        --role ${ROLE} --image ${IMAGE_URI}  --config-filename "metadata.json" \
+        --role ${ROLE} --image ${IMAGE_URI} \
```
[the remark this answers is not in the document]
thvasilo: Implicitly no, the default config-filename is "metadata.json" and that's what `gs-processing` generates.

graphstorm-processing/graphstorm_processing/distributed_executor.py
```diff
@@ -224,9 +237,52 @@ def run(self) -> None:
...
+        logging.info("Time to transform data for distributed partitioning: %s sec", t1 - t0)
+        # Stop the Spark context
+        self.spark.stop()
 
-        # This is used to upload the output JSON files to S3 on local runs,
+        all_match = verify_metadata_match(graph_meta_dict)
+        repartitioner = ParquetRepartitioner(
+            self.output_prefix,
+            self.filesystem_type,
+            region=None,
```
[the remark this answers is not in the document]
thvasilo: Agreed, I just want to communicate the intent here, even if someone is not aware of the `ParquetRepartitioner` defaults they know exactly the kind of re-partition we will perform here.

graphstorm-processing/graphstorm_processing/distributed_executor.py
```diff
@@ -224,9 +237,52 @@ def run(self) -> None:
...
+                    with open(
+                        os.path.join(loader.output_path, "updated_row_counts_metadata.json"),
+                        "w",
+                        encoding="utf-8",
+                    ) as f:
+                        json.dump(updated_metadata, f, indent=4)
+                        f.flush()
+                    logging.info("Data are now ready to be fed to DistPart pipeline")
+                except Exception as e:  # pylint: disable=broad-exception-caught
+                    logging.error(
```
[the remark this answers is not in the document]
thvasilo: Just want to print an error to allow the main (gs-processing) job to finish successfully, and only print an error indicating that we should run `gs-repartition` as a follow up.

graphstorm-processing/graphstorm_processing/graph_loaders/row_count_utils.py
thvasilo: We now do this inside distributed_executor.py, line 247.

graphstorm-processing/graphstorm_processing/repartition_files.py
thvasilo: joblib has its own verbosity level, which is an int, 0 being no logging. We impute how much logging the user wants from the log level they have chosen. I actually think I'll set it to 0 and only increase to >0 for logging.DEBUG level.

graphstorm-processing/graphstorm_processing/repartition_files.py
thvasilo: > Is this because, masks are usually 1D?

Yes, DistPart expects label and mask arrays to be marked as flat using a metadata entry, so this function, `modify_flat_array_metadata` does that.

Will add the TODO.

thvasilo:
@jalencato changed the user-facing argument (sagemaker launch) to `--do-repartition`, let me know if you think I should change the internal script as well (distributed_executor.py arguments)

jalencato (approved):
LGTM

docs/source/gs-processing/usage/amazon-sagemaker.rst
```diff
@@ -62,21 +62,30 @@ job, followed by the re-partitioning job, both on SageMaker:
...
+        --do-repartition True \
         --wait-for-job
 
-    # This will run the follow-up re-partitioning job
+Launch the gs-repartition job on Amazon SageMaker
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+In the above we have set `--do-repartition True` to perform the re-partition step on the Spark
+leader instance. The re-partitioning job runs on a single instance, so for large graphs you will
```
classicsong, line 73:
```suggestion
leader instance, when the re-partitioning job runs on a single instance. For large graphs you will
```

graphstorm-processing/graphstorm_processing/distributed_executor.py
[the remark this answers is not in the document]
classicsong: Can you put some comments here? 
As catching an exception without throwing it out is not a normal routine.

classicsong (approved):

docs/source/gs-processing/usage/amazon-sagemaker.rst
thvasilo: >In the above we have set `--do-repartition True` to perform the re-partition step on the Spark
leader instance, when the re-partitioning job runs on a single instance. For large graphs you will [..]

By `when the re-partitioning job runs on a single instance` do we mean to say when it's possible to run re-partitioning on a single instance?

PR #774, opened 2024-03-19
feature: add support for PyTorch Lightning

clumsy:
*Issue #, if available:* N/A

*Description of changes:* Adds PyTorch Lightning support, `LightningModule` and `LightningDataModule` for Node GNN model and also a Jupyter nodebook demonstrating how it works.

### DIFF
--- a/examples/lightning/node_gnn.ipynb
+++ b/examples/lightning/node_gnn.ipynb
@@ -1,0 +1,413 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "9f7fa7f2-8b5c-4690-8f5b-800dee3f8e86",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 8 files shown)


thvasilo:
Hello @clumsy took a closer look at the PR and had a couple of requests. We'd like to maintain a similar level of requirements between GSProcessing and this library.

* We should try to have some unit tests if possible.
* Let's put some more work pyproject.toml. Right now in the dev group I see a lot of dependencies that I'm not sure the project needs. For DGL in particular there's only support for py3.10 on Linux?
* Let's add some documentation. The example notebook is helpful but we also want to have a guide that helps users from installation to training, with AWS as the target cloud infra. At least one example on how to run this on SageMaker should be included.
* This also means that we likely have to include a compatible Dockerfile. If the existing SageMaker and local Dockerfiles can be slightly modified to add PTL support, we can have that in documentation.

Issue #754, opened 2024-02-27
[Change Request] Set default values to GraphStorm's `gs.initialize()` function


zhjwy9343:
1. For users to initialize GraphStorm standalone environment more easier, need to set default argument value for initialize() method, i.e., `ip_config=None`, `backend='gloo'`, `device='cpu'`, and `log_level='info'`.

2. In order to facilitate convert code written in APIs to run with CLI interface, the GraphStorm need to provide `gs.get_device()` method for device setup internally in other classes, e.g., DataLoaders, Trainers, and Inferrers, without users' input.

Sample code
```
import graphstorm as gs
gs.initialize()
```

classicsong:
To Note: log_level is not set by `gs.initialize`

zhjwy9343:
How about device? Can gs.initalize() set up device?

PR #782, opened 2024-03-25
Change logic of setup device

classicsong:
*Issue #, if available:*
#754 

*Description of changes:*
1. provide a global get_device function for any function to get the current computation device.
2. remove the device argument for any DataLoader. If a DataLoader want to know which device to use for some computation, it should use get_device to acquire the device to use.
3. Update tests accordingly.


zhjwy9343 (approved):
Except for the format comments, others LGMT.

PR #790, opened 2024-03-29
[Partition Image] Add dependency for gs partition

jalencato:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/python/graphstorm/gpartition/dist_partition_graph.py
+++ b/python/graphstorm/gpartition/dist_partition_graph.py
@@ -39,7 +39,8 @@ def run_build_dglgraph(
         ip_list,
         output_path,
         metadata_filename,
-        dgl_tool_path):
+        dgl_tool_path,
+        ssh_port):
     """ Build DistDGL Graph
 
     Parameters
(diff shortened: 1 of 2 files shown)


python/graphstorm/gpartition/dist_partition_graph.py
```diff
@@ -153,6 +157,7 @@ def parse_args() -> argparse.Namespace:
                            help="Path to store the partitioned data")
     argparser.add_argument("--num-parts", type=int, required=True,
                            help="Number of partitions to generate")
+    argparser.add_argument("--ssh-port", type=int, default=22, help="SSH Port")
```
thvasilo, line 160: Here do we expect to have more common execution on Docker or plain EC2? Maybe we should default to `2222`?

thvasilo (approved):
LGTM minor question

python/graphstorm/gpartition/dist_partition_graph.py
jalencato: Keep the tradition as we do as we always do: https://github.com/awslabs/graphstorm/blob/6e63e4c458b2967e247b23ecbd143313a5ec3445/python/graphstorm/run/launch.py#L1140

PR #791, opened 2024-04-02
[GSProcessing] Move to Spark 3.5, bump version to 0.3.0, add homogeneous edge mapping optimization

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Optimizations for edge re-mapping:
    * When an edge is homogeneous (same src and dst node type) we cache and re-use the node id mapping DF instead of loading from storage twice. 
    * Remove enforced re-partitions where possible, because they'd trigger entire DF shuffles.
    * These changes combined have improved total job runtime by up to 33% in a random graph with 10B edges and 1B nodes.
* Bumps GSProcessing version to 0.3.0.
* Allows Spark up to 3.5 in project dependencies.
* Adds 0.3.0 Dockerfiles with Spark 3.5 for emr-s and sagemaker.

*Testing*

pytest, test jobs on SageMaker and EMR-S using the latest image versions with ml-100k, and large-scale test on SM with 1B-edges-100M-nodes-1024feat graph.

### DIFF
--- a/graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
+++ b/graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
@@ -818,12 +818,16 @@ class DistHeterogeneousGraphLoader(HeterogeneousGraphLoader):
         self.graph_info["ntype_label"] = []
         self.graph_info["ntype_label_property"] = []
         for node_config in node_configs:
-            logging.info("node_config: %s", node_config)
             files = node_config.files
             file_paths = [f"{self.input_prefix}/{f}" for f in files]
 
             node_type = node_config.ntype
             node_col = node_config.node_col
+            logging.info(
+                "Processing data for node type %s with config: %s",
+                node_type,
+                node_config,
+            )
 
             read_nodefile_start = perf_counter()
             # TODO: Maybe we use same enforced type for Parquet and CSV
@@ -879,6 +883,7 @@ class DistHeterogeneousGraphLoader(HeterogeneousGraphLoader):
                     NODE_MAPPING_INT
                 )
             else:
+                logging.info("Creating node str-to-int mapping for node type: %s", node_type)
                 nodes_df = self.create_node_id_map_from_nodes_df(nodes_df, node_col)
                 self._write_nodeid_mapping_and_update_state(nodes_df, node_type)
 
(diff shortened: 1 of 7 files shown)


graphstorm-processing/pyproject.toml
```diff
@@ -10,7 +10,7 @@ authors = [
 
 [tool.poetry.dependencies]
 python = "~3.9.12"
-pyspark = ">=3.3.0, < 3.5.0"
+pyspark = ">=3.3.0, <= 3.5.0"
```
jalencato, line 13: I see spark has 3.5.1 version available on Feb.23rd, 2024, may we change here to 3.5.1?

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -1114,10 +1119,13 @@ def write_edge_structure(
             )
             .withColumnRenamed(NODE_MAPPING_INT, "src_int_id")
             .withColumnRenamed(NODE_MAPPING_STR, "src_str_id")
-            .repartition(self.num_output_files, F.col("src_str_id"))
         )
+
+        # If edge is homogeneous, we'll re-use the same mapping for both src and dst
+        if src_ntype == dst_ntype:
+            src_node_id_mapping.cache()
```
jalencato, line 1126: A question here: why do we only do cache when src_ntype == dst_ntype? Whether we should always do that?

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -1173,6 +1183,8 @@ def write_edge_structure(
 
         edge_structure_path = os.path.join(self.output_prefix, f"edges/{edge_type}")
         logging.info("Writing edge structure for edge type %s...", edge_type)
+        if self.add_reverse_edges:
```
jalencato, line 1186: Same question here.

graphstorm-processing/pyproject.toml
thvasilo: Good point, I'll change to `<3.6.0`

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: If the edge type is heterogeneous, the node_id mapping DF will only end up being used once so there's no point in trying to cache it.

Caching makes sense in occasions where a DF is being used more than once so we want to maintain it in cluster memory.

jalencato (approved):
LGTM

PR #808, opened 2024-04-20
[Docker] Move local image to CUDA 12.1. Add option for CPU local image.

thvasilo:
*Issue #, if available:*

*Description of changes:*

Motivation: we want to be able to run local CPU clusters for cost-efficient partition and inference. Also the 11.7 Nvidia image will be EOL and removed by Nvidia in July 2024.

* Move local image to CUDA 12.1, ~~DGL 2.1.0~~
* Add the option to create CPU images for local clusters, allowing us to run partition/inference in CPU local clusters.
* Also changes tagging for SageMaker image to include image type as tag suffix, cpu/gpu.

### Testing

Created local CPU and GPU clusters on EC2, run example ACM job.

### DIFF
--- a/docker/Dockerfile.local
+++ b/docker/Dockerfile.local
@@ -1,64 +1,74 @@
 ARG DEVICE=gpu
+ARG SOURCE
 
-FROM nvidia/cuda:12.1.0-runtime-ubuntu20.04 as branch-gpu
-ENV dev_type=GPU
+FROM ${SOURCE} as base
 
-RUN apt update && apt install -y --no-install-recommends \
-    python3-dev \
-    python3-pip \
-    && rm -rf /var/lib/apt/lists/*
-
-# Install DGL and Pytorch for GPU
-RUN pip3 install torch==2.1.2 \
-    && rm -rf /root/.cache
-
-RUN pip3 install networkx==3.1 pydantic dgl==2.1.0+cu121 \
-    -f https://data.dgl.ai/wheels/cu121/repo.html && rm -rf /root/.cache
-
-FROM public.ecr.aws/ubuntu/ubuntu:22.04_stable as branch-cpu
-
-# Install DGL and Pytorch for CPU
-ENV dev_type=CPU
-
-RUN apt update && apt install -y --no-install-recommends \
-    python3-dev \
-    python3-pip \
-    && rm -rf /var/lib/apt/lists/*
-
-RUN pip3 install torch==2.1.2 --index-url https://download.pytorch.org/whl/cpu
-
-RUN pip3 install networkx==3.1 pydantic dgl==2.1.0 \
-    -f https://data.dgl.ai/wheels-internal/repo.html && rm -rf /root/.cache
-
-FROM branch-${DEVICE} as runtime
-
-ENV PYTHONPATH="/root/dgl/tools/:${PYTHONPATH}"
 ENV DEBIAN_FRONTEND=noninteractive
 ENV HOME=/root
 
+ARG DGL_VERSION=1.1.3
+ARG OGB_VERSION=1.3.6
+ARG TORCH_VERSION=2.1.2
+ARG TRANSFORMERS_VERSION=4.28.1
+
 RUN apt update && apt install -y --no-install-recommends \
-    build-essential \
-    cmake \
-    cython3 \
-    g++ \
     git \
     libicu-dev \
-    make \
-    net-tools \
     openssh-client \
     openssh-server \
-    psmisc \
-    unzip \
-    wget \
+    python3.9 \
+    python3.9-distutils \
+    python3.9-venv \
     && rm -rf /var/lib/apt/lists/*
 
+# Create and activate a Python venv
+RUN python3.9 -m venv /opt/gs-venv
+ENV PATH="/opt/gs-venv/bin:$PATH"
 
 # Install GraphStorm dependencies
-RUN pip3 install ogb==1.3.6 scipy pyarrow boto3 scikit-learn transformers h5py psutil \
+RUN pip install \
+    boto3==1.34.89 \
+    botocore==1.34.89 \
+    h5py==3.11.0 \
+    networkx==3.1 \
+    psutil==5.9.8 \
+    pyarrow==14.0.0 \
+    pydantic==2.7.0 \
+    scikit-learn==1.4.2 \
+    scipy==1.13.0 \
     && rm -rf /root/.cache
 
+FROM base as base-cpu
+
+# Install torch, DGL, and GSF deps that require torch
+RUN pip install \
+    torch==${TORCH_VERSION} \
+    --index-url https://download.pytorch.org/whl/cpu \
+    && rm -rf /root/.cache
+
+RUN pip install \
+    dgl==${DGL_VERSION} \
+    ogb==${OGB_VERSION} \
+    transformers==${TRANSFORMERS_VERSION} \
+    -f https://data.dgl.ai/wheels-internal/repo.html && rm -rf /root/.cache
+
+FROM base as base-gpu
+
+# Install torch, DGL, and GSF deps that require torch
+RUN pip install \
+    dgl==${DGL_VERSION}+cu121 \
+    ogb==${OGB_VERSION} \
+    torch==${TORCH_VERSION} \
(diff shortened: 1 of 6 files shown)


docker/Dockerfile.local
```diff
@@ -1,34 +1,66 @@
...
+    python3-dev \
+    python3-pip \
+    && rm -rf /var/lib/apt/lists/*
 
-# Install Pytorch
-RUN pip3 install networkx==3.1 pydantic
-RUN pip3 install torch==2.1.0+cu118 --extra-index-url https://download.pytorch.org/whl/cu118
+RUN pip3 install torch==2.1.2 --index-url https://download.pytorch.org/whl/cpu
+
+RUN pip3 install networkx==3.1 pydantic dgl==2.1.0 \
```
jalencato, line 30: I have met a problem for this, check issue here: https://github.com/awslabs/graphstorm/issues/809 It is the problem with DGL with version >= 2.0.0, but it has no problem with DGL == 1.1.3

jalencato: I have also created an issue on DGL repo: https://github.com/dmlc/dgl/issues/7333, we can use DGL 1.1.3 version first.

thvasilo: Ok I'll try to reproduce this and can keep using DGL 1.1.x for now.

docs/source/scale/sagemaker.rst
```diff
@@ -54,9 +54,9 @@ The ``build_docker_sagemaker.sh`` script takes four arguments:
 .. warning::
     In order to upload the GraphStorm SageMaker Docker image to Amazon ECR, users need to define the <DOCKER_NAME> to include the ECR URI string, **<AWS_ACCOUNT_ID>.dkr.ecr.<REGION>.amazonaws.com/**, e.g., ``888888888888.dkr.ecr.us-east-1.amazonaws.com/graphstorm``.
 
-4. **DOCKER_TAG** (optional), is the assigned tag name of the to-be built Docker image. Default is ``sm``.
+4. **DOCKER_TAG** (optional), is the assigned tag name of the to-be built Docker image. Default is ``sm-<cpu/gpu>``.
```
classicsong, line 57: `sm-<cpu/gpu>` is a little bit confusing. Let's use `sm-cpu` and `sm-gpu` for cpu image and gpu image respectively.

docs/source/install/env-setup.rst
```diff
@@ -121,13 +121,14 @@ Please use the following command to build a Docker image from source:
...
+    bash /path-to-graphstorm/docker/build_docker_oss4local.sh /path-to-graphstorm/ docker-name docker-tag device
 
-There are three arguments of the ``build_docker_oss4local.sh``:
+There are four positional arguments for ``build_docker_oss4local.sh``:
 
 1. **path-to-graphstorm** (**required**), is the absolute path of the "graphstorm" folder, where you cloned the GraphStorm source code. For example, the path could be ``/code/graphstorm``.
 2. **docker-name** (optional), is the assigned name of the to be built Docker image. Default is ``graphstorm``.
-3. **docker-tag** (optional), is the assigned tag name of the to be built docker image. Default is ``local``.
+3. **docker-tag** (optional), is the assigned tag name of the to be built docker image. Default is ``local-<device>``.
+3. **device** (optional), is the intended device for the docker image. Default is ``gpu``, can also build a ``cpu`` image.
```
classicsong, line 131:
```suggestion
4. **device** (optional), is the intended device for the docker image. Default is ``gpu``, can also build a ``cpu`` image.
```

docker/build_docker_sagemaker.sh
```diff
@@ -39,7 +39,7 @@ cp -r "${GSF_HOME}/sagemaker" code/graphstorm/sagemaker
 cp -r "${GSF_HOME}/docker/sagemaker/build_artifacts" build_artifacts
 
 # Build OSS docker for EC2 instances that an pull ECR docker images
-DOCKER_FULLNAME="${IMAGE_NAME}:${TAG}"
+DOCKER_FULLNAME="${IMAGE_NAME}:${TAG}-${IMAGE_TYPE}"
```
classicsong, line 42: DEVICE_TYPE?

thvasilo: We're using IMAGE_TYPE right now see https://github.com/awslabs/graphstorm/blob/b0a8c7750896c8aa0866be2318c25fabbf850b48/docker/build_docker_sagemaker.sh#L13-L18

Agree that DEVICE_TYPE is a better name, I can change it to that if you think we should.

docker/Dockerfile.local
thvasilo: I tried to repro this inside a CPU Docker container but haven't encountered this error. Do you know what yuo particular execution environment was?

jalencato: I think you can try that in any GPU docker container and you will meet that problem.

thvasilo: Ah only happens in GPU envs, I got the error when running on a G5 with the GPU image.

thvasilo:
Moved back to using DGL 1.1.3 and updated the Dockerfile structure to use a common source.

Could you take another look @classicsong @jalencato ? Tested on single instance and multi-instance training, with the repro example that @jalencato developed for the DGL issue.

docker/Dockerfile.local
```diff
@@ -1,34 +1,76 @@
...
-RUN apt-get install -y cython3 libicu-dev
-RUN pip3 install h5py psutil
+FROM base as base-gpu
 
-RUN apt-get install -y unzip
+# Install torch, DGL, and GSF deps that require torch
+RUN pip install \
+    dgl==${DGL_VERSION}+cu121 \
+    ogb==${OGB_VERSION} \
+    torch==${TORCH_VERSION} \
```
jalencato, line 61: Why do we not specify the torch version with CUDA version any more here? I see we are doing so for dgl library installation. It is a good practice to keep them same.

docker/Dockerfile.local
```diff
@@ -1,34 +1,76 @@
...
+    -f https://data.dgl.ai/wheels/cu121/repo.html \
+    && rm -rf /root/.cache
+
+FROM base-${DEVICE} as runtime
+
+ENV PYTHONPATH="/root/dgl/tools/:${PYTHONPATH}"
 
 # Download DGL source code
-RUN cd /root; git clone --recursive https://github.com/dmlc/dgl.git
+RUN cd /root; git clone --single-branch https://github.com/dmlc/dgl.git
```
jalencato, line 71: We can add ```ENV PYTHONPATH="/root/dgl/tools:$PYTHONPATH"``` as well here.

jalencato (commented):
Some minor comments. Overall looks good to me.

docker/Dockerfile.local
thvasilo: I don't think it's possible to specify CUDA 12.1 right now for PT, however it's the default. See https://pytorch.org/get-started/locally/

```
ERROR: Could not find a version that satisfies the requirement torch==2.1.2+cu121 (from versions: 1.7.1, 1.8.0, 1.8.1, 1.9.0, 1.9.1, 1.10.0, 1.10.1, 1.10.2, 1.11.0, 1.12.0, 1.12.1, 1.13.0, 1.13.1, 2.0.0, 2.0.1, 2.1.0, 2.1.1, 2.1.2, 2.2.0, 2.2.1, 2.2.2)
#8 5.320 ERROR: No matching distribution found for torch==2.1.2+cu121
```

For 11.8 we would need to use `pip3 install torch torchvision torchaudio --index-url https://download.pytorch.org/whl/cu118`

docker/Dockerfile.local
thvasilo: It's in line 68.

docker/Dockerfile.local
jalencato: I see. Then we can leave if for now.

docker/build_docker_sagemaker.sh
classicsong: Let's use DEVICE_TYPE

classicsong (approved):
LGTM

jalencato (approved):
LGTM DGL said they have fixed the bug and they will include that change in 2.2 version. We can move to 2.2 later in May

PR #810, opened 2024-04-23
[Enhancement] Remove GSF dependency on lint check CI

jalencato:
*Issue #, if available:*

*Description of changes:*

Remove the part for installing the graphstorm when running lint test.


thvasilo (approved):
If we tested this and the check still works LGTM

Issue #789, opened 2024-03-29
[Feature Request] Support Multi-task learning


classicsong:
GraphStorm should support multi-task learning as its built-in capability. 
* Unsupervised multi-task learning: Other than using link prediction as the sole supervision signal for unsupervised learning, one can add other supervision signals like reconstruct node features, predicting the motifs of specific nodes, etc (https://arxiv.org/abs/2007.10445). 
* Entities with different node types or edges from different edge types of a graph can have different tasks associated with them. There are also cases when the same node types or edge types have multiple tasks associated with them.

PR #825, opened 2024-05-06
Refactor gsf.py to make the code of creating decoders and creating link prediction dataloders reusable.

classicsong:
*Issue #, if available:*
Related to #789 

*Description of changes:*
1. decouple the code of creating decoders from the code of creating built-in models to make the code of creating decoders more reusable. 
2. add `get_builtin_lp_train_dataloader_class` and `get_builtin_lp_eval_dataloader_class` in gsf.py

### DIFF
--- a/python/graphstorm/gsf.py
+++ b/python/graphstorm/gsf.py
@@ -312,7 +312,7 @@ def create_builtin_node_model(g, config, train_task):
     encoder_out_dims = model.gnn_encoder.out_dims \
         if model.gnn_encoder is not None \
             else model.node_input_encoder.out_dims
-    decoder, loss_func = create_builtin_node_decoder(g, encoder_out_dims, config)
+    decoder, loss_func = create_builtin_node_decoder(g, encoder_out_dims, config, train_task)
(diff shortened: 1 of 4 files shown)


python/graphstorm/gsf.py
```diff
@@ -209,61 +234,88 @@ def create_builtin_node_gnn_model(g, config, train_task):
...
-    """ Create a built-in model for node prediction.
+# pylint: disable=unused-argument
+def create_builtin_node_decoder(g, decoder_input_dim, config, train_task):
+    """ create builtin node decoder according to task config
 
     Parameters
     ----------
     g: DGLGraph
-        The graph used in training and testing
+        The graph data.
```
zhjwy9343, line 244: Since this method does not use the `g` argument, should we keep it? 
Use `g` as an input argument is really odd.

zhjwy9343 (approved):
LGTM

python/graphstorm/gsf.py
classicsong: Let's keep it and replace it with g_metainfo in the future.

PR #832, opened 2024-05-13
[Docker] Move the Amazon ECR Public Gallery auth to only the CPU branch of the local build script.

thvasilo:
*Issue #, if available:*

*Description of changes:*
* To pull the CPU source image we need to authenticate with Amazon Public ECR gallery using the AWS cli. This is not needed to build the GPU image, so we move that auth step under the CPU branch, to avoid customers having to add the permission to their roles if they're only building the GPU image.


jalencato (approved):
LGTM

PR #839, opened 2024-05-15
[Bug] Fix error of reconstruct embed encoder when use_mini_batch_infer=true

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
This PR fixed two bugs:
- Missing `use_mini_batch_infer` argument when calling the `eval()` method in `lp_trainer`;
- `GNNEncoderWithReconstructedEmbed` does not support mini-batch inference because of the mis-matching of number of blocks. Add an assertion to check the setting of `use_mini_batch_infer` for GNNEncoderWithReconstructedEmbed. And changed the end2end test command to avoid this case.


classicsong (approved):
LGTM

PR #840, opened 2024-05-15
[Fix] Fix the error on prospector

jalencato:
*Issue #, if available:*

*Description of changes:*

Prospector dependency always has lots of problem. Let us move to pylint with same version as the gsp_lint pipeline does.


zhjwy9343 (approved):
lgtm

Issue #835, opened 2024-05-13
[Call For Discussion] Shall we set the default value of remove_target_edge_type to False?


classicsong:
Originally, we set remove_target_edge_type to True. So in edge prediction tasks (classification and regression) the target edge type is not included in message passing by default. But, is this the right practice?

Related code: https://github.com/awslabs/graphstorm/blob/ea5e5848dd0619ed8205730e11ab389a63212720/python/graphstorm/config/argument.py#L1695-L1709

zheng-da:
if the edge classification is to predict the existence of an edge between two nodes, we should remove the target edge in the message passing to avoid information leak.
if it's to predict some attributes associated with an edge, it may depend on.
since we don't know what to predict, to be safe, we should remove the target edge in message passing by default.

classicsong:
I see what you mean. Make sense.

PR #845, opened 2024-05-16
Update docstring of remove_target_edge_type

classicsong:
*Issue #, if available:*
#835 

*Description of changes:*


python/graphstorm/config/argument.py
```diff
@@ -1706,6 +1713,10 @@ def remove_target_edge_type(self):
 
         # By default, remove training target etype during
         # message passing to avoid information leakage
+        logging.warning("remove_target_edge_type is set to True by default. "
+                        "If your edge classification task is not predicting "
+                        "the existence of the target edge, we suggest you to "
+                        "set it to False.")
```
zheng-da, line 1719: this will cause many printings in distributed setting. if this is called in a loop, we will see a lot of warning messages. do you think if this is necessary?

classicsong: I think it is necessary.
My understanding is that most of edge tasks are not the task of predicting the existence of the target edge, so the target edge type is useful in most of the cases. If a user does not set remove_target_edge_type to False, he/she should be notified.

zheng-da: should the default value be set for the most common use case?

zheng-da: I'm not sure what the best solution is. I think the modification is fine with me.

zheng-da (approved):

PR #850, opened 2024-05-17
[BugFix] Fix node class full graph infer

classicsong:
*Issue #, if available:*

*Description of changes:*
Originally, node_mini_batch_predict expect its argument `loader` is a node dataloader with 0 GNN layers (fanout=[]). However, this can not been guaranteed.

Change the implementation to make it work with node dataloader with/without GNN layers.

### DIFF
--- a/python/graphstorm/model/node_gnn.py
+++ b/python/graphstorm/model/node_gnn.py
@@ -323,7 +323,7 @@ def node_mini_batch_predict(model, emb, loader, return_proba=True, return_label=
     # TODO(zhengda) I need to check if the data loader only returns target nodes.
     model.eval()
     with th.no_grad():
-        for _, seeds, _ in loader:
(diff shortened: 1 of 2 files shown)


python/graphstorm/model/node_gnn.py
```diff
@@ -323,17 +323,17 @@ def node_mini_batch_predict(model, emb, loader, return_proba=True, return_label=
     # TODO(zhengda) I need to check if the data loader only returns target nodes.
     model.eval()
     with th.no_grad():
-        for input_nodes, seeds, _ in loader:
-            for ntype, in_nodes in input_nodes.items():
+        for _, seeds, _ in loader:
```
zhjwy9343, line 326: Suggest to add a comment here, explaining why use seeds, not input_nodes.

zhjwy9343 (approved):
LGTM. Suggest to add a comment line to explain this corner case.

Issue #796, opened 2024-04-10
[Documentation] Add doc about saving configuration files in gconstruct and gsprocessing.


classicsong:
Add doc about saving configuration files when processing graph input data (e.g. --output-conf-file). The saved configuration file will record how data are transformed (numerical features, categorical features, etc.).

For example:
When constructing a graph for training, we can save the configuration file recording the data transformation information by adding `--output-conf-file data_transform_new.conf`
```
python3 -m graphstorm.gconstruct.construct_graph --conf-file data_transform.conf --num-processes 2 --output-dir /tmp/test_out --graph-name test --output-format DGL --output-conf-file data_transform_new.conf
```

Then during inference, we can load the `data_transform_new.conf` for creating inference graphs.

zhjwy9343:
A better user experience would be automatically save the configuration file. If users specify the `--output-conf-file`, save the data transformation configuration file to the given location. If not, use the default `--output-dir` and save the file with name `data_transform_new.json`.

PR #853, opened 2024-05-22
[Enhancement] Change save new gconstruct configuration file behavior

zhjwy9343:
*Issue #796 , if available:*

*Description of changes:*
This PR changed the behavior of saving the new gconstruct configuration file. If users do not specify the `--output-conf-file` argument, we will automatically save it to the `output-dir` folder with the name "data_transform_new.json".
This PR also update the Configuration documents, adding explanations of the `--output-conf-file` argument.

### DIFF
--- a/python/graphstorm/gconstruct/construct_graph.py
+++ b/python/graphstorm/gconstruct/construct_graph.py
@@ -770,6 +770,9 @@ def process_graph(args):
                           skip_nonexist_edges=args.skip_nonexist_edges)
     sys_tracker.check('Process the edge data')
     num_nodes = {ntype: len(raw_node_id_maps[ntype]) for ntype in raw_node_id_maps}
+
+    os.makedirs(args.output_dir, exist_ok=True)
+
     if args.output_conf_file is not None:
         outfile_path = args.output_conf_file
     else:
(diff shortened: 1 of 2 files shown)


python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -770,10 +770,17 @@ def process_graph(args):
...
+
+    os.makedirs(args.output_dir, exist_ok=True)
+
     if args.output_conf_file is not None:
-        # Save the new config file.
-        with open(args.output_conf_file, "w", encoding="utf8") as outfile:
-            json.dump(process_confs, outfile, indent=4)
+        outfile_path = args.output_conf_file
+    else:
+        outfile_path = os.path.join(args.output_dir, 'data_transform_new.json')
```
classicsong, line 779: Do we need to check whether 'data_transform_new.json' is already there? 
Are we going to overwrite it?

classicsong (approved):

python/graphstorm/gconstruct/construct_graph.py
zhjwy9343: If users do not provide this argument, we generate it by default. So the default behavior is to overwrite the existing one.

We can check the existence of the file, but the next operation is complex. Options are
- rename the existing one, and the problem is how to rename the existing one.
- rename the new file, and the problem is same, i.e., how to rename the new file.

If we follow some rules to rename files, we need to check if renamed files exist too, like MacOS does. But Python os library does not have this function.

Another potential issue might be the permission to rename existing files. We may confront some error when we cannot rename existing files.

So I would vote for overwriting the existing file.

zhjwy9343: Add new warning message when overwriting happens.

PR #857, opened 2024-05-28
[GSProcessing] Add structure for saving transformation JSON files.

thvasilo:
## Description of changes:

* This commit only adds saving the transformations in a JSON representation.
* Loading and applying the pre-computed transformation will come in an upcoming PR.
* First implemented for categorical transformation.

Detailed changes:

To support the saving and loading of pre-computed transformations, we use the following design.

* A new JSON output is created in the output that includes JSON representations of the transformations
  for which we have implemented a representation.
* The base class `DistributedTransformation` now takes an _optional_ SparkSession and and _optional_ `json_representation` `dict` during initialization.
* We add a `get_json_representation()` function to `DistributedTransformation` which returns the JSON representation of the transformation if it's not None, or an empty dict otherwise. The assumed contract is that each individual transformation must populate its `json_representation` dict during the call to `apply()` which applies the transformation for the first time.
* The `DistFeatureTransformer` now also takes a SparkSession and a json_representation `dict` in its constructor.
  The Spark session is currently only passed to the constructor of `DistCategoryTransformation`. We need it to
  create Spark DataFrames during `apply()`.
* The `apply_transformation`function of `DistFeatureTransformer` now returns a tuple, `(processed_df: DataFrame, json_representation: dict)`. The JSON representation is retrieved using the `get_json_representation()` function of
  each distributed transformation implementation. Currently only the categorical transformation will return a non-empty dict.
* We add a `transformation_representations` dict as a member of `DistHeterogeneousGraphLoader`. These dict is used to gather the representations of each feature as we iterate through them. The structure is:

```python
{
  "node_features": {
    "node_type1": {
      "feature_name1": {
         # feature representation goes here
      }
    },
    "node_type2": {...}
  },
  "edges_features": {...}
}
```
* At the end of graph loading this dict is saved to storage under the output prefix as `precomputed_transformations.json`. This file will be used to re-construct the feature transformations in an upcoming PR.

Particularly for the `DistCategoryTransformation`:

* We choose to save its representation as a dict with the following structure:

```
string_indexer_labels_array:
    tuple[tuple[str]], outer tuple has num_cols elements,
    each inner tuple has num_cats elements, each str is a category string.
    Spark uses this to represent the one-hot index for each category, its
    position in the inner tuple is the one-hot-index position for the string.
    Categories are sorted by their frequency in the data.
cols:
    list[str], with num_cols elements, each item is one column name that was used in the 
    original transformation.
per_col_label_to_one_hot_idx: 
    dict[str, dict[str, int]], with num_cols elements, each with num_categories elements, 
    a more verbose mapping from column name to dict of string to one-hot-index position.
transformation_name:
    str, will be 'DistCategoryTransformation'
``` 

The string_indexer_labels_array comes from Spark's own representation of its StringIndexer class, encapsulated in the `labelsArray` var of StringIndexerModel. See docs here: https://spark.apache.org/docs/3.1.1/api/python/reference/api/pyspark.ml.feature.StringIndexerModel.html#pyspark.ml.feature.StringIndexerModel.labelsArray and https://spark.apache.org/docs/3.1.1/api/python/reference/api/pyspark.ml.feature.StringIndexer.html#pyspark.ml.feature.StringIndexer

Example representation for input data:

```
state
wa
ca
wa
# no string here, represents missing value
ny
```

```json
{
    "node_features": {
        "user": {
            "state": {
                "transformation_name": "DistCategoryTransformation",
                "string_indexer_labels_arrays": [["wa", "ca", "ny"]],
                "cols": ["state"],
                "per_col_label_to_one_hot_idx": {
                    "state": {
                        "wa": 0,
                        "ca": 1,
                        "ny": 2
                    }
                }
            }
        }
    },
    "edge_features": {}
}
```

To reconstruct the transformation on a new DF we will iterate over the `cols`, and for each, create one-hot vectors according the position of each string in the corresponding `labels_array`. 

E.g. given a labels array `["string2", "string1", "string3"]`, when in the input data we encounter`string2`, it will be transformed to `[1, 0, 0]`, since its position in the labels array is position 0. `string1` will have the representation `[0, 1, 0]` etc.

All the code changes in the apply function of `DistCategoryTransformation` are meant to build up this representation.


graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_category_transformation.py
```diff
@@ -105,8 +113,77 @@ def apply(self, input_df: DataFrame) -> DataFrame:
...
+        # see get_json_representation() docstring for structure
+        self.json_representation = {
+            "string_indexer_labels_array": str_indexer_model.labelsArray,
+            "cols": self.cols,
+            "per_col_label_to_one_hot_idx": per_col_label_to_one_hot_idx,
+        }
+
         return dense_vector_features
 
+    def get_json_representation(self) -> dict:
```
jalencato, line 167: I think we have already had one in the base class, what is the new one for?

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -64,68 +66,90 @@
...
-        Local path to where the output metadata files will be created.
+    add_reverse_edges : bool
+        Whether to add reverse edges to the graph.
     data_configs : Dict[str, Sequence[StructureConfig]]
         Dictionary of node and edge configurations objects.
+    enable_assertions : bool, optional
+        When true enables sanity checks for the output created.
+        However these are costly to compute, so we disable them by default.
+    graph_name: str
+        The name of the graph we will process.
```
jalencato, line 82:
```suggestion
        Name of the graph.
```

jalencato (commented):
Leave some comments. 

I see we have modified some functions in the DistCategoryTransformation, but the description of these changes are not detailed. May we add some description about this change?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_category_transformation.py
thvasilo: I wanted to add this to have a specific docstring describing the representation. There's ways to do that without overloading the function, but they're all uglier with re-assigning the value of the `__doc__` for the function during init.

This should be cleaner

graphstorm-processing/graphstorm_processing/data_transformations/dist_feature_transformer.py
```diff
@@ -37,13 +37,19 @@ class DistFeatureTransformer(object):
...
 
-    def __init__(self, feature_config: FeatureConfig):
+    def __init__(
+        self, feature_config: FeatureConfig, spark: SparkSession, json_representation: dict
+    ):
         feat_type = feature_config.feat_type
         feat_name = feature_config.feat_name
         args_dict = feature_config.transformation_kwargs
         self.transformation: DistributedTransformation
+        # TODO: We will use this to re-apply transformations
```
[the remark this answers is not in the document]
thvasilo: When loading the JSON representation from storage to re-apply the same, pre-computed transformation on future data, the dict here will hold the representation read from the JSON file.

jalencato (approved):
LGTM

classicsong (commented):
Can we save the transformation info back to the input config?

graphstorm-processing/graphstorm_processing/constants.py
```diff
@@ -85,3 +85,6 @@ class FilesystemType(Enum):
     "3.4": "3.3.4",
     "3.3": "3.3.2",
 }
+
+########## Precomputed transformations ################
+TRANSFORMATIONS_FILENAME = "precomputed_transformations.json"
```
classicsong, line 90: Is this a separate file from the construct json file?

graphstorm-processing/graphstorm_processing/data_transformations/dist_feature_transformer.py
```diff
@@ -56,7 +62,9 @@ def __init__(self, feature_config: FeatureConfig):
         elif feat_type == "bucket-numerical":
             self.transformation = DistBucketNumericalTransformation(**default_kwargs, **args_dict)
         elif feat_type == "categorical":
-            self.transformation = DistCategoryTransformation(**default_kwargs, **args_dict)
+            self.transformation = DistCategoryTransformation(
+                **default_kwargs, **args_dict, spark=spark
```
classicsong, line 66: Why we need spark session here, but for others we dont.

thvasilo: To save the transformation representation, we need to be able to create Spark DataFrames and we need a SparkContext to do so. So far the other transformations do not need one.

graphstorm-processing/graphstorm_processing/distributed_executor.py
```diff
@@ -350,18 +357,136 @@ def run(self) -> None:
...
+            structure_type: str,
+        ):
+            """Appends the pre-computed transformations to the input dicts."""
+            assert structure_type in ["edge", "node"]
+            for input_dict in structure_input_dicts:
+                type_name = get_structure_type(input_dict, structure_type)
+                # If we have pre-computed transformations for this type
+                if type_name in structure_transformations:
+                    type_transformations: Mapping[str, Mapping] = structure_transformations[
+                        type_name
```
classicsong, line 459: The indent is weird.

graphstorm-processing/graphstorm_processing/distributed_executor.py
```diff
@@ -350,18 +357,136 @@ def run(self) -> None:
...
+            structure_transformations: Mapping[str, Mapping[str, Mapping]],
+            structure_type: str,
+        ):
+            """Appends the pre-computed transformations to the input dicts."""
+            assert structure_type in ["edge", "node"]
+            for input_dict in structure_input_dicts:
+                type_name = get_structure_type(input_dict, structure_type)
+                # If we have pre-computed transformations for this type
+                if type_name in structure_transformations:
+                    type_transformations: Mapping[str, Mapping] = structure_transformations[
```
classicsong, line 458: type_transformations is feature_name -> transformation_information?
Will all the transformation information be a Mapping?

graphstorm-processing/graphstorm_processing/distributed_executor.py
```diff
@@ -350,18 +357,136 @@ def run(self) -> None:
...
+                    # and append representation if one exists
+                    for type_feat_dict in input_dict["features"]:
+                        feat_name = (
+                            type_feat_dict["name"]
+                            if "name" in type_feat_dict
+                            else type_feat_dict["column"]
+                        )
+                        if feat_name in type_transformations:
+                            type_feat_dict["precomputed_transformation"] = type_transformations[
+                                feat_name
```
classicsong, line 475: The indent is weird.

graphstorm-processing/graphstorm_processing/distributed_executor.py
```diff
@@ -503,25 +628,27 @@ def main():
...
-        local_output_path = "/opt/ml/processing/output"
+        local_metadata_output_path = "/opt/ml/processing/output"
     else:
         if filesystem_type == FilesystemType.LOCAL:
-            local_output_path = gsprocessing_args.output_prefix
+            local_metadata_output_path = gsprocessing_args.output_prefix
         else:
             # Only needed for local execution with S3 data
-            local_output_path = tempdir.name
+            local_metadata_output_path = tempdir.name
```
classicsong, line 637: Why change the name?

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -98,18 +98,50 @@ class HeterogeneousLoaderConfig:
...
+
+        For chunked graph format see
+        https://docs.dgl.ai/guide/distributed-preprocessing.html#specification
+    transformation_representations : dict
+        A dictionary containing the processed graph transformations.
+    timers : dict
+        A dictionary containing the timing information for different steps of processing.
+    """
+
+    processed_graph_metadata_dict: dict
```
classicsong, line 139: Previously, the only output is processed_graph_metadata_dict?
But now we have transformation_representations and timers?

graphstorm-processing/tests/test_dist_executor.py
```diff
@@ -0,0 +1,79 @@
...
+        },
+        "edge_features": {},
+    }
+
+    input_config_with_transforms = dist_executor._merge_config_with_transformations(
+        dist_executor.gsp_config_dict,
+        pre_comp_transormations,
+    )
+
+    # Ensure the "user" node type's "age" feature includes a transformation entry
```
classicsong, line 74: Does "age" feature has a transformation entry in is original config?
What if the "transformation" has other information like "normalizer": "min-max"?

thvasilo: Yes, here we're just ensuring the `_merge_config_with_transformations` function works as intended, not actually checking the correctness of the transformation.

The contract we have for a transformation representation in the JSON file is that it always has one "transformation" key (or maybe we should call it "transformation_name") and the rest of the keys allow us to reconstruct the transformation.

For a numerical feature yes, we might have other keys like min_val/max_val, but this test would still apply (since we hardcode the representation).

In the follow-up PR we actually test the produced JSON representations and re-apply them to the data to ensure they match the original.

graphstorm-processing/graphstorm_processing/distributed_executor.py
thvasilo: All code formatting is done automatically by `black`. Here it looks weird because of the long variable names, I'll shorten them to allow better formatting.

graphstorm-processing/graphstorm_processing/distributed_executor.py
thvasilo: Correct, each transformation is `"feature_name" -> dict`. We use Mapping instead of dict because it helps with type annotation, and communicates that we don't expect to modify the mapping. In cases where we expect to modify the contents we use `dict`.

graphstorm-processing/graphstorm_processing/distributed_executor.py
thvasilo: We also have a member var named `self.output_prefix` which is where we put the processed data, and it can be S3 or local. 

We rename this to make it clear that this is a local path intended for metadata files only

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: The function started with no return value, then with the processed_graph_metadata_dict, then added the timers, now we are adding a third dict.

I thought it's better to encapsulate everything in a class so 
1. We don't keep modifying the return type of the function
2. It's more clear what each item is, e.g. instead of return_val[1] we say  return_val.timers which clearly communicates what is inside the returned object.

graphstorm-processing/graphstorm_processing/distributed_executor.py
thvasilo: Answered above, formatting is automatic, we shortened the var names to make it a bit better.

graphstorm-processing/tests/test_dist_executor.py
thvasilo: I changed this test to use the categorical transformation to ensure we're using a feature for which we support transformations.

We test the actual saved values in `graphstorm-processing/tests/test_dist_category_transformation:test_multiple_single_cat_cols_json`

graphstorm-processing/graphstorm_processing/distributed_executor.py
classicsong: Can you add some comments about the "feature_name" -> Mapping ?
For example ` "feature_name" is the feature name and the Mapping records the transformation related meta information such as xxx`

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
classicsong: That makes sense

classicsong (approved):
One comment.
Others LGTM.

PR #861, opened 2024-06-02
Support multi-task inference.

classicsong:
*Issue #, if available:*
#789 

*Description of changes:*
This PR adds inference support for multi-task learning. Users can use `python3 -m graphstorm.run.gs_multi_task_learning --inference ` to launch a inference task.

This PR also changes remap_result.py to support remapping prediction results from multi-task learning inference. (The prediction results of each task are stored separately on different folders with the name of the corresponding task id.)


python/graphstorm/model/gnn_encoder_base.py
```diff
@@ -416,6 +416,6 @@ def get_input_embeds1(input_nodes, node_feats):
                 get_input_embeds = partial(get_input_embeds1, node_feats=next_layer_input)
             next_layer_input = dist_inference_one_layer(str(i), g, dataloader,
                                                         list(infer_nodes.keys()),
-                                                        layer, get_input_embeds, device,
+                                                        layer, get_input_embeds,device,
```
zhjwy9343, line 419: Better not to remove the white space.

python/graphstorm/run/gsgnn_mt/mt_infer_gnn.py
```diff
@@ -0,0 +1,275 @@
...
+                                    GSgnnEdgeDataLoader,
+                                    GSgnnMultiTaskDataLoader)
+from graphstorm.dataloading import (GSgnnLinkPredictionTestDataLoader,
+                                    GSgnnLinkPredictionJointTestDataLoader,
+                                    GSgnnLinkPredictionPredefinedTestDataLoader)
+from graphstorm.dataloading import BUILTIN_LP_UNIFORM_NEG_SAMPLER
+from graphstorm.dataloading import BUILTIN_LP_JOINT_NEG_SAMPLER
+
+from graphstorm.model.multitask_gnn import GSgnnMultiTaskSharedEncoderModel
+from graphstorm.inference import GSgnnMultiTaskLearningInferer
```
zhjwy9343, line 38: If changed above, should change to `GSgnnMultiTaskLearningInferrer` too.

python/graphstorm/gsf.py
```diff
@@ -909,3 +907,78 @@ def create_task_decoder(task_info, g, decoder_input_dim, train_task):
...
+                                            config.early_stop_strategy)
+    elif task_info.task_type in [BUILTIN_TASK_NODE_REGRESSION]:
+        return GSgnnRegressionEvaluator(config.eval_frequency,
+                                        config.eval_metric,
+                                        config.use_early_stop,
+                                        config.early_stop_burnin_rounds,
+                                        config.early_stop_rounds,
+                                        config.early_stop_strategy)
+    elif task_info.task_type in [BUILTIN_TASK_EDGE_CLASSIFICATION]:
+        return GSgnnClassificationEvaluator(config.eval_frequency,
```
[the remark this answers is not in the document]
classicsong: That's different.
For node classification, originally, we support multiple node tasks on different node types. That's why we allow multilabel to be a dict as ntype->multilabel_flag. So, L925-L926 handles multiple-labels.
For edge classification, we do not support multiple edge tasks on different edge types. So we do not need to handle it.

But I think here as task_info is a TaskInfo from multi-task learning, we should assume multilabel is a bool.

python/graphstorm/run/gsgnn_mt/mt_infer_gnn.py
```diff
@@ -0,0 +1,275 @@
...
+from graphstorm.dataloading import BUILTIN_LP_JOINT_NEG_SAMPLER
+
+from graphstorm.model.multitask_gnn import GSgnnMultiTaskSharedEncoderModel
+from graphstorm.inference import GSgnnMultiTaskLearningInferrer
+
+from graphstorm.utils import get_device, use_wholegraph
+from graphstorm.utils import get_lm_ntypes
+from graphstorm.eval import GSgnnMultiTaskEvaluator
+
+def create_task_infer_dataloader(task, config, infer_data):
```
zhjwy9343, line 44: Should we move this function into the `gsf.py`?

zhjwy9343 (commented):
Just finished all reviews except for test. Let's see the test results filrst.

python/graphstorm/run/gsgnn_mt/mt_infer_gnn.py
classicsong: This is only used by infer_gnn.py. So I would like to keep it here now.

zhjwy9343: Alright, might move it once we see other functions use it in the future.

zhjwy9343 (approved):
LGTM

PR #864, opened 2024-06-04
Relax the train/val/test mask check for multi-task learning

classicsong:
*Issue #, if available:*

*Description of changes:*
The original train/val/test mask fields check is too strict. As there can be some exception when users want to provide masks through node features or some tasks are sharing the same mask.

Relax the check and turn the assertion into warning.


jalencato (approved):
LGTM

PR #866, opened 2024-06-06
[Enhancement] Allow detach on the local docker container

jalencato:
*Issue #, if available:*

*Description of changes:*

* Change the default option for the local container to detach the container so local user can start local job without rewrite the command


classicsong:
Can you give a description of this PR?

thvasilo (approved):
LGTM, but we'll need to ensure we're still able to use fetch and run, which I think should be fine by overriding the CMD at runtime

classicsong:
Can we merge this PR?

PR #870, opened 2024-06-10
[GSProcessing] Add pre-computed categorical transformation loading

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Follow-up to #857 
* Allow us to re-apply a previously saved categorical transformation to new data. See below for design details.

To be able to re-apply the categorical transformations that we create using the code in #857 , we first create a mapping from original string to one-hot representation, that we read from the saved JSON file, then use a UDF to use the mapping(s) on the column(s).

The `DistributedTransformation` class from which all transformation implementations inherit, gains a new function, `apply_precomputed_transformation`. When a pre-computed transformation JSON file exists in the input, and the feature is one of those listed in that file, we use this function to re-apply the existing transformation instead of creating a new one.

The default implementation for `apply_precomputed_transformation` is to log a warning and apply a new transformation. 

When we implement a pre-computed transform for a new transformation (e.g. numerical) we need to:
* Ensure the the transformation's `self.json_representation` is populated during the call to `apply()`. This ensures the transformation info will be saved in the output JSON.
* Override the `apply_precomputed_transformation` function (as we did for `DistCategoryTransformation` here), so that it uses the dict loaded from the JSON file to re-apply the transformation to the new data.

### DIFF
--- a/docs/source/gs-processing/usage/example.rst
+++ b/docs/source/gs-processing/usage/example.rst
@@ -173,7 +173,7 @@ we can use the following command to run the processing job locally:
 
 .. code-block:: bash
 
-    gs-processing --config-filename gconstruct-config.json \
+    gs-processing --config-filename gsprocessing-config.json \
         --input-prefix ./tests/resources/small_heterogeneous_graph \
         --output-prefix /tmp/gsprocessing-example/ \
         --do-repartition True
@@ -211,26 +211,41 @@ and can be used downstream to create a partitioned graph.
 .. code-block:: bash
 
     $ cd /tmp/gsprocessing-example
-    $ ls
+    $ ls -l
 
-    edges/  launch_arguments.json  metadata.json  node_data/
-    raw_id_mappings/  perf_counters.json  updated_row_counts_metadata.json
+    edges/
+    gconstruct-config_with_transformations.json
+    launch_arguments.json
+    metadata.json
+    node_data/
+    perf_counters.json
+    precomputed_transformations.json
+    raw_id_mappings/
+    updated_row_counts_metadata.json
 
 We have a few JSON files and the data directories containing
 the graph structure, features, and labels. In more detail:
 
+* ``gsprocessing-config_with_transformations.json``: This is the input configuration
+  we used, modified to include representations of any supported transformations
+  we applied. This file can be used to re-apply the transformations on new data.
 * ``launch_arguments.json``: Contains the arguments that were used
   to launch the processing job, allowing you to check the parameters after the
   job finishes.
-* ``updated_row_counts_metadata.json``:
-  This file is meant to be used as the input configuration for the
-  distributed partitioning pipeline. ``gs-repartition`` produces
-  this file using the original ``metadata.json`` file as input.
 * ``metadata.json``: Created by ``gs-processing`` and used as input
   for ``gs-repartition``, can be removed the ``gs-repartition`` step.
 * ``perf_counters.json``: A JSON file that contains runtime measurements
   for the various components of GSProcessing. Can be used to profile the
   application and discover bottlenecks.
+* ``precomputed_transformations.json``: A JSON file that contains representations
+  of supported transformations. This file can be copied to the input path of another
+  set of raw files, and GSProcessing will use the transformation values listed here
+  instead of creating new ones. Use this to re-apply the same transformations to new
+  data.
+* ``updated_row_counts_metadata.json``:
+  This file is meant to be used as the input configuration for the
+  distributed partitioning pipeline. ``gs-repartition`` produces
+  this file using the original ``metadata.json`` file as input.
 
 The directories created contain:
 
--- a/docs/source/gs-processing/usage/example.rst
+++ b/docs/source/gs-processing/usage/example.rst
@@ -214,7 +214,7 @@ and can be used downstream to create a partitioned graph.
     $ ls -l
 
     edges/
-    gconstruct-config_with_transformations.json
+    gsprocessing-config_with_transformations.json
     launch_arguments.json
     metadata.json
     node_data/
--- a/docs/source/gs-processing/usage/example.rst
+++ b/docs/source/gs-processing/usage/example.rst
@@ -238,10 +238,13 @@ the graph structure, features, and labels. In more detail:
   for the various components of GSProcessing. Can be used to profile the
   application and discover bottlenecks.
 * ``precomputed_transformations.json``: A JSON file that contains representations
-  of supported transformations. This file can be copied to the input path of another
-  set of raw files, and GSProcessing will use the transformation values listed here
-  instead of creating new ones. Use this to re-apply the same transformations to new
-  data.
+  of supported transformations. To re-use these transformations on another dataset,
+  place this file in the top level of another set of raw data, at the same level
+  as the input GSProcessing/GConstruct configuration JSON file.
+  GSProcessing will use the transformation values listed here
+  instead of creating new ones, ensuring that models trained with the original
+  data can still be used in the newly transformed data. Currently only
+  categorical transformations can be re-applied.
 * ``updated_row_counts_metadata.json``:
   This file is meant to be used as the input configuration for the
   distributed partitioning pipeline. ``gs-repartition`` produces
--- a/graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/base_dist_transformation.py
+++ b/graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/base_dist_transformation.py
@@ -16,6 +16,7 @@ limitations under the License.
 
 from abc import ABC, abstractmethod
 from typing import Optional, Sequence
+import logging
 
 from pyspark.sql import DataFrame, SparkSession
 
@@ -23,6 +24,15 @@ from pyspark.sql import DataFrame, SparkSession
 class DistributedTransformation(ABC):
     """
     Base class for all distributed transformations.
+
+    Parameters
+    ----------
+    cols : Sequence[str]
+        Column names to which we will apply the transformation
+    spark : Optional[SparkSession], optional
+        Optional SparkSession if needed by the underlying implementation, by default None
+    json_representation : Optional[dict], optional
+        Pre-computed transformation representation to use, by default None
     """
 
     def __init__(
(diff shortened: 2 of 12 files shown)


graphstorm-processing/graphstorm_processing/distributed_executor.py
thvasilo, line 1: Changes to this file are mostly cosmetic, e.g. we moved the initialization of the `DistHeterogeneousGraphLoader` object from the `run` function to the DistributedExecutor's __init__, ensuring a `loader` class variable exists the moment the `DistributedExecutor` object is created.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -1029,16 +1036,26 @@ def _process_node_features(
...
         node_type_feature_metadata = {}
         ntype_feat_sizes = {}  # type: Dict[str, int]
+
         for feat_conf in feature_configs:
-            json_representation = self.transformation_representations.get(feat_conf.feat_name, {})
+            json_representation = (
+                self.pre_computed_transformations.get("node_features", {})
+                .get(node_type, {})
+                .get(feat_conf.feat_name, {})
+            )
```
thvasilo, line 1045: This will get a value if there exists a pre-computed representation for this feature name and node type, None otherwise. We do the same for the edges.

graphstorm-processing/pyproject.toml
```diff
@@ -40,6 +40,7 @@ black = "~24.2.0"
 pre-commit = "^3.3.3"
 types-mock = "^5.1.0.1"
 pylint = "~2.17.5"
+diff-cover = "^9.0.0"
```
thvasilo, line 43: I use this to determine how test coverage changes between commits.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/base_dist_transformation.py
```diff
@@ -52,6 +62,28 @@ def get_json_representation(self) -> dict:
         else:
             return {}
 
+    def apply_precomputed_transformation(self, input_df: DataFrame) -> DataFrame:
```
thvasilo, line 65: This function acts as a fail-safe, in case somehow we try to re-apply a transformation for a transformer that doesn't support re-applying.

This can only happen if a `precomputed_transformations.json` is modified by hand to include a value for a not-supported transformation.

The alternative is to raise a `NotImplementedError()` here, but I think this is safer.

jalencato: I did not fully understand the point here. I think we want to avoid to run the not-supported transformation when re-applying the transformation, right? But in this way we are still running the previous apply function on not-supported transformation, but `NotImplementedError()` will stop the whole process. Why will the method be safer?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_category_transformation.py
```diff
@@ -157,14 +165,90 @@ def apply(self, input_df: DataFrame) -> DataFrame:
...
+        per_col_label_to_one_hot_idx: dict[str, dict[str, int]] = self.json_representation[
+            "per_col_label_to_one_hot_idx"
+        ]
+        precomputed_cols: list[str] = self.json_representation["cols"]
+
+        # Assertions to ensure correctness of representation
+        assert set(precomputed_cols) == set(self.cols), (
+            f"Mismatched columns in precomputed transformation: "
+            f"pre-computed cols: {sorted(precomputed_cols)}, "
+            f"columns in current config: {sorted(self.cols)}"
```
jalencato, line 189: May we print out the diff here? I think the users will care more about the diff.

docs/source/gs-processing/usage/example.rst
```diff
@@ -173,7 +173,7 @@ we can use the following command to run the processing job locally:
 
 .. code-block:: bash
 
-    gs-processing --config-filename gconstruct-config.json \
+    gs-processing --config-filename gsprocessing-config.json \
```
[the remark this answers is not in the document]
thvasilo: Yes changed the config to the gsprocessing one as that one includes more features, and the category feature state, so the example makes more sense. I'll fix the follow-up example thanks.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/base_dist_transformation.py
thvasilo: The idea here is to have is do apply a new transformation instead of errorring out, in the convoluted even that a non-supported transformation ends up in `precomputed_transformations.json`. The scenario is the following:


* A user runs gs-processing on some data, let's say they have a numerical feature, `age`, and a categorical feature, `state`.
* We save the transformation for `state` in `precomputed_transformations.json` under the output path.
* The user modifies `precomputed_transformations.json` to include the `age` feature, and copies the file to some new input data.
* When the user runs gs-processing on the new data, we'll read an entry for `age` in `precomputed_transformations.json`, and will call `DistNumericalTransformation.apply_precomputed_transformation()` to re-apply it.
* The default implementation will run here, printing a warning and creating a new transformation.

I realize though that the scenario that "The user modifies `precomputed_transformations.json` to include the `age` feature" is very unlikely. Maybe just having `NotImplementedError()` here is better? That way if a non-supported feature ends up in `precomputed_transformations.json` for any reason we just error out.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_category_transformation.py
thvasilo: Will do.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/base_dist_transformation.py
jalencato: That makes sense.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_category_transformation.py
```diff
@@ -157,14 +165,91 @@ def apply(self, input_df: DataFrame) -> DataFrame:
...
             "per_col_label_to_one_hot_idx": per_col_label_to_one_hot_idx,
             "transformation_name": self.get_transformation_name(),
         }
 
         return dense_vector_features
 
+    def apply_precomputed_transformation(self, input_df: DataFrame) -> DataFrame:
+
+        # Get JSON representation of categorical transformation
+        labels_arrays: list[list[str]] = self.json_representation["string_indexer_labels_arrays"]
```
classicsong, line 179: Is this for task labels? and per_col_label_to_one_hot_idx is for features?
Can you add some comments to clarify it?

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -49,16 +48,21 @@
     HUGGINGFACE_TOKENIZE,
     TRANSFORMATIONS_FILENAME,
 )
-from ..config.config_parser import EdgeConfig, NodeConfig, StructureConfig
-from ..config.label_config_base import LabelConfig
-from ..config.feature_config_base import FeatureConfig
-from ..data_transformations.dist_feature_transformer import DistFeatureTransformer
-from ..data_transformations.dist_label_loader import DistLabelLoader, SplitRates, CustomSplit
-from ..data_transformations import s3_utils, spark_utils
+from graphstorm_processing.config.config_parser import EdgeConfig, NodeConfig, StructureConfig
```
classicsong, line 51: Why change it to use graphstorm_processing.config.xxx ?

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -1493,6 +1511,7 @@ def _process_edge_features(
             for each feature, which tells us the column size of the
             encoded features.
         """
+        # TODO: Add unit tests for edge feature processing
```
classicsong, line 1514: When will you add the tests?

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: pylint was throwing errors about trying to import packages beyond the top level. This ensures the right thing happens regardless of how the package is imported

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: The file is at over 80% code coverage now, so I didn't consider that a priority. This is more of a future pointer about which parts to add tests for.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_category_transformation.py
thvasilo: No the naming comes from Spark and I agree is a bit confusing: https://spark.apache.org/docs/3.5.0/api/python/reference/api/pyspark.ml.feature.StringIndexerModel.html#pyspark.ml.feature.StringIndexerModel.labelsArray

Spark refers to unique values (strings) in a categorical column as "labels". 

I'm adding comments here to clarify.

docs/source/gs-processing/usage/example.rst
```diff
@@ -211,26 +211,41 @@ and can be used downstream to create a partitioned graph.
...
 
     $ cd /tmp/gsprocessing-example
-    $ ls
-
-    edges/  launch_arguments.json  metadata.json  node_data/
-    raw_id_mappings/  perf_counters.json  updated_row_counts_metadata.json
+    $ ls -l
+
+    edges/
+    gconstruct-config_with_transformations.json
```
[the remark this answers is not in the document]
thvasilo: Agree on the confusing part, but since we want to support cross-compatibility with GConstruct, it's the least invasive solution. 

Otherwise we'd have to re-parse the entire config file and do that for both versions of the configs. 

I'd prefer to keep it this way, one is intended for end users (`gconstruct-config_with_transformations.json`) the other just to be read by GSProcessing.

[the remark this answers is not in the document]
classicsong: OK

thvasilo:
Any concerns left for this PR?

classicsong:
I am OK with the PR.
Please check with @jalencato.

classicsong (approved):

jalencato (approved):
LGTM

PR #872, opened 2024-06-11
Make IP config argument for distpart consistent with the rest of GSF

thvasilo:
*Issue #, if available:*

*Description of changes:*

All other launchers use `--ip-config` to refer to the IP list file, this PR makes the dist part launcher be consistent with th rest.


jalencato (approved):
LGTM

classicsong (approved):
LGTM

thvasilo:
Also added a default value for the `dgl-tool-path` argument, the path provided is the one we use in the SM and local images. Otherwise the job fails if the argument is not provided.

PR #875, opened 2024-06-13
[Fix] Fix save_model_frequency None error in trainer

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
This PR fixed the error when calling `fit()` function in `ep_trainer` and `lp_trainer` without specifying the `save_model_frequency` value, which will cause an operation error when compare None to int in the code `save_model_frequency > 0`.

Also add default value and its explanation in the doc string for all three kinds of trainer.


jalencato (approved):
LGTM

PR #880, opened 2024-06-17
[GPartition/SageMaker] Improvements to gpartition and SageMaker partition.

### LINKED ISSUE #802: DistPartition should move the raw id mappings to the partition output
Right now for the distributed pipeline we would first create raw id mappings using GSProcessing, which the second step of DistPart does not touch.

When a user runs a training job on the DistPart output, they will not be able to get embeddings with their original ids by default.

To fix this, our local and SageMaker DistPart runners, that both expect GSProcessing-produced files as input, should also copy the raw_id_mappings under the same output directory as the dist_graph is created.

Our current code assumes the raw id mappings are placed at the same level as the dist graph: https://github.com/awslabs/graphstorm/blob/7e2cdfdc91266f50f31e7d192d4a51e84dd55e55/python/graphstorm/gconstruct/remap_result.py#L655

We can also provide the option to provide the raw id mapping path during training.

thvasilo:
We use this commit as a base for range partitioning.

*Issue #, if available:*

Fixes #802 

*Description of changes:*

* Download files in parallel from S3 for SageMaker partition.
* Copy raw_id_mappings from the input path to the generated DGL graph path for SageMaker and local. 
* Change dtype of random partition assignment pyarrow array, from int64 to uint8 for cases where num_parts < 256, uint16 otherwise.
* Support getting SageMaker job information for both SageMaker Processing jobs, and SageMaker Training jobs, allowing us to use more instance types.
* Make gpartition unit test verification re-usable across partition algorithm implementations.

We use this PR as a base for RangePartition.

### DIFF
--- a/python/graphstorm/gpartition/dist_partition_graph.py
+++ b/python/graphstorm/gpartition/dist_partition_graph.py
@@ -24,13 +24,17 @@ import logging
 import os
 import queue
 import time
+import shutil
 import subprocess
 import sys
 from typing import Dict
 from threading import Thread
 
-from graphstorm.gpartition import (ParMetisPartitionAlgorithm, ParMETISConfig,
-    RandomPartitionAlgorithm)
+from graphstorm.gpartition import (
+    ParMetisPartitionAlgorithm,
+    ParMETISConfig,
+    RandomPartitionAlgorithm,
+)
 from graphstorm.utils import get_log_level
 
 
@@ -133,7 +137,10 @@ def main():
         part_assignment_dir)
 
     part_end = time.time()
-    logging.info("Partition assignment took %f sec", part_end - part_start)
+    logging.info("Partition assignment with algorithm '%s' took %f sec",
+                 args.partition_algorithm,
+                 part_end - part_start,
+    )
 
     if args.do_dispatch:
         run_build_dglgraph(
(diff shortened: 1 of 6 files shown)


python/graphstorm/gpartition/dist_partition_graph.py
```diff
@@ -147,6 +154,18 @@ def main():
 
         logging.info("DGL graph building took %f sec", part_end - time.time())
 
+    # Copy raw_id_mappings to dist_graph if they exist in the input
+    raw_id_mappings_path = os.path.join(args.input_path, "raw_id_mappings")
+
+    if os.path.exists(raw_id_mappings_path):
+        logging.info("Copying raw_id_mappings to dist_graph")
```
jalencato, line 161: Any reason we do the copy here?

python/graphstorm/sagemaker/sagemaker_partition.py
```diff
@@ -24,15 +24,20 @@
...
 import subprocess
 from threading import Thread, Event
+from typing import List
 
-import numpy as np
 import boto3
+import botocore
+import numpy as np
 import sagemaker
+from joblib import Parallel, delayed # pylint: disable=wrong-import-order
```
jalencato, line 33: If the pylint error is caused by the line 18? If we put the ``from dataclasses import dataclass`` under the ``from typing import List``, it may solve the pylint problem.

jalencato (commented):
Some small comments

python/graphstorm/sagemaker/sagemaker_partition.py
thvasilo: Nope it's from the line above importing `sagemaker`, see https://github.com/awslabs/graphstorm/issues/881

python/graphstorm/gpartition/dist_partition_graph.py
thvasilo: You here in particular or in why do we copy in general? We need to copy the mappings from their S3 location at the input (GSProcessing output) to where training will run, so that the ID remapping script can work. ID remapping assumes the `raw_id_mappings` directory exists at the same level as the part config JSON

classicsong (approved):
LGTM

jalencato (approved):
LGTM

PR #886, opened 2024-06-20
[Doc] Fix the type error in example Notebook 2

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
This PR fix the type error of `reverse_edge_types_map` in the API example notebook 2. When it should be a dictionary. When set the `exclude_training_targets` to be `True`, this map will be used.


classicsong (approved):
LGTM

PR #888, opened 2024-06-20
Reduce the number of warning prints of heterogeneous GNN encoders when some nodes do not have in edges

classicsong:
*Issue #, if available:*

*Description of changes:*
To avoid excessive warning logs from RGCN, RGAT, and HGT encoders, this PR ensures that identical warnings (warnings that share the same message) are printed only once.


zhjwy9343 (approved):
LGTM

Issue #799, opened 2024-04-15
Print warnings or error out when invalid arguments are passed


thvasilo:
With the number of arguments that we parse, it's likely that an arg can be misspelled.

e.g. I ran into the case where I was trying to pass `--save-embeds-path` instead of `save-embed-path` which led to my job hanging without an obvious reason.

When unknown arguments are passed to the launch scripts, we should be at least print out a warning, or error out the job (as it will likely lead to something the user did not intend).

More context:

What's happened is that the yaml file had a value for the path, which I tried to override through the CLI.

But the misspelled CLI arg was just ignored and it tried to use the one from the yaml.

classicsong:
Print out a warning.
We need to support customized args.

thvasilo:
Just encountered this with a user launching a SageMaker job. They provided `--instance_type` instead of `--instance-type` to the launch script, so the job was being launched with the default instance type (`ml.g4dn.12xlarge`)

PR #891, opened 2024-06-25
Print unknown arguments when executing GraphStorm command-lines.

classicsong:
*Issue #, if available:*
#799 

*Description of changes:*
Add Warnings for unknown arguments. 

### DIFF
--- a/python/graphstorm/run/gsgnn_dt/distill_gnn.py
+++ b/python/graphstorm/run/gsgnn_dt/distill_gnn.py
@@ -17,6 +17,7 @@
 """
 
 import os
+import logging
 import torch as th
 import graphstorm as gs
 from graphstorm.config import GSConfig, get_argument_parser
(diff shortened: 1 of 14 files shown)


thvasilo:
Do we want warnings or error out when unknown args are provided?

I would vote for error out (which is also the default Python behavior), to ensure we honor the user's intent.

e.g. if the user provides:

```
--use-node-embedings true # this is a typo
```

the default value (false) will be used instead, and the only way for the user to know something went wrong is if they dig into the logs, which are hard to parse with all the printouts.

classicsong:
> Do we want warnings or error out when unknown args are provided?
> 
> I would vote for error out (which is also the default Python behavior), to ensure we honor the user's intent.
> 
> e.g. if the user provides:
> 
> ```
> --use-node-embedings true # this is a typo
> ```
> 
> the default value (false) will be used instead, and the only way for the user to know something went wrong is if they dig into the logs, which are hard to parse with all the printouts.

There are two concerns for error out:
1) We have to add all the arguments from remap_result.py into the training loop, which is meaningless.
2) Originally, people can customize the GSConfig by just passing those "unknown arguments" and they can access them through config.udf_1, config.udf_2. If we error out, we will lose this flexibility.

thvasilo (approved):
Approve but let's keep in mind that if we run into problems in the future we can re-visit this decision.

Issue #773, opened 2024-03-15
DGL edge_split returns None when there's single edge of an etype


haozhouamzn:
I got following error using graphstorm
```
Traceback (most recent call last):
  File "/graphstorm/python/graphstorm/run/gsgnn_lp/gsgnn_lp.py", line 254, in <module>
    main(gs_args)
  File "/graphstorm/python/graphstorm/run/gsgnn_lp/gsgnn_lp.py", line 106, in main
    train_data = GSgnnLPTrainData(config.graph_name,
  File "/graphstorm/python/graphstorm/dataloading/dataset.py", line 641, in __init__
    super(GSgnnLPTrainData, self).__init__(graph_name, part_config,
  File "/graphstorm/python/graphstorm/dataloading/dataset.py", line 520, in __init__
    super(GSgnnEdgeTrainData, self).__init__(graph_name, part_config, label_field,
  File "/graphstorm/python/graphstorm/dataloading/dataset.py", line 398, in __init__
    super(GSgnnEdgeData, self).__init__(graph_name, part_config,
  File "/graphstorm/python/graphstorm/dataloading/dataset.py", line 219, in __init__
    self.prepare_data(self._g)
  File "/graphstorm/python/graphstorm/dataloading/dataset.py", line 565, in prepare_data
    assert train_idx is not None, "There is no training data."
```

This trace back to following code block in [dataset.py](https://github.com/awslabs/graphstorm/blob/f042050424ff157a0c42454f0975b223adaf149a/python/graphstorm/dataloading/dataset.py#L557-L559)
```
        for canonical_etype in self.train_etypes:
            if 'train_mask' in g.edges[canonical_etype].data:
                train_idx = dgl.distributed.edge_split(
                    g.edges[canonical_etype].data['train_mask'],
                    pb, etype=canonical_etype, force_even=True)
            else:
                # If there are no training masks, we assume all edges can be used for training.
                # Therefore, we use a more memory efficient way to split the edge list.
                # TODO(zhengda) we need to split the edges properly to increase the data locality.
                train_idx = split_full_edge_list(g, canonical_etype, get_rank())
            assert train_idx is not None, "There is no training data."
            num_train += len(train_idx)
            train_idxs[canonical_etype] = train_idx
```

The error seems because there's a single edge of this `canonical_etype`, that is `len(g.edges[canonical_etype].data['train_mask'])==1`. As a result, `edge_split` returns None.

classicsong:
[...]

PR #896, opened 2024-06-26
Add more information in the error message when the training set is empty or too small.

classicsong:
*Issue #, if available:*
#773

If the size of the training set is smaller than the number of trainers (GPUs), it is possible that some trainers will not have any training data. The training loop will finally fail in this case.

*Description of changes:*
Update the error message to provide more information for users to debug the input data.

### DIFF
--- a/python/graphstorm/dataloading/dataset.py
+++ b/python/graphstorm/dataloading/dataset.py
@@ -544,8 +544,8 @@ class GSgnnData():
                     "empty or the number of training nodes is smaller than " \
                     f"the number of trainers {get_world_size()}" \
                     "Please check your training data and make sure " \
-                    "the number of trainers (GPUs) is smaller than the " \
-                    "number of training data."
+                    "the number of trainers (GPUs, if using GPU-supported " \
+                    "machines) is smaller than the number of training data."
                 num_train += len(train_idx)
                 train_idxs[ntype] = train_idx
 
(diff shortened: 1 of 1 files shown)


python/graphstorm/dataloading/dataset.py
```diff
@@ -539,7 +539,13 @@ def get_node_train_set(self, ntypes, mask="train_mask"):
                 train_idx = dgl.distributed.node_split(g.nodes[ntype].data[msk],
                                                        pb, ntype=ntype, force_even=True,
                                                        node_trainer_ids=node_trainer_ids)
-                assert train_idx is not None, "There is no training data."
+                assert train_idx is not None, \
+                    f"The training set of the {ntype} nodes is " \
+                    "empty or the number of training nodes is smaller than " \
+                    f"the number of trainers {get_world_size()}" \
+                    "Please check your training data and make sure " \
+                    "the number of trainers (GPUs) is smaller than the " \
```
zhjwy9343, line 547: "(GPUs)" -> "(GPUs, if using GPU-supported machines)"

python/graphstorm/dataloading/dataset.py
```diff
@@ -770,7 +776,13 @@ def get_edge_train_set(self, etypes=None, mask="train_mask",
                 # TODO(zhengda) we need to split the edges properly to increase the data locality.
                 train_idx = split_full_edge_list(g, canonical_etype, get_rank())
 
-            assert train_idx is not None, "There is no training data."
+            assert train_idx is not None, \
+                f"The training set of the {canonical_etype} edges is " \
+                "empty or the number of training edges is smaller than " \
+                f"the number of trainers {get_world_size()}" \
+                "Please check your training data and make sure " \
+                "the number of trainers (GPUs) is smaller than the " \
```
zhjwy9343, line 784: ditto

zhjwy9343 (approved):
LGTM, with a minor suggestion

Issue #785, opened 2024-03-26
When running a link prediction task, it won't give --node-emb-dir to remap script.


classicsong:
When running a link prediction training task, it won't give --node-emb-dir to remap script when --save-embedding-path is True.

PR #901, opened 2024-06-27
Add some test for testing ID remapping after link prediction training

classicsong:
*Issue #, if available:*
#785

*Description of changes:*
The end2end test does not cover the ID remapping after link prediction training use case.
Add some tests to cover the use cases.


jalencato (commented):
Overall it LGTM. It seems to break some old tests. May you take a look at it?

jalencato (approved):

PR #909, opened 2024-07-06
[Bugfix] Loss funcs for multitask learning should also be added as torch nn.module

classicsong:
*Issue #, if available:*
The loss funcs of GSgnnMultiTaskSharedEncoderModel are not added as nn.modules of GSgnnMultiTaskSharedEncoderModel. This will cause device error in some cases.

*Description of changes:*
Adding loss funcs as a nn.ModuleDict() of GSgnnMultiTaskSharedEncoderModel.


jalencato (approved):
LGTM

PR #915, opened 2024-07-10
[GSProcessing] Fix bug in repartition on leader, fix mypy errors and latent bugs

thvasilo:
*Issue #, if available:*

*Description of changes:*
* During the refactor with https://github.com/awslabs/graphstorm/pull/870 we moved the loader to be a class var for DistributedExecutor, but because the S3 path is not unit tested we missed on case where the output is on S3 and the user requests repartition on leader.
* This error was actually picked up by mypy, I fixed some other potential issues and type annotations here.

### Testing

Pre-commit, unit tests, and one test SageMaker job all succeed.

The S3 codepath can only be integration tested currently.

### DIFF
--- a/graphstorm-processing/graphstorm_processing/distributed_executor.py
+++ b/graphstorm-processing/graphstorm_processing/distributed_executor.py
@@ -313,6 +313,7 @@ class DistributedExecutor:
         )
 
         repartition_start = time.perf_counter()
+        updated_metadata = {}
         if all_match:
             logging.info(
                 "All file row counts match, applying Parquet metadata modification on Spark leader."
--- a/graphstorm-processing/graphstorm_processing/distributed_executor.py
+++ b/graphstorm-processing/graphstorm_processing/distributed_executor.py
@@ -327,7 +327,7 @@ class DistributedExecutor:
                 try:
                     # Upload existing output files before trying to re-partition
                     if self.filesystem_type == FilesystemType.S3:
-                        self._upload_output_files(loader, force=True)
+                        self._upload_output_files(self.loader, force=True)
                     updated_metadata = repartition_files(graph_meta_dict, repartitioner)
                 except Exception as e:  # pylint: disable=broad-exception-caught
                     # If an error happens during re-partition, we don't want to fail the entire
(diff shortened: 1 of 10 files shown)


graphstorm-processing/graphstorm_processing/distributed_executor.py
```diff
@@ -327,7 +327,7 @@ def run(self) -> None:
                 try:
                     # Upload existing output files before trying to re-partition
                     if self.filesystem_type == FilesystemType.S3:
-                        self._upload_output_files(loader, force=True)
+                        self._upload_output_files(self.loader, force=True)
```
thvasilo, line 330: This was the bug that would cause repartition on leader to fail when the output is on S3.

jalencato: Just want to double check, does the error only occur on emr ?

graphstorm-processing/graphstorm_processing/repartition_files.py
```diff
@@ -325,7 +325,9 @@ def _repartition_parquet_files_in_memory(
             for idx in range(len(desired_counts))
         ]
         with Parallel(
-            n_jobs=min(NUM_WRITE_THREADS, os.cpu_count()), verbose=self.verbosity, prefer="threads"
+            n_jobs=min(NUM_WRITE_THREADS, os.cpu_count() or 16),
```
jalencato, line 328: How do we determine we need to use 16 here?

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -25,6 +25,7 @@
 from time import perf_counter
 from typing import Any, Dict, Optional, Set, Tuple
 
+from pyspark import RDD
```
jalencato, line 28: Where do we use RDD here?

jalencato (commented):
Some small comment

graphstorm-processing/graphstorm_processing/distributed_executor.py
thvasilo: No it would occur any time S3 is used with `--do-repartition True`

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: It's used in a type annotation within the file.

graphstorm-processing/graphstorm_processing/repartition_files.py
thvasilo: It's a safe number to use according to my testing. In some systems `os.cpu_count()` might return `None` leading to an error here. We provide 16 as a fallback value.

jalencato (approved):
LGTM

PR #916, opened 2024-07-10
Add support to deduplicate same input features when computing rank gauss norm.

classicsong:
*Issue #, if available:*
Rank gauss norm usually works well with floating point features. However, there can be cases when the input features have many duplicated values (for example, a feature can be [0,1,1,2,1,2, ...]). In such cases, we would better deduplicate the input features before computing the rank gauss norm for each feature and then remap the normalized features back to the original features, so that input features with the same value will have the same normalized feature value.

*Description of changes:*
Add an option to the "rank_gauss" normalization operation to enable/disable deduplication.


docs/source/configuration/configuration-gconstruction.rst
```diff
@@ -90,7 +90,7 @@ Currently, the graph construction pipeline supports the following feature transf
...
-* **Numerical Rank Gauss transformation** normalizes numerical input features with rank gauss normalization. It maps the numeric feature values to gaussian distribution based on ranking. The method follows https://www.kaggle.com/c/porto-seguro-safe-driver-prediction/discussion/44629#250927. The ``name`` field in the feature transformation dictionary is ``rank_gauss``. The dict can contains one optional field, i.e., ``epsilon`` which is used to avoid INF float during computation.
+* **Numerical Rank Gauss transformation** normalizes numerical input features with rank gauss normalization. It maps the numeric feature values to gaussian distribution based on ranking. The method follows https://www.kaggle.com/c/porto-seguro-safe-driver-prediction/discussion/44629#250927. The ``name`` field in the feature transformation dictionary is ``rank_gauss``. The dict can contains one optional field, i.e., ``epsilon`` which is used to avoid INF float during computation and ``uniquify`` which controls whether deduplicating input features before computing rank gauss norm.
```
zhjwy9343, line 93: Suggest to give an example of the dict, e.g.,

`... gauss norm. For example, the transform could be like 
``"transform": {
                "name": 'rank_gauss',
                "epsilon": 1e-5,
                "uniquify": True
            },
``
'

zhjwy9343 (approved):
LGTM, except for an exmaple.

docs/source/configuration/configuration-gconstruction.rst
classicsong: Shall we have another PR to add one example for each transformation?

PR #917, opened 2024-07-11
[Docker] Pin the numpy version to 1.26.4 for low torch version

jalencato:
*Issue #, if available:*

*Description of changes:*


thvasilo (approved):
LGMT overall, but could we remove the comment on line 30?

jalencato:
> LGMT overall, but could we remove the comment on line 30?

I delete that sentense, then we should keep in mind if any users face issues like that

PR #922, opened 2024-07-13
[GSProcessing] Add option to truncate vectors with no-op transformation.

thvasilo:
*Issue #, if available:*

*Description of changes:*

* We add a keyword argument to the GSProcessing no-op transformation, named `truncate_dim`. When a user provides this parameter, we will try to truncate the input vectors to the dimension specified. This allows users to easily make use of input Matryoshka embeddings.
* (Revision) Added the same transformation for GConstruct.

## Testing

Unit tests passing, unit and integration test added for the new transformation option.


graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_noop_transformation.py
```diff
@@ -95,15 +122,17 @@ def str_list_to_float_vec(string_list: Optional[List[str]]) -> Optional[List[flo
                     for column in self.cols
                 ]
             )
-            return input_df
         else:
-            return input_df.select(
+            input_df = input_df.select(
```
jalencato, line 126: Any reason we remove the return statement here?

docs/source/graph-construction/gs-processing/input-configuration.rst
```diff
@@ -380,6 +380,12 @@ arguments.
          split the values in the column and create a vector column
          output. Example: for a separator ``'|'`` the CSV value
          ``1|2|3`` would be transformed to a vector, ``[1, 2, 3]``.
+      - ``truncate_dim`` (Integer, optional): Relevant for vector inputs.
```
jalencato, line 383:
```suggestion
      - ``truncate_dim`` (Integer, Optional): Relevant for vector inputs.
```

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_noop_transformation.py
```diff
@@ -72,13 +94,18 @@ def apply(self, input_df: DataFrame) -> DataFrame:
...
                     f"for column {self.cols[0]}"
                 )
-            return input_df
+            if self.truncate_dim:
+                return self._truncate_vector_df(input_df)
+            else:
+                return input_df
         elif isinstance(col_datatype, NumericType):
+            if self.truncate_dim is not None:
+                warnings.warn(f"Trying use {self.truncate_dim=} on a DataFrame of scalars!")
```
jalencato, line 103: What about using assert here? It is an undefined behavior and should be avoided by the user.

jalencato (commented):
Some small comment

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_noop_transformation.py
thvasilo: The logic is changed below, either side of the if/else statement now just a assigns a value to input_df, which is the conditionally passed through the truncate function before returning the DF.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_noop_transformation.py
thvasilo: Depends on where we're leaning with interpreting user input. In most cases in GS we try to guess what the user intended and proceed with execution, minimizing the number of times users will see failed jobs, if it's safe to do so (i.e. not overriding user intentions).

I'd rather keep the warning, but won't mind changing.

thvasilo:
Added GCosntsruct support @jalencato let me know if there's something missing for config parsing. I couldn't find any gconstruct docs in the current doc structure, they'll need to be added later or we can rebase this PR once those are added back.

jalencato:
> Added GCosntsruct support @jalencato let me know if there's something missing for config parsing. I couldn't find any gconstruct docs in the current doc structure, they'll need to be added later or we can rebase this PR once those are added back.

It should be just about the feature transformation configuration for GConstruct: https://graphstorm.readthedocs.io/en/latest/configuration/configuration-gconstruction.html#configurations-gconstruction

python/graphstorm/gconstruct/transform.py
```diff
@@ -941,6 +946,12 @@ def call(self, feats):
         assert np.issubdtype(feats.dtype, np.integer) \
                 or np.issubdtype(feats.dtype, np.floating), \
                 f"The feature {self.feat_name} has to be integers or floats."
+        if self.truncate_dim is not None:
+            if isinstance(feats, np.ndarray):
+                feats = feats[:, :self.truncate_dim]
+            else:
+                assert isinstance(feats, ExtMemArrayWrapper)
```
jalencato, line 953: May we add a comment here about a converting the feats from an ExtMemArrayWrapper to in-memory array?

python/graphstorm/gconstruct/transform.py
```diff
@@ -1148,7 +1159,12 @@ def parse_feat_ops(confs, input_data_format=None):
 
         out_dtype = _get_output_dtype(feat['out_dtype']) if 'out_dtype' in feat else None
         if 'transform' not in feat:
-            transform = Noop(feat['feature_col'], feat_name, out_dtype=out_dtype)
+            transform = Noop(
+                feat['feature_col'],
+                feat_name,\
```
thvasilo, line 1164:
```suggestion
                feat_name,
```

jalencato (approved):
LGTM

PR #931, opened 2024-07-23
[GPartition] Rename the input argument for dispatch

jalencato:
*Issue #, if available:*

*Description of changes:*

We now reverse the meaning of the original ``--do-dispatch`` input argument.

Successfully test with 
```
python3 -m graphstorm.gpartition.dist_partition_graph \
    --input-path ./test-gsprocessing/ \
    --metadata-filename updated_row_counts_metadata.json \
    --output-path ./gspartition-example/ \
    --num-parts 1 \
    --dgl-tool-path ./dgl/tools \
    --partition-algorithm random \
    --ip-config ip_list.txt \
    --ssh-port 2222 \
    --partition-assignment-only
```


python/graphstorm/gpartition/dist_partition_graph.py
```diff
@@ -189,7 +191,8 @@ def parse_args() -> argparse.Namespace:
     argparser.add_argument("--ip-config", type=str,
                            help=("A file storing a list of IPs, one line for "
                                 "each instance of the partition cluster."))
-    argparser.add_argument("--do-dispatch", action='store_true')
+    argparser.add_argument("--partition-assignment-only", action='store_true',
+                           help="Only generate partition assignment, do not build DGL graph")
```
classicsong, line 195:
```suggestion
                           help="Only generate partition assignments for nodes, the process will not build the partitioned DGL graph")
```

thvasilo:
Marking as breaking change since this changes a user-facing argument.

thvasilo (approved):
Changes LGTM, this better communicates intent and ensures we do the file creation by default, which is the most common use case.

Do we have integ tests for dist_partition_graph? We should in the future if not.

jalencato:
> Changes LGTM, this better communicates intent and ensures we do the file creation by default, which is the most common use case.
> 
> Do we have integ tests for dist_partition_graph? We should in the future if not.

Let us create an issue and do it in another PR, it should request us to add some more dependencies.

Issue #930, opened 2024-07-23
[Bug] In multitask learning inference, it is possible that there is no prediction task defined.


classicsong:
The mt_infer will crash when there is no prediction task defined during inference.
The error looks like:
```
"/opt/ml/code/graphstorm/python/graphstorm/inference/mt_infer.py", line 280, in infer
  File "/opt/ml/code/graphstorm/python/graphstorm/inference/mt_infer.py", line 280, in infer
    task_infos = predict_test_loader.task_infos
AttributeError: 'NoneType' object has no attribute 'task_infos'
```

PR #938, opened 2024-07-26
[Bugfix] Fix the case when there is no prediction task defined in multitask learning inference.

classicsong:
*Issue #, if available:*
#930 

*Description of changes:*
Fix the bug.


zhjwy9343 (approved):
LGTM

PR #944, opened 2024-07-30
[GSProcessing] Add default argument values for optional numerical transformation kwargs

thvasilo:
Also adds tests and fixes some pylint errors in test file.

*Issue #, if available:*

*Description of changes:*

* Previously if a user provided a numerical transformation config where imputer, norm, or separator (for multi-num) was missing, the job would fail since the object could not be instantiated without all required arguments, although we list those args as optional.
* This PR adds default values for all optional args in the classes' __init__
* Also adds a couple of tests and fixes a few pylint errors in the test file.
* Also add a test for the case where the multi-numerical input is a vector value (which is the case for Parquet vector inputs)

### DIFF
--- a/graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
+++ b/graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
@@ -236,8 +236,8 @@ class DistNumericalTransformation(DistributedTransformation):
     def __init__(
         self,
         cols: Sequence[str],
-        normalizer: str,
-        imputer: str,
+        normalizer: str = "none",
+        imputer: str = "none",
         out_dtype: str = TYPE_FLOAT32,
         epsilon: float = 1e-6,
     ) -> None:
@@ -290,9 +290,9 @@ class DistMultiNumericalTransformation(DistNumericalTransformation):
     def __init__(
         self,
         cols: Sequence[str],
-        separator: Optional[str],
-        normalizer: str,
-        imputer: str,
+        separator: Optional[str] = None,
+        normalizer: str = "none",
+        imputer: str = "none",
         out_dtype: str = TYPE_FLOAT32,
     ) -> None:
         assert (
--- a/graphstorm-processing/tests/test_dist_numerical_transformation.py
+++ b/graphstorm-processing/tests/test_dist_numerical_transformation.py
@@ -30,7 +30,7 @@ from pyspark.sql.types import (
     StringType,
     LongType,
 )
-from scipy.special import erfinv
+from scipy.special import erfinv  # pylint: disable=no-name-in-module
 
 from graphstorm_processing.data_transformations.dist_transformations import (
     DistNumericalTransformation,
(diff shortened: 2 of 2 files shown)


graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
```diff
@@ -236,8 +236,8 @@ class DistNumericalTransformation(DistributedTransformation):
     def __init__(
         self,
         cols: Sequence[str],
-        normalizer: str,
-        imputer: str,
+        normalizer: str = "none",
```
jalencato, line 239: Do we address this default value in any doc?

graphstorm-processing/tests/test_dist_numerical_transformation.py
```diff
@@ -30,7 +30,7 @@
     StringType,
     LongType,
 )
-from scipy.special import erfinv
+from scipy.special import erfinv  # pylint: disable=no-name-in-module
```
jalencato, line 33: Why do we need this disable? I suppose the erfinv should be in the scipy.special

graphstorm-processing/tests/test_dist_numerical_transformation.py
```diff
@@ -377,10 +427,11 @@ def test_rank_gauss(spark: SparkSession, check_df_schema, epsilon):
 
 @pytest.mark.parametrize("epsilon", [0.0, 1e-6])
 def test_rank_gauss_reshuffling(spark: SparkSession, check_df_schema, epsilon):
+    """Test rank-Gauss transformation results order with enforced shuffling."""
     # Create DF with 10k values
     random_values = np.random.rand(10**3, 1)
 
-    # Convert the array of values into a list of single-value lists
+    # Convert the array of values into a list of single-value tuples
```
jalencato, line 434: Is this a new change in this PR or we fixed it in the previous PR?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_numerical_transformation.py
thvasilo: Yup, can see the descriptions here: https://w.amazon.com/bin/view/Amazon-Science/publishing/Overview/#HSciPubPolicyDeadlines

thvasilo: Yup can see the descriptions here: https://graphstorm.readthedocs.io/en/latest/graph-construction/gs-processing/input-configuration.html#supported-transformations

graphstorm-processing/tests/test_dist_numerical_transformation.py
thvasilo: It should but for some reason it's not picked up by pylint

graphstorm-processing/tests/test_dist_numerical_transformation.py
thvasilo: We don't make any changes in the code, just making the comment more accurate.

```python
data = [(float(value),) for value in random_values]
```

`(float(value),)` is a single-element tuple (therefore immutable), not a list.

jalencato (approved):
LGTM

PR #955, opened 2024-08-06
[GSProcessing] Fix isinstance check for numerical features

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Fixes corner case where a feature value might not be one of int/float (we don't cast Parquet input types), replacing with the more general `numbers.Number`.

Example of how to trigger a failure, as observed by a customer:

```python
>>> import numbers
>>> import decimal
>>> dn = decimal.Decimal(5)
>>> intnum = 5
>>> isinstance(intnum, (float, int))
True
>>> isinstance(dn, (float, int))
False
>>> isinstance(dn, numbers.Number)
True
```

### DIFF
--- a/graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
+++ b/graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
@@ -1086,7 +1086,9 @@ class DistHeterogeneousGraphLoader(object):
                 node_type_feature_metadata[feat_name] = node_feature_metadata_dict
 
                 feat_val = single_feature_df.take(1)[0].asDict().get(feat_name)
-                assert isinstance(feat_val, (list, numbers.Number))
+                assert isinstance(
+                    feat_val, (list, numbers.Number)
+                ), "We expect features to either be scalars or lists of scalars."
                 nfeat_size = 1 if isinstance(feat_val, numbers.Number) else len(feat_val)
                 ntype_feat_sizes.update({feat_name: nfeat_size})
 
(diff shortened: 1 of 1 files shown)


graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -1085,8 +1085,9 @@ def write_processed_feature(
                 }
                 node_type_feature_metadata[feat_name] = node_feature_metadata_dict
 
-                feat_val = single_feature_df.take(1)[0].asDict().get(feat_name, None)
-                nfeat_size = 1 if isinstance(feat_val, (int, float)) else len(feat_val)
```
jalencato, line 1089: Better to have a comment here to briefly explain about it.

jalencato (approved):
LGTM, but a small comment.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: I'll add a message to the assertion that can act as an explanation as well.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -1085,8 +1085,11 @@ def write_processed_feature(
...
+                feat_val = single_feature_df.take(1)[0].asDict().get(feat_name)
+                assert isinstance(
+                    feat_val, (list, numbers.Number)
+                ), "We expect features to either be scalars or lists of scalars."
```
classicsong, line 1091: If feat_val is a list, shall we check whether every object in the list is a Number?

thvasilo: Sure, I'll add that.

thvasilo:
Made a small change to avoid code repetition and to implement @classicsong suggestion

@jalencato can you take a second look? The last commit pulls out the write_feature function and shares the implementation between nodes and edges.

PR #968, opened 2024-08-14
Fix the bug when the number of val/test samples is small

classicsong:
*Issue #, if available:*
#959 

*Description of changes:*

### DIFF
--- a/python/graphstorm/dataloading/dataset.py
+++ b/python/graphstorm/dataloading/dataset.py
@@ -594,7 +594,7 @@ class GSgnnData():
                         # To avoid the error, force each trainer or
                         # inferencer to use the entire validation
                         # or test set.
-                        idx = th.nonzero(g.nodes[ntype].data[msk])
+                        idx = th.nonzero(g.nodes[ntype].data[msk]).reshape(-1,) # 1D tensor
                         idxs[ntype] = idx
 
                 logging.debug('part %d | ntype %s, mask %s | num nodes: %d',
--- a/python/graphstorm/dataloading/dataset.py
+++ b/python/graphstorm/dataloading/dataset.py
@@ -598,7 +598,7 @@ class GSgnnData():
                             th.arange(g.num_nodes(ntype))]).reshape(-1,) # 1D tensor
                         idxs[ntype] = idx
 
-                logging.debug('part %d | ntype %s, mask %s | num nodes: %d',
+                logging.debug('part %d | ntype %s, mask %s | val/test: %d',
                           get_rank(), ntype, msk, len(idx))
         return idxs, num_data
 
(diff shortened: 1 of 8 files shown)


python/graphstorm/dataloading/dataset.py
```diff
@@ -579,10 +579,26 @@ def _get_node_set(self, ntypes, mask):
...
+                        idxs[ntype] = idx
+                    else:
+                        # There is not enough validation or test data.
+                        # One or more validation/test dataloader will be
+                        # empty, which will cause an evaluation error.
+                        #
+                        # To avoid the error, force each trainer or
+                        # inferencer to use the entire validation
+                        # or test set.
+                        idx = th.nonzero(g.nodes[ntype].data[msk][ \
```
jalencato, line 597: Shall we add a log here to indicate users that we are using the entire validation/test set? We can set it to debug message.

python/graphstorm/dataloading/dataset.py
```diff
@@ -814,8 +830,24 @@ def _get_edge_set(self, etypes, mask, reverse_edge_types_map):
...
+                        # will not be empty
+                        idxs[canonical_etype] = idx
+                    else:
+                        # There is not enough validation or test data.
+                        # One or more validation/test dataloader will be
+                        # empty, which will cause an evaluation error.
+                        #
+                        # To avoid the error, force each trainer or
+                        # inferencer to use the entire validation
+                        # or test set.
```
jalencato, line 847: Same here.

python/graphstorm/model/multitask_gnn.py
```diff
@@ -553,14 +553,14 @@ def multi_task_mini_batch_predict(
                                                     device,
                                                     return_proba,
                                                     return_label)
-                    assert labels is None or len(labels) == 1, \
+                    assert not labels or len(labels) == 1, \
```
jalencato, line 556: Any reason we make the change here? Ask because ``not labels`` control larger scope than ``labels is None``.

tests/end2end-tests/data_gen/movielens_small_val.json
```diff
@@ -0,0 +1,55 @@
...
+                    "source_id_col":    "src_id",
+                    "dest_id_col":      "dst_id",
+                    "relation":         ["user", "rating", "movie"],
+                    "format":           {"name": "parquet"},
+                    "files":        "/data/ml-100k/edges.parquet",
+                    "labels":	[
+                        {
+                            "label_col":	"rate",
+                            "task_type":	"classification",
+                            "split_pct":	[0.1, 0.00001, 0.00001]
```
jalencato, line 50: What is the size of the validation/test set here? Will it be zero?

classicsong: exactly 1.

python/graphstorm/model/multitask_gnn.py
classicsong: not labels can handle the empty dict cases. If the labels is `{}`, it will not crash the program.

python/graphstorm/dataloading/dataset.py
classicsong: Good callout

PR #994, opened 2024-08-28
[GSProcessing] Initial support for multi-task mask generation

### LINKED ISSUE #992: Add multi-task GSProcessing support for random mask generation

thvasilo:
Currently support only random mask generation.

*Issue #, if available:*

Fixes #992

*Description of changes:*

We add support for multi-task mask generation to GSProcessing.

~~To achieve this we add a boolean `is_multitask` class variable to `DistHeterogeneousGraphLoader`. We set this to true when there's more than one label defined in the input configuration.~~

~~When `is_multitask` is True, we append the label column name to the mask name of every label. For link prediction tasks, we append `_lp` to the mask name. We also modify `repartition_files.py` accordingly.~~

We use the same UX as with GConstruct: users are expected to provide a `mask_field_names` entry for each label they want to use in MT training.

We add an example to documentation and link to it from the main multi-task training guide. 

### DIFF
--- a/docs/source/cli/graph-construction/distributed/gsprocessing/input-configuration.rst
+++ b/docs/source/cli/graph-construction/distributed/gsprocessing/input-configuration.rst
@@ -497,18 +497,16 @@ arguments.
 Creating a graph for multi-task training
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-To create a graph for multi-task training, you simply need to
-define a label in your config for each of the labels you want to
-use during training. GSProcessing will generate separate
-train/val/test masks for each of your labels, which
-will be named ``<mask_type>_<label_column>``, e.g.
-if your label columns were ``label_class``, and ``label_reg``
-GSProcessing will generate masks named as ``train_mask_label_class``,
-``train_mask_label_reg`` etc. For link prediction tasks, the masks
-will have ``_lp`` added as a suffix, e.g. ``train_mask_lp``.
+To create a graph for multi-task training, you need to
+define custom label names in your config for each of the labels you want to
+use during training in a ``mask_field_names`` entry for each label config.
+GSProcessing will generate separate
+train/val/test masks for each of your labels named
+accordingly.
 
-After partitioning the data, you can then provide these masks
-in your train YAML file to use during multi-task training.
+After partitioning the data, you need to then provide the same
+mask names in a ``mask_fields`` entry
+in your train YAML file during multi-task training.
 
 For details on running multi-task training see
 :doc:`/advanced/multi-task-learning`.
@@ -555,7 +553,12 @@ type.
                             "train": 0.8,
                             "val": 0.1,
                             "test": 0.1
-                        }
+                        },
+                        "mask_field_names": [
+                            "train_mask_class",
+                            "val_mask_class",
+                            "test_mask_class"
+                        ]
                     }
                 ]
             }
@@ -587,7 +590,12 @@ type.
                             "train": 0.8,
                             "val": 0.1,
                             "test": 0.1
-                        }
+                        },
+                        "mask_field_names": [
+                            "train_mask_lp",
+                            "val_mask_lp",
+                            "test_mask_lp"
+                        ]
                     }
                 ]
             }
@@ -636,9 +644,9 @@ train YAML file to run multi-task training:
                 target_ntype: "paper"
                 label_field: "label"
                 mask_fields:
-                    - "train_mask_label"
-                    - "val_mask_label"
-                    - "test_mask_label"
+                    - "train_mask_class"
+                    - "val_mask_class"
+                    - "test_mask_class"
                 num_classes: 14
                 task_weight: 1.0
             - link_prediction:
@@ -671,7 +679,7 @@ when you want to produce a graph just for inference.
 Examples
 ~~~~~~~~
 
-OAG-Paper dataset
+Node classification for node type `field` in OAG-Paper dataset
 -----------------
 
 .. code-block:: json
--- a/docs/source/cli/graph-construction/distributed/gsprocessing/input-configuration.rst
+++ b/docs/source/cli/graph-construction/distributed/gsprocessing/input-configuration.rst
@@ -489,9 +489,174 @@ arguments.
         You can find all models in the `Huggingface model repository <https://huggingface.co/models>`_.
       - ``max_seq_length`` (Integer, required): Specifies the maximum number of tokens of the input.
         You can use a length greater than the dataset's longest sentence; or for a safe value choose 128. Make sure to check
-        the model's max suported length when setting this value,
+        the model's max supported length when setting this value.
 
---------------
+
+..  _gsprocessing-multitask-ref:
+
+Creating a graph for multi-task training
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To create a graph for multi-task training, you simply need to
+define a label in your config for each of the labels you want to
+use during training. GSProcessing will generate separate
+train/val/test masks for each of your labels, which
+will be named ``<mask_type>_<label_column>``, e.g.
+if your label columns were ``label_class``, and ``label_reg``
+GSProcessing will generate masks named as ``train_mask_label_class``,
+``train_mask_label_reg`` etc. For link prediction tasks, the masks
+will have ``_lp`` added as a suffix, e.g. ``train_mask_lp``.
+
+After partitioning the data, you can then provide these masks
+in your train YAML file to use during multi-task training.
+
+For details on running multi-task training see
+:doc:`/advanced/multi-task-learning`.
+
+Here we list an example multi-task GSProcessing config for the ACM data described in
+:doc:`/tutorials/use-own-data`, where we prepare a node classification label for the
+``paper`` node type, and a link prediction task on the ``paper,citing,paper`` edge
+type.
+
+.. raw:: html
+
+   <details>
+   <summary><a>Example multi-task GSProcessing config</a></summary>
+
+.. code-block:: json
+
+    {
+    "version": "gsprocessing-v1.0",
+    "graph": {
+        "nodes": [
+            {
+                "data": {
+                    "format": "parquet",
+                    "files": [
+                        "nodes/paper.parquet"
+                    ]
+                },
+                "type": "paper",
+                "column": "node_id",
+                "features": [
+                    {
+                        "column": "feat",
+                        "name": "feat",
+                        "transformation": {
+                            "name": "no-op"
+                        }
+                    }
+                ],
+                "labels": [
+                    {
+                        "column": "label",
+                        "type": "classification",
+                        "split_rate": {
+                            "train": 0.8,
+                            "val": 0.1,
+                            "test": 0.1
+                        }
+                    }
+                ]
+            }
+        ],
+        "edges": [
+            {
+                "data": {
+                    "format": "parquet",
+                    "files": [
+                        "edges/paper_citing_paper.parquet"
+                    ]
+                },
+                "source": {
+                    "column": "source_id",
+                    "type": "paper"
+                },
+                "dest": {
+                    "column": "dest_id",
+                    "type": "paper"
+                },
+                "relation": {
+                    "type": "citing"
+                },
+                "labels": [
+                    {
+                        "column": "",
+                        "type": "link_prediction",
+                        "split_rate": {
+                            "train": 0.8,
+                            "val": 0.1,
+                            "test": 0.1
+                        }
+                    }
+                ]
+            }
+        ]
+    }
+
+.. raw:: html
+
+   </details><p></p>
+
+When using the above GSProcessing config you can then use the following
+train YAML file to run multi-task training:
+
+.. raw:: html
+
+    <details>
+    <summary><a>Example multi-task training YAML</a></summary>
+
+.. code-block:: yaml
+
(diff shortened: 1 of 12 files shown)


docs/source/cli/graph-construction/distributed/gsprocessing/input-configuration.rst
```diff
@@ -489,9 +489,174 @@ arguments.
...
+
+Here we list an example multi-task GSProcessing config for the ACM data described in
+:doc:`/tutorials/use-own-data`, where we prepare a node classification label for the
+``paper`` node type, and a link prediction task on the ``paper,citing,paper`` edge
+type.
+
+.. raw:: html
+
+   <details>
+   <summary><a>Example multi-task GSProcessing config</a></summary>
```
thvasilo, line 524: This tag (and corresponding closing tag below) create collapsible code sections see the rendered example:

image

graphstorm-processing/graphstorm_processing/config/label_config_base.py
```diff
@@ -28,7 +28,6 @@ def __init__(self, config_dict: Dict[str, Any]):
             self._label_column = config_dict["column"]
         else:
             self._label_column = ""
-            assert config_dict["type"] == "link_prediction"
```
thvasilo, line 31: This check is done in the sanity check, no need to check twice.

graphstorm-processing/tests/test_converter.py
```diff
@@ -81,42 +81,31 @@ def test_try_read_invalid_gconstruct_config(converter: GConstructConfigConverter
     with pytest.raises(AssertionError):
         _ = converter.convert_nodes(node_dict["nodes"])
 
-    """Feature Name must exist for multiple feature columns"""
+    # Feature Name must exist for multiple feature columns
```
thvasilo, line 84: Dev information should be in comments, not strings that are parsed

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -1919,6 +1926,35 @@ def multinomial_sample(label_col: str) -> Sequence[int]:
 
         return train_mask_df, val_mask_df, test_mask_df
 
+    @staticmethod
+    def _is_multitask(data_configs: Mapping[str, Sequence[StructureConfig]]) -> bool:
+        """Returns True when the input configuration contains more than one label.
```
jalencato, line 1931: A case is that if users now build a graph with both node classification label and link prediction label, I think our current code will treat them as a multi-task graph construction and build graph with new mask name like ``train_mask_lp``.

However, if user just want to create a graph with both node classification label and link prediction label, and then separately launch node classification task and link prediction task. The design here will meet the backward compatibility problem because the mask name is already changed. In this way we are forcing the user to use multi-task training framework.

jalencato (commented):
Generally speaking, I think there is some backward compatibility problem.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: Yes I just submitted an RFC that raises that: https://github.com/awslabs/graphstorm/issues/997

I think it's an OK feature to drop, I don't know of any user that would create multiple individual labels. And it might be possible to also fix downstream, we can use `str.startswith()` when determining which features to load as masks.

jalencato: Or maybe a more appropriate example is that we create labels on multiple edges for link prediction and we do training on these multiple types of edges which can be a common case for users. Previously we construct a graph with all named by default with `train_mask`, and the training will go to the link prediction training workflow. But now it will not support this case any more. Users need to go to the multi-task flow to do all of these.

thvasilo: Yeah that's a legitimate usecase.

One way we could have both is by modifying [GSgnnData.get_edge_train_set()](https://github.com/thvasilo/graphstorm/blob/ca07f7f2f020202223ae222e9c89d40a2cc7d591/python/graphstorm/dataloading/dataset.py#L764) to use the `train_mask` _prefix_ to retrieve the train masks. This I think ensures backwards compatibility.

docs/source/cli/graph-construction/distributed/gsprocessing/input-configuration.rst
```diff
@@ -489,9 +489,174 @@ arguments.
...
+
+To create a graph for multi-task training, you simply need to
+define a label in your config for each of the labels you want to
+use during training. GSProcessing will generate separate
+train/val/test masks for each of your labels, which
+will be named ``<mask_type>_<label_column>``, e.g.
+if your label columns were ``label_class``, and ``label_reg``
+GSProcessing will generate masks named as ``train_mask_label_class``,
+``train_mask_label_reg`` etc. For link prediction tasks, the masks
+will have ``_lp`` added as a suffix, e.g. ``train_mask_lp``.
```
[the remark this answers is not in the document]
classicsong: If there are two link prediction tasks, how does gsprocess generate the masks?

jalencato (approved):
LGTM

graphstorm-processing/tests/test_dist_heterogenous_loader.py
```diff
@@ -631,8 +667,152 @@ def test_process_edge_labels_link_prediction(
...
+def test_process_edge_labels_multitask(
+    spark: SparkSession, dghl_loader: DistHeterogeneousGraphLoader
+):
+    """Test processing multi-task link prediction and regression edge labels"""
+    edges_df = create_edges_df_num_label(spark, 0)
+    lp_split_rates = {"train": 0.6, "val": 0.2, "test": 0.2}
+    reg_split_rates = {"train": 0.8, "val": 0.1, "test": 0.1}
+
+    lp_mask_names = [
+        "train_mask_lp",
```
classicsong, line 682: You also need to add a test like

lp_mask_names = ["lp_train_mask", "lp_val_mask", "lp_test_mask"]

graphstorm-processing/tests/test_dist_heterogenous_loader.py
```diff
@@ -631,8 +667,152 @@ def test_process_edge_labels_link_prediction(
...
+def test_process_node_labels_multitask(
+    spark: SparkSession, dghl_loader: DistHeterogeneousGraphLoader
+):
+    """Test processing multi-task link prediction and regression edge labels"""
+    nodes_df = create_edges_df_num_label(spark, 0)
+    class_split_rates = {"train": 0.6, "val": 0.3, "test": 0.1}
+    reg_split_rates = {"train": 0.7, "val": 0.2, "test": 0.1}
+
+    class_mask_names = [
+        f"train_mask_{STR_LABEL_COL}",
```
classicsong, line 755: Ditto

docs/source/cli/graph-construction/distributed/gsprocessing/input-configuration.rst
```diff
@@ -501,10 +674,12 @@ in the input configuration, the processed data will not
 include any train/val/test masks. You can use this mode
 when you want to produce a graph just for inference.
 
+--------------
```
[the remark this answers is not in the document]
thvasilo: It's a visual separator from the examples can remove as well.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -1821,32 +1830,32 @@ def _create_split_files(
...
+            mask_dfs = self._create_split_files_split_rates(
+                input_df, label_column, split_rates, seed, mask_field_names
             )
         else:
-            train_mask_df, val_mask_df, test_mask_df = self._create_split_files_custom_split(
-                input_df, custom_split_file
-            )
+            if mask_field_names:
+                raise NotImplementedError(
+                    "Custom split files with custom mask field names currently not supported."
```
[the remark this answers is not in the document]
thvasilo: That's the plan, @jalencato was assigned that task.

thvasilo:
@classicsong updated the PR to fix the support for re-partitioning. In short the changes are:

* In our output metadata, under the "graph_info" key that we fully control, added two keys, "ntype_to_mask_names" and "etype_to_mask_names". These will hold the name of each mask for every node/edge type. By default these are "train_mask", "val_mask", "test_mask".
* The repartition script now will read in those values when transforming data for one node/edge type, and ensure that all mask Parquet files get the proper metadata appended.
    * To maintain backwards compatibility, we use the default "train_mask", "val_mask", "test_mask" for repartition_files when `ntype_to_mask_names`/`ntype_to_mask_names` is missing from `graph_info`.
* The other changes in `graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py` mostly have to do with having to pull most of the code for processing a single edge in its own function, because we reached pylint's limit of 5 nested blocks. This should make the code a bit more readable.

## Testing

To test the above changes we include new/modified unit and integration tests for gsprocessing and gsrepartition:

* For GSProcessing we now use custom mask names like `class_train_mask` as suggested. We perform unit tests on both edge and node processing, and update our integration tests to ensure the expected values are added to the `graph_info` dict.
* For gsprepartition, we modify the input to the integration tests to include the `graph_info` updates, and create additional train/val/test mask files that we ensure get processed as expected by DistPart downstream.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -1417,104 +1451,149 @@ def process_edge_data(self, edge_configs: Sequence[EdgeConfig]) -> Tuple[Dict, D
...
-        logging.info("Finished processing edge features")
-        return edge_data_dict, edges_dict
+                    if self.add_reverse_edges:
+                        # For reverse edges only the label metadata
+                        # (labels + split masks) are relevant.
+                        edge_data_dict[reverse_edge_type] = label_metadata_dicts
+                    self.timers["_process_edge_labels"] += perf_counter() - edge_label_start
+                else:
+                    # Different edges can have different relation types
+                    # TODO: For each rel_type we need get the label and output it separately.
```
classicsong, line 1582: Why we need this TODO? The function is supposed to handle just one edge type. So it is impossible that there are multiple relation types.

graphstorm-processing/graphstorm_processing/repartition_files.py
```diff
@@ -825,27 +835,16 @@ def modify_flat_array_metadata(
...
-    def key_is_mask(key: str):
-        for mask in [
-            "train_mask",
-            "val_mask",
-            "test_mask",
-        ]:
-            if key.startswith(mask):
-                return True
-        return False
+    graph_info: dict = metadata_dict["graph_info"]
```
classicsong, line 838: Is metadata_dict per edge type? or per node type?

classicsong: Do you have a docstring describing the meaning of each field of graph_info?

graphstorm-processing/graphstorm_processing/repartition_files.py
```diff
@@ -926,8 +934,18 @@ def key_is_mask(key: str):
...
                 )
                 repartitioner.modify_metadata_for_flat_arrays(node_label_files)
-                for ntype_data_entry_key, ntype_data_entry_val in ntype_data_dict.items():
-                    if key_is_mask(ntype_data_entry_key):
+                for (
+                    ntype_data_entry_key,
+                    ntype_data_entry_val,
+                ) in ntype_data_dict.items():
+                    if "ntype_to_label_masks" in graph_info:
+                        ntype_label_masks: list[str] = graph_info["ntype_to_label_masks"][type_name]
```
classicsong, line 942: How do you handle the case when there are more than one training mask on a node type?

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: This is a leftover from NeptuneML, where we had the `relname/prefixname` distinction and was planning to support that. We can remove this in a future PR.

graphstorm-processing/graphstorm_processing/repartition_files.py
thvasilo: Part of metadata_dict is structured as the "chunked graph format" of DGL: https://docs.dgl.ai/guide/distributed-preprocessing.html#chunked-graph-format

We add the additional key `graph_info` where we hold a different representation of the graph information in a more readable/verbose format. I'll add some documentation on its contents.

graphstorm-processing/graphstorm_processing/repartition_files.py
thvasilo: During graph construction all the mask names are gathered in a list for each node type. The structure is:


```python
# graph_info
{
    "ntype_to_label_masks": {
        "paper": [
            "class_train_mask",
            "class_val_mask",
            "class_test_mask",
            "reg_train_mask",
            "reg_val_mask",
            "reg_test_mask"
        ]
    }
}
```

and similarly for each edge type.

thvasilo: I'll add this info in graph_info documentation.

classicsong (approved):
LGTM

PR #1001, opened 2024-09-03
Support for launching jobs with GraphBolt enabled (GraphBolt support PR1)

thvasilo:
*Issue #, if available:*

*Description of changes:*

* We introduce support for GraphBolt training and inference, see https://docs.dgl.ai/stochastic_training/index.html for details
* We add a new optional argument `--use-graphbolt` to the configuration. When set to true that sets the dgl.initialize call to include `use_graphbolt=True`, on supported DGL versions.
* For versions of DGL < 2, we print a warning and run regular distributed environment init.

### DIFF
--- a/python/graphstorm/gsf.py
+++ b/python/graphstorm/gsf.py
@@ -18,6 +18,8 @@
 
 import os
 import logging
+from importlib.metadata import version
+
 import numpy as np
 import dgl
 import torch as th
@@ -111,7 +113,13 @@ from .inference import (GSgnnLinkPredictionInferrer,
 
 from .tracker import get_task_tracker_class
 
-def initialize(ip_config=None, backend='gloo', local_rank=0, use_wholegraph=False):
+def initialize(
+        ip_config=None,
+        backend='gloo',
+        local_rank=0,
+        use_wholegraph=False,
+        use_graphbolt=False,
+    ):
     """ Initialize distributed training and inference context. For GraphStorm Standalone mode,
     no argument is needed. For Distributed mode, users need to provide an IP address list file.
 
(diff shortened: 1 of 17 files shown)


thvasilo:
Depends on https://github.com/awslabs/graphstorm/pull/1002

python/graphstorm/gsf.py
```diff
@@ -141,10 +149,32 @@ def initialize(ip_config=None, backend='gloo', local_rank=0, use_wholegraph=Fals
...
+            ip_config,
+            net_type='socket',
+            use_graphbolt=use_graphbolt,
+        )
+    else:
+        if use_graphbolt:
+            logging.warning(
+                (
+                    "use_graphbolt was 'true' but but DGL version was %s. "
+                    "GraphBolt requires DGL version >= 2.x."
```
classicsong, line 168: 2.4?

python/graphstorm/gsf.py
```diff
@@ -141,10 +149,32 @@ def initialize(ip_config=None, backend='gloo', local_rank=0, use_wholegraph=Fals
...
+                ),
+                dgl_version
+                )
+        # We need to use socket for communication in DGL 0.8. The tensorpipe backend has a bug.
```
classicsong, line 172: Why does the comment mean?

python/graphstorm/gsf.py
```diff
@@ -141,10 +149,32 @@ def initialize(ip_config=None, backend='gloo', local_rank=0, use_wholegraph=Fals
...
+        # This problem will be fixed in the future.
```
classicsong, line 173: Who is going to fix the issue? Can you check with the DGL team?

python/graphstorm/gsf.py
thvasilo: This was pre-existing I just moved it. If we don't need it any more I'll remove it.

python/graphstorm/gsf.py
thvasilo: Same as above, I didn't write this comment, but looking at the DGL code, there's only support for socket anyway, so not much difference.

classicsong: Let's remove it.

classicsong (approved):

PR #1011, opened 2024-09-05
[GraphBolt] Add GraphBolt graph construction for GConstruct and GPartition (GraphBolt PR2)

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Add `--use-graphbolt` argument to GConstruct and GSPartition entry points. When set to `true`, we convert the graph data to GraphBolt format after partitioning.
* Add new `gpartition_test.sh` set of integration tests for GSPartition entry point. 
* Add new tests/end2end-tests/graphbolt-gs-integration/graphbolt-graph-construction.sh set of integration tests for graph construction with GraphBolt enabled. Requires DGL >= 2.x.
  * To accommodate this, we we modify `tests/end2end-tests/data_gen/process_movielens.py` so that it also produces a JSON file that can be used as input the the DistPart DGL pipeline.
* To parse and handle different DGL versions we use `importlib.metadata` and the `packaging` modules. We introduce an explicit dependency on `packaging` to GSF. This shouldn't change the user environment, as we already had a transitive dependency on `packaging` through the `transformers` lib.

### DIFF
--- a/docs/source/cli/graph-construction/distributed/gspartition/ec2-clusters.rst
+++ b/docs/source/cli/graph-construction/distributed/gspartition/ec2-clusters.rst
@@ -129,4 +129,5 @@ The arguments that ``graphstorm.gpartition.dist_partition_graph`` accepts are th
 * ``--ip-config str``: A file storing a list of IPs, one line for each instance of the partition cluster.
 * ``--partition-assignment-only``: Only generate partition assignments for nodes, the process will not build the partitioned DGL graph.
 * ``--logging-level str``: The logging level. The possible values: debug, info, warning, error. The default value is info. (default: info)
-* ``--use-graphbolt "true"/"false"``: ``New in v0.4``. Whether to convert the partitioned data to the GraphBolt format after creating the DistDGL graph. (default: ``"false"``)
+* ``--use-graphbolt "true"/"false"``: ``New in v0.4``. Whether to convert the partitioned data to the GraphBolt format after creating the DistDGL graph.
+  Requires installed DGL version to be at least ``2.1.0``. (default: ``"false"``)
--- a/docs/source/cli/graph-construction/distributed/gspartition/ec2-clusters.rst
+++ b/docs/source/cli/graph-construction/distributed/gspartition/ec2-clusters.rst
@@ -121,7 +121,6 @@ The arguments that ``graphstorm.gpartition.dist_partition_graph`` accepts are th
 * ``--input-path str``: Path to input DGL chunked data directory. (required). The format for the data
   is available at : https://docs.dgl.ai/guide/distributed-preprocessing.html#chunked-graph-format
 * ``--output-path str``: Path to store the partitioned data. (required)
-* ``--output-path str``: Path to store the partitioned data under. (required)
 * ``--num-parts int``: Number of partitions to generate. (required)
 * ``--metadata-filename str``: Name for the chunked DGL data metadata file. (default: ``metadata.json``)
 * ``--ssh-port int``: SSH Port. (default: 22)
--- a/docs/source/cli/graph-construction/distributed/gspartition/ec2-clusters.rst
+++ b/docs/source/cli/graph-construction/distributed/gspartition/ec2-clusters.rst
@@ -103,7 +103,7 @@ Now we can ssh into the **leader node** of the EC2 cluster, and start GSPartitio
 
 .. code:: bash
 
-    python3 -m graphstorm.gpartition.dist_partition_graph
+    python3 -m graphstorm.gpartition.dist_partition_graph \
         --input-path ${LOCAL_INPUT_DATAPATH} \
         --metadata-filename ${METADATA_FILE} \
         --output-path ${LOCAL_OUTPUT_DATAPATH} \
@@ -116,4 +116,16 @@ Now we can ssh into the **leader node** of the EC2 cluster, and start GSPartitio
     2. The number of instances in the cluster should be equal to ``NUM_PARTITIONS``.
     3. For users who only want to generate partition assignments instead of the partitioned DGL graph, please add ``--partition-assignment-only`` flag.
 
-Currently we support both ``random`` and ``parmetis`` as the partitioning algorithm for EC2 clusters.
+The arguments that ``graphstorm.gpartition.dist_partition_graph`` accepts are the following:
+
+* ``--input-path str``: Path to input DGL chunked data directory. (required)
+* ``--output-path str``: Path to store the partitioned data under. (required)
+* ``--num-parts int``: Number of partitions to generate. (required)
+* ``--metadata-filename str``: Name for the chunked DGL data metadata file. (default: ``metadata.json``)
+* ``--ssh-port int``: SSH Port. (default: 22)
+* ``--dgl-tool-path str``: The path to `dgl/tools` code. (default: ``/root/dgl/tools``)
+* ``--partition-algorithm str``: Partition algorithm to use. (choices: ``random``, ``parmetis``, default: ``random``)
+* ``--ip-config str``: A file storing a list of IPs, one line for each instance of the partition cluster.
+* ``--partition-assignment-only``: Only generate partition assignments for nodes, the process will not build the partitioned DGL graph.
+* ``--logging-level str``: The logging level. The possible values: debug, info, warning, error. The default value is info. (default: info)
+* ``--use-graphbolt "true"/"false"``: ``New in v0.4``. Whether to convert the partitioned data to the GraphBolt format after creating the DistDGL graph. (default: ``"false"``)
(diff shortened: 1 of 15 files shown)


thvasilo:
Blocked on having a CI Python env with DGL >= 2.x.

python/graphstorm/gconstruct/utils.py
```diff
@@ -997,12 +1005,36 @@ def partition_graph(g, node_data, edge_data, graph_name, num_partitions, output_
...
+    dgl_version = version.parse(dgl_version_str)
+    part_position_args = [
+        g, graph_name, num_partitions, output_dir,
+    ]
+    part_kwargs = {
+        'part_method': part_method,
+        'balance_ntypes': balance_ntypes,
+        'balance_edges': True,
+        'return_mapping': save_mapping,
+    }
```
thvasilo, line 1022: We use this way to call `dgl.distributed.partition_graph` otherwise pylint errors out because the `partition_graph` function signature is different between DGL 1.x and 2.x

python/graphstorm/gpartition/dist_partition_graph.py
```diff
@@ -154,6 +158,26 @@ def main():
...
+            # TODO: Use dgl.distributed.partition.gb_convert_single_dgl_partition()
+            # to run distributed conversion
+            logging.info("Converting partitions to GraphBolt format")
+            dgl.distributed.dgl_partition_to_graphbolt(
+                os.path.join(output_path, "dist_graph", "metadata.json"),
+                store_eids=True,
+                graph_formats="coo",
+            )
+        else:
+            logging.warning(
```
classicsong, line 173: ditto

python/graphstorm/gpartition/dist_partition_graph.py
```diff
@@ -154,6 +158,26 @@ def main():
 
         logging.info("DGL graph building took %f sec", part_end - time.time())
 
+    if args.use_graphbolt:
+        dgl_version = importlib.metadata.version('dgl')
+        if version.parse(dgl_version) >= version.parse("2.0.0"):
+            # TODO: Use dgl.distributed.partition.gb_convert_single_dgl_partition()
```
classicsong, line 164: Why not use now? Can you also add the reason in the comment?

tests/end2end-tests/data_gen/process_movielens.py
```diff
@@ -139,4 +147,46 @@ def write_data_parquet(data, data_file):
 # generate synthetic user data with label
 user_labels = np.random.randint(11, size=feat.shape[0])
 user_data = {'id': user['id'].values, 'feat': feat, 'occupation': user['occupation'], 'label': user_labels}
-write_data_parquet(user_data, '/data/ml-100k/users_with_synthetic_labels.parquet')
\ No newline at end of file
+write_data_parquet(user_data, '/data/ml-100k/users_with_synthetic_labels.parquet')
+
+
+user_feat_only = {'feat': feat}
```
classicsong, line 153: Are the codes below for dist partition test?
Why we store features and labels separately here?

tests/end2end-tests/graphbolt-gs-integration/graphbolt-graph-construction.sh
```diff
@@ -0,0 +1,120 @@
...
+-o, --output-path   Path under which the converted data will be created.
+
+EOF
+  exit
+}
+
+# Parse command-line arguments
+parse_params() {
+    # Default values for input and output paths
+    INPUT_PATH="/storage/ml-100k"
```
classicsong, line 25: The ml-100k is stored in `storage/`?

python/graphstorm/gpartition/dist_partition_graph.py
thvasilo: It's a non-trivial implementation (we need to write pytorch.distributed/MPI-style code) that I'd rather leave for later.

tests/end2end-tests/graphbolt-gs-integration/graphbolt-graph-construction.sh
thvasilo: That's the hardcoded assumption in all our tests, the raw dataset is baked into the CI image at that location, which is then processed and output to `/data/ml-100k/`, see https://github.com/awslabs/graphstorm/blob/f7796f29402948d4e477bd8adb491a41db5195a3/tests/end2end-tests/create_data.sh#L6

tests/end2end-tests/data_gen/process_movielens.py
thvasilo: Yes, we need to store them separately as that's the expected input format for the DistPart pipeline: https://docs.dgl.ai/guide/distributed-preprocessing.html#chunked-graph-format

docs/source/cli/graph-construction/distributed/gspartition/ec2-clusters.rst
```diff
@@ -118,7 +118,9 @@ Now we can ssh into the **leader node** of the EC2 cluster, and start GSPartitio
 
 The arguments that ``graphstorm.gpartition.dist_partition_graph`` accepts are the following:
 
-* ``--input-path str``: Path to input DGL chunked data directory. (required)
+* ``--input-path str``: Path to input DGL chunked data directory. (required). The format for the data
+  is available at : https://docs.dgl.ai/guide/distributed-preprocessing.html#chunked-graph-format
+* ``--output-path str``: Path to store the partitioned data. (required)
 * ``--output-path str``: Path to store the partitioned data under. (required)
```
classicsong, line 124: You have two ``--output-path str``

python/graphstorm/gpartition/dist_partition_graph.py
classicsong: I c.

python/graphstorm/gpartition/dist_partition_graph.py
```diff
@@ -154,6 +158,23 @@ def main():
 
         logging.info("DGL graph building took %f sec", part_end - time.time())
 
+    if args.use_graphbolt:
+        dgl_version = importlib.metadata.version('dgl')
+        if version.parse(dgl_version) >= version.parse("2.0.0"):
+            # TODO: Implement distributed conversion using
+            # dgl.distributed.partition.gb_convert_single_dgl_partition()
+            logging.info("Converting partitions to GraphBolt format")
+            dgl.distributed.dgl_partition_to_graphbolt(
```
classicsong, line 167: So, do you assume the master node is going to convert partitioned graphs into graphbolt format?
Can you add a comment here that the current code assume there is a shared filesystem that the master node can access the partition result?

classicsong: Another question, what will happen if dist_partition_graph is running on SageMaker?

thvasilo: Sure, I'll add a comment here to make that clear.

Generally speaking dist_partition_graph always assumes a shared FS, the way we use it in SM is by "emulating" a shared FS by pulling data locally and relying on consistent paths and file layout so that local files accesses behave like shared ones.

We haven't added SageMaker support for this PR, for SageMaker support it will be easier to use the distributed implementation, where each worker will work on its local partition. That will come in an upcoming PR.

classicsong (approved):

python/graphstorm/gconstruct/utils.py
```diff
@@ -997,12 +1005,33 @@ def partition_graph(g, node_data, edge_data, graph_name, num_partitions, output_
...
+        "graph_name": graph_name,
+        "num_parts": num_partitions,
+        "out_path": output_dir,
+        'part_method': part_method,
+        'balance_ntypes': balance_ntypes,
+        'balance_edges': True,
+        'return_mapping': save_mapping,
+    }
+
+    if dgl_version >= version.parse("2.0.0"):
```
jalencato, line 1024: Do we need to check for 2.3.0+ or just 2.0.0?

tests/end2end-tests/data_process/gpartition_test.sh
```diff
@@ -0,0 +1,53 @@
...
+    rm -rf "${OUTPUT_PATH}"
+  fi
+}
+
+# Ensure test data have been generated
+if [ ! -f "${INPUT_PATH}/chunked_graph_meta.json" ]; then
+    python3 "$GS_HOME/tests/end2end-tests/data_gen/process_movielens.py"
+fi
+
+echo "********* Test partition creation with DistDGL graph format ********"
```
jalencato, line 30: Can we add a test for --partition-assignment-only？

jalencato (commented):
Two small comments

python/graphstorm/gconstruct/utils.py
thvasilo: Good catch, seems like the new argument was introduced in 2.1.0. I'll update the checks accordingly

jalencato (approved):
LGTM

PR #1015, opened 2024-09-06
[Bugfix] Excluding valid/test edges during link prediction training in multi-task learning setting

classicsong:
*Issue #, if available:*
The training dataloader for link prediction use the default "train_mask" as the edge mask for gnn embedding computation. However, this assumption does not hold in multi-task learning.

*Description of changes:*


thvasilo (commented):
LGTM, this should ensure the mask names are correct for each task

thvasilo (approved):

PR #1022, opened 2024-09-12
[Bugfix] Fix the bug when the first fanout of the fanout list is -1

classicsong:
*Issue #, if available:*
When the first fanout of the fanout list is -1, the program will fail.

*Description of changes:*
Fix the bug


jalencato (approved):
LGTM

PR #1030, opened 2024-09-19
[Bug Fix] Fix Backward Compatibility Problem for DGL 2.0.0+

jalencato:
*Issue #, if available:*

*Description of changes:*


python/graphstorm/dataloading/dataloading.py
```diff
@@ -124,6 +124,8 @@ def __init__(self, sampler, dataset, construct_feat_ntype, construct_feat_fanout
         self._sampler = sampler
         self._construct_feat_sampler = _ReconstructedNeighborSampler(
                 dataset, construct_feat_ntype, construct_feat_fanout)
+        # Temporary fix for DGL 2.0.0+ Compatabilities
+        self.prob = None
```
jalencato, line 128: DGL fix this problem for 2.4.0, but as we mainly recommend users to use 2.3.0, better to use this one.

python/graphstorm/inference/np_infer.py
```diff
@@ -157,8 +158,10 @@ def infer(self, loader, save_embed_path, save_prediction_path=None,
                     if node_id_mapping_file else None
             shuffled_preds = {}
             for ntype, pred in preds.items():
-                assert ntype in infer_ntypes, \
-                    f"{ntype} is not in the set of evaluation ntypes {infer_ntypes}"
+                if ntype not in infer_ntypes:
+                    warnings.warn(f"{ntype} is not in the evaluation ntypes {infer_ntypes}, "
+                                  f"will not do the remapping")
+                    continue
```
jalencato, line 164: The DGLBlock problem as mentioned offline.

tests/end2end-tests/graphstorm-ec/test.sh
```diff
@@ -28,7 +28,7 @@ echo "Test GraphStorm edge classification"
 date
 
 echo "**************standalone"
-python3 $GS_HOME/python/graphstorm/run/gsgnn_ep/gsgnn_ep.py --part-config /data/movielen_100k_ec_1p_4t/movie-lens-100k.json --cf $GS_HOME/training_scripts/gsgnn_ep/ml_ec.yaml
```
jalencato, line 31: DGL 2.0.0+ does not initialize `th.distributed.init_process_group(backend=backend)`, and this test does not make sense as we never claim to support the standalone mode with gsgnn_*.py.

python/graphstorm/inference/np_infer.py
zhjwy9343: Because here we `continue`, should the behavior is not remapping either saving ntype prediction?
If yes, may change the warning message to `will not do the remapping and saving {ntype} node predictions`.

zhjwy9343 (approved):
Except for the warning message, LGTM.

python/graphstorm/inference/np_infer.py
jalencato: Good callout. Make a change on that.

tests/end2end-tests/custom-gnn/train.py
```diff
@@ -108,8 +108,20 @@ def main(args):
...
+                default=0,
+                help="The rank of the trainer.",
+                )
+    else:
+        argparser.add_argument(
+                "--local_rank",
+                type=int,
+                default=0,
+                help="The rank of the trainer.",
+                )
```
thvasilo, line 124: No need to check for PyTorch version, you can just use:

```
parser.add_argument("--local-rank", "--local_rank", type=int, ...)
```

see https://pytorch.org/docs/stable/distributed.html#launch-utility

python/graphstorm/model/node_gnn.py
```diff
@@ -316,8 +317,11 @@ def node_mini_batch_gnn_predict(model, loader, return_proba=True, return_label=F
                               iter_l, max_num_batch, time.time() - iter_start)
 
     model.train()
-    for ntype, ntype_pred in preds.items():
-        preds[ntype] = th.cat(ntype_pred)
+    # MFG for DGL 2.0.0+ return all node and edge type
```
classicsong, line 320: Can you also check ``node_mini_batch_predict``?

tests/end2end-tests/custom-gnn/train.py
```diff
@@ -108,7 +108,7 @@ def main(args):
                            help="The name of the label.")
     argparser.add_argument("--num-classes", type=int, required=True,
                            help="The number of classes.")
-    argparser.add_argument("--local_rank", type=int,
+    argparser.add_argument("--local-rank", "--local_rank", type=int,
```
classicsong, line 111: Typo?

tests/end2end-tests/graphstorm-ec/test.sh
jalencato: Yes. And it was only the previous use case before we update the standalone script. So deprecate the test here.

tests/end2end-tests/custom-gnn/train.py
jalencato: Nope. local-rank is for torch >= 2.0.0, local_rank is for lower version torch.

python/graphstorm/model/node_gnn.py
jalencato: I manually check that function and do not require any other change because it only requires decoder. Only the GNN related predict function needs some modifications.

tests/end2end-tests/custom-gnn/train.py
classicsong: I c.

classicsong: Can you add a comment for that?

tests/end2end-tests/graphstorm-ec/test.sh
classicsong: Can you also update L30?

tests/end2end-tests/graphstorm-lp/test.sh
```diff
@@ -28,7 +28,7 @@ echo "Test GraphStorm link prediction"
 date
 
 echo "**************standalone"
-python3 $GS_HOME/python/graphstorm/run/gsgnn_lp/gsgnn_lp.py --part-config /data/movielen_100k_lp_train_val_1p_4t/movie-lens-100k.json --cf $GS_HOME/training_scripts/gsgnn_lp/ml_lp.yaml
+python3 -m graphstorm.run.gs_link_prediction --workspace $GS_HOME/training_scripts/gsgnn_lp/ --num-trainers $NUM_TRAINERS --num-servers 1 --num-samplers 0 --part-config /data/movielen_100k_lp_train_val_1p_4t/movie-lens-100k.json --cf ml_lp.yaml
```
classicsong, line 31: Ditto.

classicsong (approved):
Some minor comments.
LGTM.

PR #1034, opened 2024-09-20
[Bugfix] Link prediction contrastive loss does not work with edge weight for lp loss

classicsong:
Fix the bug by disabling edge weight for lp loss when the loss function is contrastive loss.

### DIFF
--- a/python/graphstorm/config/argument.py
+++ b/python/graphstorm/config/argument.py
@@ -2348,7 +2348,7 @@ class GSConfig:
                 "Edge weight for loss only works with link prediction"
 
             if self.lp_loss_func in [ BUILTIN_LP_LOSS_CONTRASTIVELOSS]:
-                logging.warning("p_edge_weight_for_loss does not work with "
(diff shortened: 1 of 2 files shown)


python/graphstorm/config/argument.py
```diff
@@ -2346,6 +2346,14 @@ def lp_edge_weight_for_loss(self):
         if hasattr(self, "_lp_edge_weight_for_loss"):
             assert self.task_type == BUILTIN_TASK_LINK_PREDICTION, \
                 "Edge weight for loss only works with link prediction"
+
+            if self.lp_loss_func in [ BUILTIN_LP_LOSS_CONTRASTIVELOSS]:
+                logging.warning("p_edge_weight_for_loss does not work with "
```
zhjwy9343, line 2351:
```suggestion
                logging.warning("lp_edge_weight_for_loss does not work with "
```

zhjwy9343 (approved):
Except for one typo, LGTM

PR #1039, opened 2024-09-24
[GSProcessing] PR2 - Custom Data Split for MT 

jalencato:
*Issue #, if available:*

*Description of changes:*

* Fix custom data split orderring problem.
* Scale test to 1000.
* Try with custom data split on 1M nodes and 100 masked nodes. Spend 10min on one instance.
 

graphstorm-processing/tests/test_dist_heterogenous_loader.py
thvasilo, line 1: For tests can we try creating datasets with at least 10k data points? I want to at least have a non-trivial dataset to check against.

graphstorm-processing/tests/test_dist_heterogenous_loader.py
```diff
@@ -1033,3 +1033,119 @@ def test_edge_custom_label(spark, dghl_loader: DistHeterogeneousGraphLoader, tmp
     assert train_total_ones == 3
     assert val_total_ones == 3
     assert test_total_ones == 3
+
+
+def test_node_custom_label_multitask(spark, dghl_loader: DistHeterogeneousGraphLoader, tmp_path):
```
thvasilo, line 1038: Are we testing for corrected order somewhere?

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -2083,7 +2088,23 @@ def _create_split_files_custom_split(
 
         # custom node/edge label
         # create custom mask dataframe for one of the types: train, val, test
-        def process_custom_mask_df(input_df: DataFrame, split_file: CustomSplit, mask_type: str):
+        def process_custom_mask_df(
+            input_df: DataFrame, split_file: CustomSplit, mask_name: str, mask_type: str
+        ):
```
thvasilo, line 2093: We add `mask_name` before the `mask_type` arg, any reason to not add it to as the last argument?

thvasilo: Can we also add a docstring for this function?

graphstorm-processing/tests/test_dist_heterogenous_loader.py
```diff
@@ -1033,3 +1033,119 @@ def test_edge_custom_label(spark, dghl_loader: DistHeterogeneousGraphLoader, tmp
...
+    test_df = spark.createDataFrame(
+        [(i, j) for i in range(3, 4) for j in range(11, 14)],
+        ["mask_src_id", "mask_dst_id"],
+    )
+
+    class_mask_names = [
+        f"custom_split_train_mask",
+        f"custom_split_val_mask",
+        f"custom_split_test_mask",
+    ]
```
thvasilo, line 1115: f-strings have no var here

graphstorm-processing/tests/test_dist_heterogenous_loader.py
thvasilo: Where does `tmp_path` get its value?

graphstorm-processing/tests/test_dist_heterogenous_loader.py
```diff
@@ -1033,3 +1033,119 @@ def test_edge_custom_label(spark, dghl_loader: DistHeterogeneousGraphLoader, tmp
...
+    test_df.repartition(1).write.parquet(f"{tmp_path}/test.parquet")
+    class_mask_names = [
+        f"custom_split_train_mask",
+        f"custom_split_val_mask",
+        f"custom_split_test_mask",
+    ]
+    config_dict = {
+        "column": "orig",
+        "type": "classification",
+        "split_rate": {"train": 0.8, "val": 0.1, "test": 0.1},
```
thvasilo, line 1060: Should probably print a warning when both split_rate and custom filenames are provided, noting we'll use the custom files and not do a random split.

graphstorm-processing/tests/test_dist_heterogenous_loader.py
```diff
@@ -1033,3 +1033,119 @@ def test_edge_custom_label(spark, dghl_loader: DistHeterogeneousGraphLoader, tmp
...
+
+
+def test_edge_custom_label_multitask(spark, dghl_loader: DistHeterogeneousGraphLoader, tmp_path):
+    """Test using custom label splits for edges"""
+    data = [(i, j) for i in range(1, 4) for j in range(11, 14)]
+    # Create DataFrame
+    edges_df = spark.createDataFrame(data, ["src_str_id", "dst_str_id"])
+
+    train_df = spark.createDataFrame(
+        [(i, j) for i in range(1, 2) for j in range(11, 14)],
```
thvasilo, line 1099: The ranges where `range(x, y)` and `y = x+1` produce just a single value and can just be written as `[x]`. Is this the intention?

jalencato: Yes. We just want the first line to be used for train mask.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
jalencato: I just think logically mask_name should refer as a static value here, so I put it before the mask_type. In this case we will iterate different mask_type for train/val/test so I put it at last.

graphstorm-processing/tests/test_dist_heterogenous_loader.py
jalencato: Built-in value for pytest, check here: https://docs.pytest.org/en/stable/how-to/tmp_path.html

graphstorm-processing/tests/test_dist_heterogenous_loader.py
jalencato: I can add a comment here.

thvasilo: I mean during the parsing of the config, the expectation is to have only one of these two, so should print a warning or error out if we don't want to guess the user's intent.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -2117,10 +2142,12 @@ def process_custom_mask_df(input_df: DataFrame, split_file: CustomSplit, mask_ty
...
                     col(split_file.mask_columns[1]).alias(f"custom_{mask_type}_mask_dst"),
                 )
+                input_df_id = create_mapping(input_df)
                 join_condition = (
-                    input_df["src_str_id"] == custom_mask_df[f"custom_{mask_type}_mask_src"]
-                ) & (input_df["dst_str_id"] == custom_mask_df[f"custom_{mask_type}_mask_dst"])
-                mask_df = input_df.join(custom_mask_df, join_condition, "left_outer")
+                    input_df_id["src_str_id"] == custom_mask_df[f"custom_{mask_type}_mask_src"]
+                ) & (input_df_id["dst_str_id"] == custom_mask_df[f"custom_{mask_type}_mask_dst"])
+                mask_df = input_df_id.join(custom_mask_df, join_condition, "left_outer")
```
thvasilo, line 2149: The join here happens after the graph structure has been written to disk right? Then the order between the structure and masks will not match.

jalencato: Not yet. Because the next line we will do sorting to maintain the order.

graphstorm-processing/tests/test_dist_heterogenous_loader.py
jalencato: I thought we have done that before. I just check our code again but did not find relative code. I will add it.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: Edge structure is written to disk here: https://github.com/awslabs/graphstorm/blob/2fd1511a239f3530cdbde3ad3d4ec7b202e63a50/graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py#L1535-L1537 and labels are processed later here: https://github.com/awslabs/graphstorm/blob/2fd1511a239f3530cdbde3ad3d4ec7b202e63a50/graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py#L1575-L1581

I guess that we add the additional ID column before the join, within the mask processing, and re-order according to that, and that will guarantee the original order (same as was used when writing the edge structure) is maintained? Is my understanding correct?

jalencato: Yes. Correct.

graphstorm-processing/tests/test_dist_heterogenous_loader.py
jalencato: 10K maybe a little bit too large for the CI, but I scale it to 1000, which will not be a non-trivial case here.

thvasilo, line 1: The tests here only verify that the masks have the correct number of items in each train/val/test list, can we add an additional feature column and ensure that and the nodes/edges have the same (expected) order?

jalencato: We did that here actually. Since train_mask_df/val_mask_df/test_mask_df is a single column dataframe, we assign 0/1 to each node/edge for if it is in the train/val/test df. 

For the node test, we assume first 1000 nodes are in the training set, following 100 are validation set and final 100 are test set. We verified if they all of them are assigned correctly here by first selecting them out and then verifying it. https://github.com/awslabs/graphstorm/pull/1039/files#diff-1e1957c9210c75ec052d8e3127bdf1ccca193d8c9e4cc10b195a4f4258127300R1096-R1109

For the edge test, we select 10 edges each for training/validation/test, and check if they are assigned to the same edges. Check here: https://github.com/awslabs/graphstorm/pull/1039/files#diff-1e1957c9210c75ec052d8e3127bdf1ccca193d8c9e4cc10b195a4f4258127300R1177-R1190

graphstorm-processing/tests/test_dist_heterogenous_loader.py
```diff
@@ -1033,3 +1033,158 @@ def test_edge_custom_label(spark, dghl_loader: DistHeterogeneousGraphLoader, tmp
...
+        "column": "",
+        "type": "link_prediction",
+        "custom_split_filenames": {
+            "train": f"{tmp_path}/train.parquet",
+            "valid": f"{tmp_path}/val.parquet",
+            "test": f"{tmp_path}/test.parquet",
+            "column": ["mask_src_id", "mask_dst_id"],
+        },
+        "mask_field_names": class_mask_names,
+    }
```
thvasilo, line 1149: This config only includes a single label right? Can we use a multi-label config instead?

jalencato: I see the point. I have added the multiple config for both node and edge config.

thvasilo (approved):
LGTM now

PR #1043, opened 2024-09-25
[New Feature] New evaluator to support MRR and Hits metrics simultaneously for link prediction.

Oxfordblue7:
*Issue #, if available:*

*Description of changes:*
New LP evaluator to support both MRR and Hits metrics.


python/graphstorm/eval/evaluator.py
```diff
@@ -830,6 +830,298 @@ def compute_score(self, pred, labels, train=True):
 
         return scores
 
+class GSgnnLPEvaluator(GSgnnBaseEvaluator, GSgnnLPRankingEvalInterface):
```
thvasilo, line 833: Would it be possible to have a single class for LP evaluation that is a super-set of the individual MRR and Hits@ classes? 

Seems like their constructors are very similar so we could have a single class that performs all the calculations instead of 3? Thinking about future maintenance otherwise we'd add a whole new class for each new metric?

Oxfordblue7: This is right the super-set of individual metric classes, and can be used for all new metrics as well. I just left the previous MRR and Hits evaluators there.

If you all agree, I can remove those single metric evaluators.

classicsong: Do you mean have a single class for GSgnnLPEvaluator and GSgnnPerEtypeLPEvaluator?

classicsong: I think GSgnnLPEvaluator is going to replace the GSgnnMrrLPEvaluator and GSgnnHitsLPEvaluator.

Can we add a deprecation note for GSgnnMrrLPEvaluator and GSgnnHitsLPEvaluator?

python/graphstorm/gsf.py
```diff
@@ -1144,3 +1149,62 @@ def create_evaluator(task_info):
...
             config.early_stop_strategy)
     return None
+
+def create_lp_evaluator(config):
+    """ Create LP specific evaluator.
+
+        Parameters
+        ----------
+        config: GSConfig
+            Configuration.
```
classicsong, line 1159: Docstring for Returns.

python/graphstorm/gsf.py
```diff
@@ -1144,3 +1149,62 @@ def create_evaluator(task_info):
...
+    """
+    assert all((x.startswith(SUPPORTED_HIT_AT_METRICS) or x == 'mrr') for x in
+               config.eval_metric), (
+        "Invalid LP evaluation metrics. "
+        "GraphStorm only supports MRR and Hit@K metrics for link prediction.")
+
+    if config.report_eval_per_type:
+        if len(config.eval_metric) == 1 and config.eval_metric[0] == 'mrr':
+            return GSgnnPerEtypeMrrLPEvaluator(eval_frequency=config.eval_frequency,
```
classicsong, line 1168: Can we use GSgnnPerEtypeLPEvaluator instead?

python/graphstorm/eval/evaluator.py
thvasilo: > Do you mean have a single class for GSgnnLPEvaluator and GSgnnPerEtypeLPEvaluator?

If possible yes, their interface seems to be the same, the only thing that changes is the set of metrics they calculate?

python/graphstorm/gsf.py
```diff
@@ -1144,3 +1149,62 @@ def create_evaluator(task_info):
...
+                                   major_etype=config.model_select_etype,
+                                   use_early_stop=config.use_early_stop,
+                                   early_stop_burnin_rounds=config.early_stop_burnin_rounds,
+                                   early_stop_rounds=config.early_stop_rounds,
+                                   early_stop_strategy=config.early_stop_strategy)
+        elif 'mrr' not in config.eval_metric:
+            return GSgnnPerEtypeHitsLPEvaluator(eval_frequency=config.eval_frequency,
```
classicsong, line 1175: ditto

python/graphstorm/eval/evaluator.py
Oxfordblue7: Yes. I will add deprecation note for MRR and Hits evaluators.

python/graphstorm/gsf.py
Oxfordblue7: Yes, GSgnnPerEtypeMrrLPEvaluator and GSgnnPerEtypeHitsLPEvaluator deprecated.

python/graphstorm/eval/evaluator.py
```diff
@@ -830,15 +830,319 @@ def compute_score(self, pred, labels, train=True):
...
+                return_metric = metric_val / get_world_size()
+                if metric_key not in return_metrics:
+                    return_metrics[metric_key] = {}
+                return_metrics[metric_key][etype] = return_metric.item()
+        return return_metrics
+
 class GSgnnMrrLPEvaluator(GSgnnBaseEvaluator, GSgnnLPRankingEvalInterface):
-    """ Evaluator for Link Prediction tasks using ``mrr`` as metric.
+    """
+    .. deprecated:: 0.4
```
thvasilo, line 1135: Let's also add deprecation warnings in the class' `__init__` function, see this example in the numpy codebase:

https://github.com/numpy/numpy/blob/9c5162192639937c53fbbbba74391d9e80df485b/numpy/fft/helper.py#L8-L15

python/graphstorm/eval/evaluator.py
```diff
@@ -830,15 +830,319 @@ def compute_score(self, pred, labels, train=True):
...
+        for metric, metric_val in metrics.items():
+            return_metric = metric_val / get_world_size()
+            return_metrics[metric] = return_metric.item()
+
+        return return_metrics
+
+class GSgnnPerEtypeLPEvaluator(GSgnnBaseEvaluator, GSgnnLPRankingEvalInterface):
+    """
+    .. versionadded:: 0.4
+        The :py:class:`GSgnnPerEtypeLPEvaluator`.
```
thvasilo, line 978: Docstring annotations should go to the bottom of the docstring

tests/unit-tests/test_evaluator.py
```diff
@@ -645,6 +663,325 @@ def check_evaluate_infer(mock_compute_score):
 
     th.distributed.destroy_process_group()
 
+def test_per_etype_lp_evaluation():
```
thvasilo, line 666: Could we add comments to describe what each stage of the tests are doing?

python/graphstorm/model/edge_decoder.py
```diff
@@ -1116,29 +1116,33 @@ def get_relembs(self):
         return self._w_relation.weight, self.etype2rid
 
 class LinkPredictRotatEDecoder(LinkPredictMultiRelationLearnableDecoder):
-    r""" Decoder for link prediction using the RotatE as the score function.
+    r"""
+    .. versionadded:: 0.4
```
classicsong, line 1120: Let's separate the docstring update PR with the New evaluator PR.

tests/unit-tests/test_evaluator.py
```diff
@@ -645,6 +663,333 @@ def check_evaluate_infer(mock_compute_score):
...
+                              eval_metric_list=["hit_at_1", "mrr", "hit_at_10"],
+                              use_early_stop=config.use_early_stop)
+    assert lp.do_eval(120, epoch_end=True) is True
+    assert lp.do_eval(200) is True
+    assert lp.do_eval(0) is True
+    assert lp.do_eval(1) is False
+
+    config3 = Dummy({
+            "eval_frequency": 0,
+            "eval_metric_list": ["hit_at_1", "mrr", "hit_at_10"],
```
classicsong, line 978: this is not used

classicsong (commented):
Overall, it looks good.

python/graphstorm/model/edge_decoder.py
Oxfordblue7: will submit another PR

classicsong (approved):
LGTM

PR #1044, opened 2024-09-26
[Enhancement] Add a strict naming rule for graph names

classicsong:
*Issue #, if available:*
#957 

*Description of changes:*
The graph name must adhere to the Python identifier naming rules, with one exception that hyphens (-) are permitted.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1050, opened 2024-09-30
[GSProcessing] Fix to make graph_name optional again

thvasilo:
*Issue #, if available:*

*Description of changes:*

* GSProcessing makes providing a graph name optional. After https://github.com/awslabs/graphstorm/pull/1044 we were checking the graph name before we filled in default values. 
* We now provide a default value before the name validity check.
* Added some tests to check various graph naming cases.

### DIFF
--- a/graphstorm-processing/graphstorm_processing/distributed_executor.py
+++ b/graphstorm-processing/graphstorm_processing/distributed_executor.py
@@ -162,11 +162,13 @@ class DistributedExecutor:
         self.filesystem_type = executor_config.filesystem_type
         self.execution_env = executor_config.execution_env
         self.add_reverse_edges = executor_config.add_reverse_edges
-        self.graph_name = (
-            executor_config.graph_name
-            if executor_config.graph_name
-            else s3_utils.s3_path_remove_trailing(executor_config.input_prefix).split("/")[-1]
-        )
+        # We use the data location as the graph name if a name is not provided
+        if executor_config.graph_name:
+            self.graph_name = executor_config.graph_name
+        else:
+            derived_name = s3_utils.s3_path_remove_trailing(self.input_prefix).split("/")[-1]
+            logging.warning("Setting graph name derived from input path: %s", derived_name)
+            self.graph_name = derived_name
         check_graph_name(self.graph_name)
         self.repartition_on_leader = executor_config.do_repartition
         # Input config dict using GSProcessing schema
--- a/graphstorm-processing/graphstorm_processing/distributed_executor.py
+++ b/graphstorm-processing/graphstorm_processing/distributed_executor.py
@@ -59,7 +59,7 @@ from pathlib import Path
 import tempfile
 import time
 from collections.abc import Mapping
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 import boto3
 import botocore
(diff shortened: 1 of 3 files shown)


graphstorm-processing/graphstorm_processing/distributed_executor.py
```diff
@@ -162,7 +162,12 @@ def __init__(
         self.filesystem_type = executor_config.filesystem_type
         self.execution_env = executor_config.execution_env
         self.add_reverse_edges = executor_config.add_reverse_edges
-        self.graph_name = executor_config.graph_name
+        self.graph_name = (
+            executor_config.graph_name
+            if executor_config.graph_name
+            else s3_utils.s3_path_remove_trailing(executor_config.input_prefix).split("/")[-1]
+        )
```
jalencato, line 169: I am thinking about a case that if we do not specify the graph name in the running command and only have local prefix. What will happen here? Then will ``s3_path_remove_trailing`` return error?

jalencato (commented):
Another overall comment, do we try to use the PR to trigger one SageMaker job?

graphstorm-processing/graphstorm_processing/distributed_executor.py
zhjwy9343: Do we need the graph name in model training and inference on SageMaker?
If yes, how to return this auto-created graph name back to users so that they will use it in the next steps;
If no, why not just give a default name, e.g., `gs_graph_name`?

thvasilo: > Then will s3_path_remove_trailing return error?

Actually no, all the function does is remove a trailing `/` if one exists, could be a local or S3 URI.

> Do we need the graph name in model training and inference on SageMaker?

AFAIK the graph name used during gsprocessing does not need to be same for gspartition and training. The reason we choose to use the path is that it's likely more descriptive than a default name.

jalencato (approved):
LGTM

zhjwy9343 (approved):
LGTM

graphstorm-processing/graphstorm_processing/distributed_executor.py
classicsong: So the users do not need to be aware of the graph name?

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -447,6 +442,10 @@ def _initialize_metadata_dict(
         metadata_dict["edge_type"] = edge_types
         metadata_dict["node_type"] = sorted(node_type_set)
 
+        # We use the data location as the graph name, can also take from user?
+        # TODO: Fix this, take from config?
```
classicsong, line 446: What does the TODO mean?

graphstorm-processing/tests/test_dist_executor.py
```diff
@@ -148,3 +143,23 @@ def test_merge_input_and_transform_dicts(tempdir: str):
                 if "state" == feature["column"]:
                     transform_for_feature = feature["precomputed_transformation"]
                     assert transform_for_feature["transformation_name"] == "categorical"
+
+
+def test_dist_executor_graph_name(executor_configuration: ExecutorConfig):
+    """Test cases for graph name"""
+
+    # Ensure default value is used when graph_name is not provided
```
classicsong, line 151: You should also test the case when graph_name is given.

graphstorm-processing/graphstorm_processing/distributed_executor.py
thvasilo: Not for GSProcessing, we provide a default. For partition/training I believe it's a required argument but doesn't need to match the GSProcessing one.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: Was planning to include graph name in the config itself, I'll remove this.

PR #1053, opened 2024-09-30
Fix version to conform to semver

thvasilo:
*Issue #, if available:*

*Description of changes:*

* `0.4` is not a valid semver version, changing to `0.4.0`. See https://semver.org/ and examples here: https://regex101.com/r/Ly7O1x/3/


jalencato:
One question, can we still use ``pip install graphstorm==0.4`` after this PR in the future? Or we have to use 0.4.0

thvasilo:
Pip installation will be the version that we publish, which should be `0.4.0`, it doesn't depend on our local files.

jalencato:
> Pip installation will be the version that we publish, which should be `0.4.0`, it doesn't depend on our local files.

Yeah I mean if in this way we release the graphstorm 0.4, we will ask users to pip install in a way that ``pip install graphstorm==0.4.0``. In this case, is there a way to support both ``pip install graphstorm==0.4`` and ``pip install graphstorm==0.4.0``?

classicsong (approved):

thvasilo:
> In this case, is there a way to support both pip install graphstorm==0.4 and pip install graphstorm==0.4.0

Every release on pypi needs to be unique, creating two identical releases with different versioning schemes would likely create confusion.

jalencato:
> > In this case, is there a way to support both pip install graphstorm==0.4 and pip install graphstorm==0.4.0
> 
> Every release on pypi needs to be unique, creating two identical releases with different versioning schemes would likely create confusion.

Then we may need to highlight it in the document to let users know.

jalencato (approved):
LGTM

PR #1061, opened 2024-10-11
Add Adjusted Mean Ranking Index metric for Link Prediction

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Add Adjust Mean Rank Index LP metric. This metric is normalized by the candidate list size, allowing for easier comparison between datasets/models and negative edge sample counts.
* To get the list sizes we modify `run_lp_mini_batch_predict` and `lp_mini_batch_predict` to _conditionally_ return a tuple of `rankings, lengths` that allows us to calculate AMRI in the cases where it's needed. The return type is determined by a new argument added, with a default value, so the changes are backwards compatible, existing calls to the two functions will function as before.
* Add `LinkPredictionTestScoreInterface` as a common ancestor to `LinkPredictNoParamDecoder` and `LinkPredictLearnableDecoder`, this way we can ensure at runtime that the decoder should implement the `calc_test_scores` function that is used by the LP evaluator.
* Modify the `GSgnnLPRankingEvalInterface` `evaluate` function to add optional `**kwargs`. In cases where we want to calculate a metric that needs the candidate list lengths we use these kwargs to pass the information.
* Modify the `def compute_score(self, rankings, train=True, candidate_sizes=None):` to add the optional `candidate_sizes` argument, this is only used during AMRI calculation.

### DIFF
--- a/python/graphstorm/eval/eval_func.py
+++ b/python/graphstorm/eval/eval_func.py
@@ -210,6 +210,7 @@ class LinkPredictionMetrics:
         """
         # Need to check if the given metric is supported first
         self.assert_supported_metric(metric)
+        # The minimum value for AMRI is -1.0 so we init with that
         return -1.0 if metric == "amri" else 0.0
 
 
@@ -758,6 +759,12 @@ def compute_amri(ranking: th.Tensor, candidate_sizes: th.Tensor) -> th.Tensor:
         assert th.all(ranking <= candidate_sizes).item(), \
             "all ranks must be <= candidate_sizes"
 
+    # We use the simplified form of AMRI calculation
+    # 1 - \frac{MR-1}{E[MR-1]} = 1 - \frac{2*\sum_n{r-1}}{\sum_n{|S|}}
+    # where n is the number of evaluations (number of positive edges),
+    # r is the ranking of the positive edge in each ranked score list,
+    # and |S| is the edge candidate set size.
+    # See equation (8) in https://arxiv.org/abs/2002.06914
     nominator = 2 * th.sum(ranking - 1)
     if candidate_sizes.shape[0] == 1:
         denominator = candidate_sizes.item() * ranking.shape[0]
--- a/python/graphstorm/eval/eval_func.py
+++ b/python/graphstorm/eval/eval_func.py
@@ -701,7 +701,7 @@ def compute_mae(pred, labels):
     diff = th.abs(pred.cpu() - labels.cpu())
     return th.mean(diff).cpu().item()
 
-def compute_mrr(ranking):
+def compute_mrr(ranking: th.Tensor) -> th.Tensor:
     """ Get link prediction Mean Reciprocal Rank (MRR) metrics
 
     Parameters
--- a/python/graphstorm/eval/eval_func.py
+++ b/python/graphstorm/eval/eval_func.py
@@ -729,7 +729,16 @@ def compute_amri(ranking: th.Tensor, candidate_sizes: th.Tensor) -> th.Tensor:
         AMRI = 1 - \\frac{\\text{MR}-1}{\\mathbb{E}[\\text{MR}-1]}
 
     where MR is the mean rank, and `E[MR]` is the expected mean rank, which is used
-    to adjust for chance. Its values will be in the [-1, 1] range, where 1 corresponds
+    to adjust for chance. E[MR] is defined as:
+
+    .. math::
+        \\mathbb{E}[\\text{MR}] = \\mathbb{E} \\left[ \\frac{1}{n} \\sum^n_{i=1}{r_i} \\right]
+
+    Where :math:`r_i` is the rank the model assigns to the positive edge,
+    compared to the negative edges in the candidate list, and :math:`n` is the number of
+    candidate lists, one per positive edge.
+
+    AMRI values will be in the :math:`[-1, 1]` range, where 1 corresponds
     to optimal performance where each individual rank is 1. A value of 0 indicates
     model performance similar to a model assigning random scores, or equal score
     to every candidate. The value is negative if the model performs worse than the
--- a/python/graphstorm/eval/eval_func.py
+++ b/python/graphstorm/eval/eval_func.py
@@ -16,9 +16,11 @@
     Evaluation functions
 """
 import logging
+import operator
+from collections.abc import Callable
 from enum import Enum
 from functools import partial
-import operator
+
 import numpy as np
 import torch as th
 from sklearn.metrics import roc_auc_score
@@ -28,7 +30,7 @@ SUPPORTED_HIT_AT_METRICS = 'hit_at'
 SUPPORTED_CLASSIFICATION_METRICS = {'accuracy', 'precision_recall', \
     'roc_auc', 'f1_score', 'per_class_f1_score', 'per_class_roc_auc', SUPPORTED_HIT_AT_METRICS}
 SUPPORTED_REGRESSION_METRICS = {'rmse', 'mse', 'mae'}
-SUPPORTED_LINK_PREDICTION_METRICS = {"mrr", SUPPORTED_HIT_AT_METRICS}
+SUPPORTED_LINK_PREDICTION_METRICS = {"mrr", SUPPORTED_HIT_AT_METRICS, "amri"}
 
 class ClassificationMetrics:
     """ object that compute metrics for classification tasks.
@@ -158,16 +160,19 @@ class LinkPredictionMetrics:
         self.supported_metrics = SUPPORTED_LINK_PREDICTION_METRICS
 
         # This is the operator used to compare whether current value is better than the current best
-        self.metric_comparator = {}
+        self.metric_comparator: dict[str, Callable] = {}
         self.metric_comparator["mrr"] = operator.le
+        self.metric_comparator["amri"] = operator.le
 
         # This is the operator used to measure each metric performance
-        self.metric_function = {}
+        self.metric_function: dict[str, Callable[..., th.Tensor]] = {}
         self.metric_function["mrr"] = compute_mrr
+        self.metric_function["amri"] = compute_amri
 
         # This is the operator used to measure each metric performance in evaluation
-        self.metric_eval_function = {}
+        self.metric_eval_function: dict[str, Callable[..., th.Tensor]] = {}
         self.metric_eval_function["mrr"] = compute_mrr
+        self.metric_eval_function["amri"] = compute_amri
 
         if eval_metric_list:
             for eval_metric in eval_metric_list:
@@ -190,20 +195,22 @@ class LinkPredictionMetrics:
             assert metric in self.supported_metrics, \
                 f"Metric {metric} not supported for link prediction"
 
-    def init_best_metric(self, metric):
+    def init_best_metric(self, metric: str):
         """
         Return the initial value for the metric to keep track of the best metric.
         Parameters
         ----------
-        metric: the metric to initialize
+        metric: str
+            the name of the metric to initialize
 
         Returns
         -------
-
+        float
+            An initial value for the metric.
         """
         # Need to check if the given metric is supported first
         self.assert_supported_metric(metric)
-        return 0
+        return -1.0 if metric == "amri" else 0.0
 
 
 def labels_to_one_hot(labels, total_labels):
@@ -695,17 +702,65 @@ def compute_mae(pred, labels):
     return th.mean(diff).cpu().item()
 
 def compute_mrr(ranking):
-    """ Get link prediction mrr metrics
+    """ Get link prediction Mean Reciprocal Rank (MRR) metrics
 
-        Parameters
-        ----------
-        ranking:
-            ranking of each positive edge
+    Parameters
+    ----------
+    ranking: torch.Tensor
+        ranking of each positive edge
 
-        Returns
-        -------
-        link prediction mrr metrics: tensor
+    Returns
+    -------
+    th.Tensor
+        link prediction mrr metrics
     """
-    logs = th.div(1.0, ranking)
-    metrics = th.tensor(th.div(th.sum(logs),len(logs)))
+    reciprocal_ranks = th.div(1.0, ranking)
+    metrics = th.tensor(th.div(th.sum(reciprocal_ranks), len(reciprocal_ranks)))
     return metrics
+
+def compute_amri(ranking: th.Tensor, candidate_sizes: th.Tensor) -> th.Tensor:
+    """Computes the Adjusted Mean Rank Index (AMRI) for the given ranking and candidate sizes.
+
+    AMRI is a metric that evaluates the performance of link prediction models by considering both
+    the rank of the correct candidate and the number of candidates considered. It is calculated as:
+
+    .. math::
+        AMRI = 1 - \\frac{\\text{MR}-1}{\\mathbb{E}[\\text{MR}-1]}
+
+    where MR is the mean rank, and `E[MR]` is the expected mean rank, which is used
+    to adjust for chance. Its values will be in the [-1, 1] range, where 1 corresponds
+    to optimal performance where each individual rank is 1. A value of 0 indicates
+    model performance similar to a model assigning random scores, or equal score
+    to every candidate. The value is negative if the model performs worse than the
+    constant-score model."
+
+    For more details see https://arxiv.org/abs/2002.06914
+
+    Parameters
+    ----------
+    ranking : torch.Tensor
+        ranking of each positive edge
+    candidate_sizes : th.Tensor
+        The size of each candidate list. If all lists have
+        the same size this will be a single-value tensor.
+
+    Returns
+    -------
+    th.Tensor
+        A single-value Tensor with the AMRI metric.
+
+    .. versionadded: 0.4.0
+    """
+    if candidate_sizes.shape[0] > 1:
+        assert ranking.shape[0] == candidate_sizes.shape[0], \
+            "ranking and candidate_sizes must have the same length"
+        assert th.all(ranking <= candidate_sizes).item(), \
+            "all ranks must be <= candidate_sizes"
+
+    nominator = 2 * th.sum(ranking - 1)
+    if candidate_sizes.shape[0] == 1:
+        denominator = candidate_sizes.item() * ranking.shape[0]
+    else:
+        denominator = th.sum(candidate_sizes)
+
+    return 1 - th.div(nominator, denominator)
--- a/python/graphstorm/eval/eval_func.py
+++ b/python/graphstorm/eval/eval_func.py
@@ -741,7 +741,7 @@ def compute_amri(ranking: th.Tensor, candidate_sizes: th.Tensor) -> th.Tensor:
     ranking : torch.Tensor
         ranking of each positive edge
     candidate_sizes : th.Tensor
-        The size of each candidate list. If all lists have
+        The size of each candidate list. If all candidate lists have
         the same size this will be a single-value tensor.
 
     Returns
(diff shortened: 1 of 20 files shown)


python/graphstorm/eval/evaluator.py
```diff
@@ -1051,6 +1116,23 @@ def evaluate(self, val_rankings, test_rankings, total_iters):
             {etype: ranking}.
         total_iters: int
             The current iteration number.
+        kwargs: dict
+            Keyword arguments to pass downstream to metric calculation functions.
+
+            Currently we support:
+
+            val_candidate_sizes : dict of tensors
+                The size of each candidate list in the validation set, in the format of
```
classicsong, line 1125: ditto

python/graphstorm/model/edge_decoder.py
```diff
@@ -909,7 +909,58 @@ def predict_proba(self, g, h, e_h):
         return out
 
 ##################### Link Prediction Decoders #######################
-class LinkPredictNoParamDecoder(GSLayerNoParam):
+class LinkPredictionTestScoreMixin(abc.ABC):
```
classicsong, line 912: Why we add this Interface?

python/graphstorm/model/edge_decoder.py
```diff
@@ -909,7 +909,58 @@ def predict_proba(self, g, h, e_h):
...
+    """ Mixin class for link prediction test score computation
```
classicsong, line 913: Shall we call this an Interface class?

python/graphstorm/model/lp_gnn.py
```diff
@@ -189,28 +207,44 @@ def run_lp_mini_batch_predict(decoder, emb, loader, device):
...
-        ranking = {}
+        ranking: dict[tuple, list[th.Tensor]] = defaultdict(list)
+        batch_lengths: dict[tuple, list[th.Tensor]] = defaultdict(list)
+        assert isinstance(decoder, LinkPredictionTestScoreMixin)
         for pos_neg_tuple, neg_sample_type in loader:
-            score = \
+            score: dict[tuple, tuple[th.Tensor, th.Tensor]] = \
                 decoder.calc_test_scores(
                     emb, pos_neg_tuple, neg_sample_type, device)
+            print(f"{score=}")
```
classicsong, line 229: remove print

classicsong (commented):
Two high level comments:
1. Shall we mark this PR "break back compatibility"?
2. Can you add more test for evaluator and add some end2end tests?

python/graphstorm/eval/eval_func.py
```diff
@@ -158,16 +160,19 @@ def __init__(self, eval_metric_list=None):
         self.supported_metrics = SUPPORTED_LINK_PREDICTION_METRICS
 
         # This is the operator used to compare whether current value is better than the current best
-        self.metric_comparator = {}
+        self.metric_comparator: dict[str, Callable] = {}
```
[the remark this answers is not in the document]
thvasilo: We could eventually, but adding piece-meal is easier, we can grow the annotations with time.

python/graphstorm/eval/eval_func.py
```diff
@@ -190,20 +195,22 @@ def assert_supported_metric(self, metric):
...
         Returns
         -------
-
+        float
+            An initial value for the metric.
         """
         # Need to check if the given metric is supported first
         self.assert_supported_metric(metric)
-        return 0
+        return -1.0 if metric == "amri" else 0.0
```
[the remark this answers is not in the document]
thvasilo: My understanding is that the initial value should be the minimum possible value for the metric, which for AMRI is `-1.0`. Is my understanding correct?

From the paper:

> AMRI has a bounded value range of [−1,1]. A value of 1 corresponds to optimal performance where each individual rank is 1. A value of 0 indicates model performance similar to a model assigning random scores, or equal score to every candidate. The value is negative if the model performs worse than the constant-score model.

python/graphstorm/eval/eval_func.py
```diff
@@ -695,17 +702,65 @@ def compute_mae(pred, labels):
...
-    metrics = th.tensor(th.div(th.sum(logs),len(logs)))
+    reciprocal_ranks = th.div(1.0, ranking)
+    metrics = th.tensor(th.div(th.sum(reciprocal_ranks), len(reciprocal_ranks)))
     return metrics
+
+def compute_amri(ranking: th.Tensor, candidate_sizes: th.Tensor) -> th.Tensor:
+    """Computes the Adjusted Mean Rank Index (AMRI) for the given ranking and candidate sizes.
+
+    AMRI is a metric that evaluates the performance of link prediction models by considering both
+    the rank of the correct candidate and the number of candidates considered. It is calculated as:
```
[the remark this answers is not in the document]
thvasilo: We include the positive example so it would be all candidates (we use `num_negative+1`). From the paper:

> To compute a rank in LP and EA evaluation, we are given a list of scores `S = [β1, . . . , βC ]` for each test instance with `|C| = C`, where C is a set of candidates. We denote the score of the ”true” entity as α, and its position in the decreasingly sorted list as `r = rank(S, α)`.

python/graphstorm/eval/eval_func.py
```diff
@@ -695,17 +702,65 @@ def compute_mae(pred, labels):
...
+
+    .. math::
+        AMRI = 1 - \\frac{\\text{MR}-1}{\\mathbb{E}[\\text{MR}-1]}
+
+    where MR is the mean rank, and `E[MR]` is the expected mean rank, which is used
```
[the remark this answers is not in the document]
thvasilo: Assuming individual ranks are independent, it's computed as

image

I could include this definition in the docstring? Just trying to keep this short, for the final PR (this is still a draft) I'll include more extensive documentation in the rst docs

python/graphstorm/eval/eval_func.py
```diff
@@ -695,17 +702,65 @@ def compute_mae(pred, labels):
...
+    constant-score model."
+
+    For more details see https://arxiv.org/abs/2002.06914
+
+    Parameters
+    ----------
+    ranking : torch.Tensor
+        ranking of each positive edge
+    candidate_sizes : th.Tensor
+        The size of each candidate list. If all lists have
```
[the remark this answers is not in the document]
thvasilo: Yes, it will be num_negative +  1 per ranking

python/graphstorm/eval/eval_func.py
```diff
@@ -695,17 +702,65 @@ def compute_mae(pred, labels):
...
+
+    .. versionadded: 0.4.0
+    """
+    if candidate_sizes.shape[0] > 1:
+        assert ranking.shape[0] == candidate_sizes.shape[0], \
+            "ranking and candidate_sizes must have the same length"
+        assert th.all(ranking <= candidate_sizes).item(), \
+            "all ranks must be <= candidate_sizes"
+
+    nominator = 2 * th.sum(ranking - 1)
```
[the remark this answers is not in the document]
thvasilo: I'm using the simplified form of the metric defined here:

image

python/graphstorm/eval/evaluator.py
```diff
@@ -224,6 +234,9 @@ def compute_score(self, rankings, train=True):
             Rankings of positive scores in the format of {etype: ranking}
         train: boolean
             If in model training.
+        candidate_sizes : torch.Tensor, optional
+            The size of each candidate list. If all lists have the
+            same size this will be an int, otherwise a list of ints.
```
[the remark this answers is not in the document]
thvasilo: Ah thanks for the catch this was my initial implementation, I switched to tensors after

python/graphstorm/model/edge_decoder.py
thvasilo: After discussing with Jian it seems like all LP decoders need to implement this function, with the same signature. Adding this as an interface allows us to apply it to the two class hierarchies (`LinkPredictNoParamDecoder`, `LinkPredictLearnableDecoder`) as a mixin. Then we in `lp_gnn.py` we can have the assertion 

```
assert isinstance(decoder, LinkPredictionTestScoreInterface)
for pos_neg_tuple, neg_sample_type in loader:
    score: dict[tuple, tuple[th.Tensor, th.Tensor]] = \
        decoder.calc_test_scores(
            emb, pos_neg_tuple, neg_sample_type, device)
```

which guarantees that the decoder indeed should implement this function and informs the type inference of the function signature e.g.

image

if we tried to pass a wrong argument type we get an error from the type-checker (`device` is of type `torch.device` but here we try to pass an `int`):

image

thvasilo:
> Shall we mark this PR "break back compatibility"?

Does it break backwards compatibility somehow? I think all modifications I made are backwards compatible through default arguments (so existing calls should still work). Let me know if I missed something.

I'll add e2e tests along with docs in the next commit for this PR and then mark as ready for review.

python/graphstorm/eval/evaluator.py
```diff
@@ -934,7 +972,12 @@ def evaluate(self, val_rankings, test_rankings, total_iters):
 
         return val_score, test_score
 
-    def compute_score(self, rankings, train=True):
+    def compute_score(
+            self,
+            rankings,
+            train=True,
+            candidate_sizes: Optional[dict[str, th.Tensor]] = None,
```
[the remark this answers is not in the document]
thvasilo: Undecided on this one, do we expect end users to override this function/implement on their own? If yes we can add the kwargs to allow the user to pass in any data they need.

python/graphstorm/eval/eval_func.py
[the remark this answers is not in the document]
classicsong: I c.

classicsong:
> > Shall we mark this PR "break back compatibility"?
> 
> Does it break backwards compatibility somehow? I think all modifications I made are backwards compatible through default arguments (so existing calls should still work). Let me know if I missed something.
> 
> I'll add e2e tests along with docs in the next commit for this PR and then mark as ready for review.

I think this PR changes the API definition, though it is still backwards compatible. We should highlight the changes in the release note.

python/graphstorm/eval/evaluator.py
```diff
@@ -2064,7 +2184,13 @@ def best_iter_num(self):
     def val_perf_rank_list(self):
         raise RuntimeError("GSgnnMultiTaskEvaluator.val_perf_rank_list not supported")
 
-    def evaluate(self, val_results, test_results, total_iters):
+    def evaluate(
+        self,
+        val_results: Dict[str, Any],
+        test_results: Dict[str, Any],
+        total_iters: int,
+        **kwargs
```
classicsong, line 2192: It seems kwargs are not used.

python/graphstorm/inference/mt_infer.py
```diff
@@ -309,13 +311,22 @@ def nfrecon_gen_embs(skip_last_self_loop=False, node_embs=embs):
...
         if do_eval:
             test_start = time.time()
+            assert isinstance(self.evaluator, GSgnnMultiTaskEvaluator)
+
             val_score, test_score = self.evaluator.evaluate(
-                pre_results, pre_results, 0)
+                pre_results,
+                pre_results,
+                0,
+                val_candidate_sizes=test_lengths,
```
classicsong, line 326: Do we need to send `val_candidate_sizes` and `test_candidate_sizes` to `GSgnnMultiTaskEvaluator.evaluate`?
You have sent them through pre_results.

python/graphstorm/model/edge_decoder.py
```diff
@@ -909,7 +911,58 @@ def predict_proba(self, g, h, e_h):
         return out
 
 ##################### Link Prediction Decoders #######################
-class LinkPredictNoParamDecoder(GSLayerNoParam):
+class LinkPredictionTestScoreInterface(abc.ABC):
```
classicsong, line 914: Shall we add this class in the API doc?

python/graphstorm/model/lp_gnn.py
```diff
@@ -187,16 +206,23 @@ def run_lp_mini_batch_predict(decoder, emb, loader, device):
...
             Rankings of positive scores in format of {etype: ranking}
+        rankings, batch_lengths: tuple[dict, dict], if `return_batch_lengths` was True
+            A tuple of rankings of positive scores in format of {etype: ranking},
+            and the corresponding batch lengths for each ranking value.
     """
     with th.no_grad():
-        ranking = {}
+        ranking: Dict[Tuple, List[th.Tensor]] = defaultdict(list)
+        batch_lengths: Dict[Tuple, List[th.Tensor]] = defaultdict(list)
+        assert isinstance(decoder, LinkPredictionTestScoreInterface)
```
classicsong, line 225: Please add an error string.

python/graphstorm/model/lp_gnn.py
```diff
@@ -205,12 +231,27 @@ def run_lp_mini_batch_predict(decoder, emb, loader, device):
...
-        rankings = {}
+                assert pos_score.shape[0] == neg_score.shape[0], \
+                    "There should be as many negative lists as there are positive examples"
+                score_ranking = calc_ranking(pos_score, neg_score)
+                ranking[canonical_etype].append(score_ranking)
+                # Set the number of candidates for each positive example
+                # (equal to neg_score.shape[0])
+                # which will be the number of negatives (equal to neg_score.shape[1])
+                # plus one for the positive example
+                lengths_tensor = th.tensor(neg_score.shape[0] * [neg_score.shape[1] + 1])
```
classicsong, line 242: Should it be a single value as (neg_score.shape[1] + 1)? 
Why do you compute the total number of negatives in a batch?

python/graphstorm/trainer/lp_trainer.py
```diff
@@ -361,16 +362,27 @@ def eval(self, model, data, val_loader, test_loader,
...
         if test_loader is not None:
-            test_scores = lp_mini_batch_predict(model, emb, test_loader, self.device)
+            test_rankings, test_lengths = lp_mini_batch_predict(
+                model, emb, test_loader, self.device, return_batch_lengths=True)
         else:
-            test_scores = None
+            test_rankings, test_lengths = None, None
         sys_tracker.check('after_test_score')
+        assert self.evaluator is not None
+        assert isinstance(self.evaluator, GSgnnLPRankingEvalInterface)
```
classicsong, line 378: Add error string.

python/graphstorm/trainer/lp_trainer.py
```diff
@@ -361,16 +362,27 @@ def eval(self, model, data, val_loader, test_loader,
...
         sys_tracker.check('after_val_score')
         if test_loader is not None:
-            test_scores = lp_mini_batch_predict(model, emb, test_loader, self.device)
+            test_rankings, test_lengths = lp_mini_batch_predict(
+                model, emb, test_loader, self.device, return_batch_lengths=True)
         else:
-            test_scores = None
+            test_rankings, test_lengths = None, None
         sys_tracker.check('after_test_score')
+        assert self.evaluator is not None
```
classicsong, line 377: Add error string.

tests/unit-tests/test_eval_func.py
```diff
@@ -567,6 +570,49 @@ def test_compute_hit_at_link_prediction():
...
+        emr = (1 / (2 * candidate_tensor.shape[0])) * th.sum(candidate_tensor)
+        expected_amri = 1 - ((mr - 1) / emr)
+        return expected_amri
+
+    actual_amri = compute_amri(ranks, candidate_lists)
+    expected_amri = amri_definition(ranks, candidate_lists)
+    assert_almost_equal(
+        actual_amri,
+        expected_amri,
+        decimal=4
```
classicsong, line 590: Can we set it to 6?

tests/unit-tests/test_trainer.py
```diff
@@ -518,6 +519,10 @@ def compare_results(target_res, check_res):
                     assert_equal(tr_1, cr_1)
                     assert_equal(tr_2, cr_2)
                 else:
+                    # In case LP results also returned candidate list
+                    # lengths, we only keep the rankings
+                    if isinstance(check_r, tuple):
+                        check_r, _ = check_r
```
classicsong, line 525: Shall we also check the lengths?

python/graphstorm/model/lp_gnn.py
thvasilo: We need one length per positive, we are using `*` as a way to create a list of all same numbers, e.g. `3 * [4] => [4, 4, 4]`.

So the tensor will just have the length of the examples with all same values.

python/graphstorm/model/edge_decoder.py
thvasilo: An end user would need to implement either `LinkPredictNoParamDecoder` or a `LinkPredictLearnableDecoder`, so we can just keep those two in the API. There's not much value in this interface in isolation, it needs to be part of a `*Decoder`

python/graphstorm/eval/eval_func.py
```diff
@@ -694,18 +701,67 @@ def compute_mae(pred, labels):
...
+    """
+    if candidate_sizes.shape[0] > 1:
+        assert ranking.shape[0] == candidate_sizes.shape[0], \
+            ("ranking and candidate_sizes must have the same length, "
+             f"got {ranking.shape=} {candidate_sizes.shape=}" )
+        assert th.all(ranking <= candidate_sizes).item(), \
+            "all ranks must be <= candidate_sizes"
+
+    nominator = 2 * th.sum(ranking - 1)
+    if candidate_sizes.shape[0] == 1:
```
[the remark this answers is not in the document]
thvasilo: Yes, `candidate_sizes` is a torch.Tensor, AFAIK `len(candidate_sizes) == candidate_sizes.shape[0]` always right?

python/graphstorm/eval/evaluator.py
```diff
@@ -954,27 +1014,37 @@ def compute_score(self, rankings, train=True):
...
+        sizes_list = []
+        candidate_sizes: Optional[Dict[str, th.Tensor]] = kwargs.get("candidate_sizes", None)
 
         # compute ranking value for each metric
-        metrics = {}
+        metrics: Dict[str, th.Tensor] = {}
         for metric in self.metric_list:
+            if metric == "amri":
+                assert candidate_sizes
+                for _, size in candidate_sizes.items():
```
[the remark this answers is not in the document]
thvasilo: Yes, they get added to the dicts in the same order, and since Python 3.7 insertion key order is guaranteed.

python/graphstorm/eval/evaluator.py
thvasilo: Yes, the MT evaluator does not use the `kwargs` currently. I just added it here to have a consistent interface with the other evaluators, maybe we can remove it?

python/graphstorm/inference/mt_infer.py
thvasilo: Maybe I'm confusing the interface here, let me see if we need both.

tests/unit-tests/test_trainer.py
thvasilo: Check the lengths to ensure rankings and batch lengths have the correct (same) lengths?

python/graphstorm/model/lp_gnn.py
classicsong: I c.

python/graphstorm/eval/evaluator.py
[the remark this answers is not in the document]
classicsong: Shall we check the order when metric is "amri". Just in case.

python/graphstorm/eval/evaluator.py
```diff
@@ -1023,8 +1023,8 @@ def compute_score(
             if metric == "amri":
                 assert candidate_sizes, \
                     f"candidate_sizes needs to have a value for AMRI, got {candidate_sizes=}."
-                for _, size in candidate_sizes.items():
-                    sizes_list.append(size)
+                for etype, _ in rankings.items():
+                    sizes_list.append(candidate_sizes[etype])
```
classicsong, line 1027: I think we need to add a TODO or a HINT here: if in the future, there is another metric that needs sizes_list, we should avoid adding the size of each candidate twice.

classicsong (approved):

PR #1073, opened 2024-10-22
[GSProcessing] Replace colons in edge feature paths with underscores to ensure EFS compatibility.

thvasilo:
*Issue #, if available:*

*Description of changes:*

* When copying data from S3 to HDFS or EFS we can't use paths that contain colons `:`. We have already fixed this for edge structures and labels, this fixes it for features as well.


jalencato (commented):
Do we need to check the label as well?

thvasilo:
Label is done as well. Only features were missing.

jalencato (approved):
LGTM

PR #1076, opened 2024-10-28
[GSProcessing] Remove poetry as a requirement for building images.

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Previously, build the gsprocessing images required Python 3.9 and poetry to be installed, leading to user friction. 
* With this commit we instead include requirements on a per-version basis, and no longer build wheels using poetry but install from source for both `prod` and `test` image targets.
* We maintain backwards compatibility if the exported requirements file is not available for a particular version.
* Also move EMRS image to 7.3.0


docs/source/cli/graph-construction/distributed/gsprocessing/distributed-processing-setup.rst
```diff
@@ -45,22 +44,6 @@ you'll need to have the Docker engine installed.
 To install Docker follow the instructions at the
 `official site <https://docs.docker.com/engine/install/>`_.
 
-Install Poetry
```
jalencato, line 48: Do we still want to mention anywhere about poetry installation or we just remove all of the existence?

graphstorm-processing/docker/0.4.0/requirements.txt
```diff
@@ -0,0 +1,942 @@
+attrs==23.2.0 ; python_full_version >= "3.9.12" and python_full_version < "3.10.0" \
```
jalencato, line 1: Why do we need the requirements.txt here?

jalencato (commented):
I feel confused about what we want to do with the PR here. If we remove the poetry dependencies, it seems like we make everything more complicated. Any reason we want to do that here?

docs/source/cli/graph-construction/distributed/gsprocessing/distributed-processing-setup.rst
thvasilo: We only keep it in the developer guide.

graphstorm-processing/docker/0.4.0/requirements.txt
thvasilo: Previously we were using poetry to generate the requirements, now we pre-generate this, once per version and include it with the Dockerfiles.

thvasilo:
> I feel confused about what we want to do with the PR here. If we remove the poetry dependencies, it seems like we make everything more complicated. Any reason we want to do that here?

The user story is this: 

**As a user trying to build and push Docker images for GSProcessing, I would like to not have to modify my Python environment to do so.**

Currently, building the GSProcessing image _requires_ Python 3.9 and having `poetry` installed in the user environment. I've seen many instances of users struggling to install Python 3.9 and setting up the dev environment correctly with pyenv/conda/mise.

This PR removes the requirement of having Python 3.9 and poetry installed, users can build the image regardless of the version of Python they have, and don't have to worry about `poetry

graphstorm-processing/docker/0.4.0/emr-serverless/Dockerfile.cpu
```diff
@@ -1,5 +1,5 @@
 ARG ARCH=x86_64
-FROM public.ecr.aws/emr-serverless/spark/emr-7.1.0:20240528-${ARCH} as base
+FROM public.ecr.aws/emr-serverless/spark/emr-7.3.0:20241008-${ARCH} as base
```
jalencato, line 2: Do we want to move with the latest here: public.ecr.aws/emr-serverless/spark/emr-7.6.0:20250107?

graphstorm-processing/docker/0.4.0/requirements.txt
jalencato: May we regenerate the requirements.txt here to make sure it works with the latest code repo?

graphstorm-processing/docker/0.4.0/emr-serverless/Dockerfile.cpu
thvasilo: I have not tested that version yet, let's move that to 0.4.1?

jalencato (approved):
LGTM. Just need to make sure the requirement.txt is latest

PR #1079, opened 2024-10-29
Allow using multiple metrics when using early stop.

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Previously we had an assertion that did not allow a dict with two metrics to be passed to the early stop check function.
* This change removes that and adds a case in the e2e tests that does early stop but has more than one eval metric.

Example error we used to get:

```
Traceback (most recent call last):
[rank77]:   File "/graphstorm/python/graphstorm/run/gsgnn_lp/gsgnn_lp.py", line 180, in <module>
[rank77]:     main(gs_args)
[rank77]:   File "/graphstorm/python/graphstorm/run/gsgnn_lp/gsgnn_lp.py", line 136, in main
[rank77]:     trainer.fit(train_loader=dataloader, val_loader=val_dataloader,
[rank77]:   File "/graphstorm/python/graphstorm/trainer/lp_trainer.py", line 247, in fit
[rank77]:     if self.evaluator.do_early_stop(val_score):
[rank77]:   File "/graphstorm/python/graphstorm/eval/evaluator.py", line 358, in do_early_stop
[rank77]:     assert len(val_score) == 1, \
[rank77]: AssertionError: validation score should be a single key value pair but got {'amri': 0.688815712928772, 'hit_at_10': 0.13315722346305847, 'mrr': 0.06222189590334892}
```


zhjwy9343 (approved):
LGTM

PR #1080, opened 2024-10-29
[GSProcessing&GSPartition] Support Hard Negative

jalencato:
*Issue #, if available:*

*Description of changes:*

* Add feature support for hard negative for distributed graph construction pipeline. For partition stage, users does not need to do any additional step for hard negative feature transformation itself.
* Implementation with the way we agreed on with the design doc.
* Create additional steps for distributed partition as we can not fully reused the previous script because different ways for mapping files. I create a separate one in the gspartition code repo to easy maintain.
* Tested with 1M nodes and 1M edges.


graphstorm-processing/tests/test_dist_hard_negative_transformation.py
```diff
@@ -0,0 +1,107 @@
...
+        "format_name": "parquet",
+    }
+    hard_negative_transformation = DistHardEdgeNegativeTransformation(
+        ["hard_negative"], spark=spark, edge_mapping_dict=edge_mapping_dict, separator=None
+    )
+    output_df = hard_negative_transformation.apply(input_df)
+    check_df_schema(output_df)
+    output_data = output_df.collect()
+
+    expected_output = [[1, -1, -1, -1], [2, 3, -1, -1], [3, 0, 1, -1], [0, -1, -1, -1]]
```
classicsong, line 61: Why the dimension of the hard negative tensor is 4?
How do you decide the dimension?

graphstorm-processing/tests/test_dist_hard_negative_transformation.py
```diff
@@ -0,0 +1,107 @@
...
+        "format_name": "parquet",
+    }
+    hard_negative_transformation = DistHardEdgeNegativeTransformation(
+        ["hard_negative"], spark=spark, edge_mapping_dict=edge_mapping_dict, separator=";"
+    )
+    output_df = hard_negative_transformation.apply(input_df)
+    check_df_schema(output_df)
+    output_data = output_df.collect()
+
+    expected_output = [[1, -1, -1, -1], [2, 3, -1, -1], [3, 0, 1, -1], [0, -1, -1, -1]]
```
classicsong, line 102: Same question.

python/graphstorm/gpartition/dist_partition_graph.py
```diff
@@ -189,12 +190,33 @@ def main():
...
         )
 
+    # Hard Negative Mapping
+    if args.gsprocessing_config:
+        gsprocessing_config = args.gsprocessing_config
+        shuffle_hard_negative_nids(f"{args.input_path}/{gsprocessing_config}",
+                                   args.num_parts, args.output_path)
+    else:
+        for filename in os.listdir(args.input_path):
+            if filename.endswith("_with_transformations.json"):
```
classicsong, line 200: Can you add some comments here?

python/graphstorm/gpartition/dist_partition_graph.py
```diff
@@ -189,12 +190,33 @@ def main():
...
+                         "please upgrade to the latest GSProcessing.")
+
+
 def parse_args() -> argparse.Namespace:
     """Parses arguments for the script"""
     argparser = argparse.ArgumentParser("Partition DGL graphs for node and edge classification "
                                         + "or regression tasks")
     argparser.add_argument("--input-path", type=str, required=True,
                            help="Path to input DGL chunked data.")
+    argparser.add_argument("--gsprocessing-config", type=str,
```
classicsong, line 218: @thvasilo Is there a way to get the path of this config automatically?

python/graphstorm/gpartition/dist_partition_graph.py
```diff
@@ -189,12 +190,33 @@ def main():
             dirs_exist_ok=True,
         )
 
+    # Hard Negative Mapping
+    if args.gsprocessing_config:
+        gsprocessing_config = args.gsprocessing_config
+        shuffle_hard_negative_nids(f"{args.input_path}/{gsprocessing_config}",
```
classicsong, line 196: How do you know, there is a hard negative feature?

graphstorm-processing/graphstorm_processing/data_transformations/dist_feature_transformer.py
```diff
@@ -38,14 +39,20 @@ class DistFeatureTransformer(object):
     """
 
     def __init__(
-        self, feature_config: FeatureConfig, spark: SparkSession, json_representation: dict
+        self,
+        feature_config: FeatureConfig,
+        spark: SparkSession,
+        json_representation: dict,
+        edge_mapping_dict: dict = None,
```
thvasilo, line 46: I'd prefer if we didn't change the function's signature to support a single transformation if possible? 

Is it possible to provide this information inside the `feature_config.transformation_kwargs`? We can modify its value adding this dict as an additional key, only in the case where we get `feature_config.feat_type == "edge_dst_hard_negative"`

python/graphstorm/gpartition/post_hard_negative.py
```diff
@@ -0,0 +1,109 @@
...
+            if single_feature["transformation"]["name"] \
+                    == "edge_dst_hard_negative":
+                edge_type = ":".join([single_edge_config["source"]["type"],
+                                     single_edge_config["relation"]["type"],
+                                     single_edge_config["dest"]["type"]])
+                hard_neg_feat_name = single_feature['name']
+                mapping_edge_list.append({"dst_node_type": single_edge_config["dest"]["type"],
+                                          "edge_type": edge_type,
+                                          "hard_neg_feat_name": hard_neg_feat_name})
+    return mapping_edge_list
```
classicsong, line 53: hard_neg_list

python/graphstorm/gpartition/post_hard_negative.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def shuffle_hard_negative_nids(gsprocessing_config, num_parts, output_path):
+    """Shuffle hard negative edge feature ids with int-to-int node id mapping.
+    The function here align with the shuffle_hard_nids in graphstorm.gconstruct.utils.
+    Create an additional function to handle the id mappings under distributed setting.
```
classicsong, line 59:
```suggestion
    Create an additional function to handle the id mappings under the distributed setting.
```

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -1654,7 +1654,15 @@ def _process_edge_features(
...
-            transformer = DistFeatureTransformer(feat_conf, self.spark, json_representation)
+            # Hard Negative Transformation use case, but should be able to be reused
+            edge_mapping_dict = {
+                "edge_type": edge_type,
+                "mapping_path": f"{self.output_prefix}/raw_id_mappings/",
+                "format_name": FORMAT_NAME,
+            }
+            transformer = DistFeatureTransformer(
+                feat_conf, self.spark, json_representation, edge_mapping_dict
+            )
```
thvasilo, line 1665: Instead of passing `edge_mapping_dict` to all transforms we could do something like 

```
if feature_config.feat_type == "edge_dst_hard_negative":
    # Hard Negative Transformation use case, but should be able to be reused
    edge_mapping_dict = {
        "edge_type": edge_type,
        "mapping_path": f"{self.output_prefix}/raw_id_mappings/",
        "format_name": FORMAT_NAME,
    }
    feature_config.transformation_kwargs["edge_mapping_dict"] = edge_mapping_dict
```

python/graphstorm/gpartition/post_hard_negative.py
```diff
@@ -0,0 +1,109 @@
...
+    Parameters
+    ----------------
+    gsprocessing_config: str
+        Path to the gsprocessing config.
+    num_parts: int
+        Number of parts.
+    output_path: str
+        Path to the output DGL graph.
+    """
+    shuffled_edge_config = load_hard_negative_config(gsprocessing_config)
```
classicsong, line 70: hard_neg_list?

python/graphstorm/gpartition/post_hard_negative.py
```diff
@@ -0,0 +1,109 @@
...
+    The function here align with the shuffle_hard_nids in graphstorm.gconstruct.utils.
+    Create an additional function to handle the id mappings under distributed setting.
+
+    Parameters
+    ----------------
+    gsprocessing_config: str
+        Path to the gsprocessing config.
+    num_parts: int
+        Number of parts.
+    output_path: str
```
classicsong, line 67: Will graph_path be a better naming?

graphstorm-processing/tests/test_dist_hard_negative_transformation.py
```diff
@@ -0,0 +1,107 @@
...
+
+    for idx, row in enumerate(output_data):
+        np.testing.assert_almost_equal(
+            row[0], expected_output[idx], decimal=3, err_msg=f"Row {idx} is not equal"
+        )
```
thvasilo, line 66: If these are supposed to be integers we should use `assert_equal`

python/graphstorm/model/utils.py
```diff
@@ -393,6 +393,11 @@ def _exchange_node_id_mapping(rank, world_size, device,
     # move mapping into CPU
     return gather_list[0].to(th.device("cpu"))
 
+def load_dist_nid_map(node_id_mapping_file, ntypes):
+    """ Wrapper for load_dist_nid_map.
```
classicsong, line 397: Please update the doc string.
_load_dist_nid_map is an internal function. You can not use an internal function to explain a public function.

python/graphstorm/gpartition/post_hard_negative.py
```diff
@@ -0,0 +1,109 @@
...
+        Path to the output DGL graph.
+    """
+    shuffled_edge_config = load_hard_negative_config(gsprocessing_config)
+
+    node_type_list = []
+    for single_shuffled_edge_config in shuffled_edge_config:
+        node_type = single_shuffled_edge_config["dst_node_type"]
+        node_type_list.append(node_type)
+    node_mapping = load_dist_nid_map(f"{output_path}/dist_graph", node_type_list)
```
classicsong, line 76: Why `f"{output_path}/dist_graph"`?

tests/unit-tests/gpartition/test_hard_negative_post_partition.py
```diff
@@ -0,0 +1,278 @@
...
+import pytest
+
+from numpy.testing import assert_almost_equal
+from graphstorm.model.utils import load_dist_nid_map
+from dgl.data.utils import load_tensors, save_tensors
+from graphstorm.gpartition.post_hard_negative import (shuffle_hard_negative_nids,
+                                                      load_hard_negative_config)
+
+@pytest.fixture(scope="module", name="gsprocessing_hard_negative_config")
+def gsprocessing_config_hard_negative_dict_fixture() -> Dict:
```
thvasilo, line 31: Cloud we save this to a file instead and use the fixture to load the JSON into a dict?

python/graphstorm/gpartition/post_hard_negative.py
```diff
@@ -0,0 +1,109 @@
...
+    gnid2pnid_mapping = {}
+
+    def get_gnid2pnid_map(ntype):
```
classicsong, line 79: We can call shuffle_hard_nids directly?

python/graphstorm/gpartition/dist_partition_graph.py
thvasilo: Good point, if the input data were created by GSProcessing we always save a copy of the gsprocessing config file in the output. We also save the job launch arguments in launch_arguments.json. So if we load `launch_arguments.json` the name of the GSP config file will be available there with the key `config_filename`

graphstorm-processing/graphstorm_processing/constants.py
```diff
@@ -58,6 +58,10 @@
 HUGGINGFACE_TOKENIZE = "tokenize_hf"
 HUGGINGFACE_EMB = "embedding_hf"
 
+################# Node Mapping  ################
+NODE_MAPPING_STR = "orig"
```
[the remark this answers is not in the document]
jalencato: They are referred to node mapping. orig is the string id, and new is the int id. I am just referring them to a constant value. https://github.com/awslabs/graphstorm/blob/de31f7b65acdcbc6aecc8c214fb6771640e5409b/graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py#L74

graphstorm-processing/graphstorm_processing/data_transformations/dist_feature_transformer.py
```diff
@@ -38,14 +39,20 @@ class DistFeatureTransformer(object):
...
     ):
         feat_type = feature_config.feat_type
         feat_name = feature_config.feat_name
         args_dict = feature_config.transformation_kwargs
         self.transformation: DistributedTransformation
         # We use this to re-apply transformations
         self.json_representation = json_representation
+        # Node Mapping Info for hard negative feature transformation
+        self.edge_mapping_dict = edge_mapping_dict
```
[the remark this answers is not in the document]
jalencato: Make sense. I will make the change.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
jalencato: That makes sense. I will make the change.

graphstorm-processing/tests/test_dist_hard_negative_transformation.py
jalencato: In total there are 4 kinds of nodes for the dst_nodes, so the length here should be 4. I will add a comment to explain that.

graphstorm-processing/tests/test_dist_hard_negative_transformation.py
jalencato: Same as above, 4 kinds of nodes for the dst node type.

python/graphstorm/gpartition/post_hard_negative.py
jalencato: For gspartition, the generated dgl graph will be stored under subdirectory ``dist_graph``.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
jalencato: We still need to pass something here, because inside the ``DistFeatureTransformer`` class, we have no idea about the current edge_type, mapping_path, format_name, we still need to find some ways to pass them into the feature transformer class.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hard_negative_transformation.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+    node_mapping_length = hard_negative_node_mapping.count()
+
+    # TODO: This method may suffer from scalability issue,
+    # we can make this method to join-based solution.
+    hard_negative_node_mapping_dict = {
+        row[NODE_MAPPING_STR]: row[NODE_MAPPING_INT] for row in hard_negative_node_mapping.collect()
+    }
+
+    # Same length for feature to convert to tensor
```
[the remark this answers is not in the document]
jalencato: I do not think we need add an assertion for this because that is what the UDF function will do. The UDF function will map all the feature input to the same length (same as the number of different destination type).

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -1654,7 +1654,15 @@ def _process_edge_features(
                 .get(edge_type, {})
                 .get(feat_conf.feat_name, {})
             )
-            transformer = DistFeatureTransformer(feat_conf, self.spark, json_representation)
+            # Hard Negative Transformation use case, but should be able to be reused
+            edge_mapping_dict = {
```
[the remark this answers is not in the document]
jalencato: Change to edge-hard-negative only.

python/graphstorm/gpartition/dist_partition_graph.py
jalencato: Remove the config loading code here.

python/graphstorm/gpartition/post_hard_negative.py
jalencato: We can keep the name here, to keep the same as https://github.com/awslabs/graphstorm/blob/19fac3ba136e77c404c551cb59d53b6f79a5b2c3/python/graphstorm/gconstruct/utils.py#L1139. They both have some function.

python/graphstorm/gpartition/post_hard_negative.py
jalencato: Good callout. We can make a change.

tests/unit-tests/gpartition/test_hard_negative_post_partition.py
jalencato: Update to save into the file, as we need to test with loading from files.

python/graphstorm/gpartition/post_hard_negative.py
thvasilo: +1 can we avoid code duplication by making the gconstruct function serve both use-cases?

python/graphstorm/model/utils.py
```diff
@@ -393,6 +393,23 @@ def _exchange_node_id_mapping(rank, world_size, device,
     # move mapping into CPU
     return gather_list[0].to(th.device("cpu"))
 
+def load_dist_nid_map(node_id_mapping_file, ntypes):
```
thvasilo, line 396: Can we just rename the old function instead of adding new one that is just a wrapper?

tests/unit-tests/gpartition/test_hard_negative_post_partition.py
```diff
@@ -0,0 +1,114 @@
...
+from graphstorm.model.utils import load_dist_nid_map
+from dgl.data.utils import load_tensors, save_tensors
+from graphstorm.gpartition.post_hard_negative import (shuffle_hard_negative_nids,
+                                                      load_hard_negative_config)
+
+
+def test_load_hard_negative_config():
+    # For config with hard negative transformation
+    json_file_path = (f"/graphstorm/tests/unit-tests/gpartition/"
+                      f"config/gsprocessing_hard_negative_config.json")
```
thvasilo, line 34: Use paths relative to the location of the test scripts, usually you'd add the static files in a `resources` dir like we do here:

https://github.com/awslabs/graphstorm/blob/19fac3ba136e77c404c551cb59d53b6f79a5b2c3/graphstorm-processing/tests/conftest.py#L34

https://github.com/awslabs/graphstorm/blob/19fac3ba136e77c404c551cb59d53b6f79a5b2c3/graphstorm-processing/tests/conftest.py#L80-L85

tests/unit-tests/gpartition/test_hard_negative_post_partition.py
```diff
@@ -0,0 +1,114 @@
...
+    # For config without hard negative transformation
+    json_file_path = (f"/graphstorm/tests/unit-tests/gpartition/"
+                      f"config/gsprocessing_non_hard_negative_config.json")
+
+    res = load_hard_negative_config(json_file_path)
+
+    assert res == []
+
+
+def test_shuffle_hard_negative_nids(tmp_path):
```
thvasilo, line 50: Let's add a doc string explaining what this test does.

thvasilo, line 50: There's quite a bit of setup for this test is there any way to simplify, or at least move the setup in a fixture so we can focus on the test functionality?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hard_negative_transformation.py
```diff
@@ -0,0 +1,121 @@
...
+    else:
+        transformed_df = input_df
+    # Edge type should be (src_ntype:relation_type:dst_ntype)
+    # Only support hard negative for destination nodes. Get the node type of destination nodes.
+    # TODO: support hard negative for source nodes.
+    _, _, dst_type = hard_node_mapping_dict["edge_type"].split(":")
+    mapping_prefix = hard_node_mapping_dict["mapping_path"]
+    format_name = hard_node_mapping_dict["format_name"]
+    hard_negative_node_mapping = spark.read.parquet(
+        f"{mapping_prefix}{dst_type}/{format_name}/*.parquet"
```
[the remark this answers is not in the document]
jalencato: Yes, these are the mapping created before. I will try if simply read the directory work.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hard_negative_transformation.py
```diff
@@ -0,0 +1,136 @@
...
+        "inner",
+    ).select(NODE_MAPPING_INT, ORDER_INDEX)
+    transformed_df = transformed_df.groupBy(ORDER_INDEX).agg(
+        F.collect_list(NODE_MAPPING_INT).alias(cols[0])
+    )
+
+    # Same length for feature to convert to tensor
+    def pad_mapped_values(hard_neg_list):
+        while len(hard_neg_list) < node_mapping_length:
+            hard_neg_list.append(-1)
```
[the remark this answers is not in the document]
jalencato: Yes, ``node_mapping_length`` is expected to be equal to the total number of nodes for the node type. And we expect the processed result have the same length as the total number of nodes so we can have the same length of tensor later.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hard_negative_transformation.py
```diff
@@ -0,0 +1,121 @@
...
+from pyspark.sql.functions import split, col
+from pyspark.sql.types import ArrayType, IntegerType, StringType
+from pyspark.sql import DataFrame, functions as F, SparkSession
+
+from graphstorm_processing.constants import NODE_MAPPING_STR, NODE_MAPPING_INT
+
+from .base_dist_transformation import DistributedTransformation
+
+
+def apply_transform(
```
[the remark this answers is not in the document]
jalencato: I am referring to the case we did for hugging-face transformation, which we put the function outside the .

python/graphstorm/gpartition/dist_partition_graph.py
jalencato: ``shuffle_hard_negative_nids`` will first check if there is a hard negative transformation, if no matched, skip.

python/graphstorm/gpartition/post_hard_negative.py
jalencato: I will need to make some changes to our previous code repo. Previously I thought better not to do that, as we all want to address the issue here, let me make some changes here.

python/graphstorm/model/utils.py
jalencato: Same as here, let me make the function public.

tests/unit-tests/gpartition/test_hard_negative_post_partition.py
jalencato: Let us make it a fixture.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hard_negative_transformation.py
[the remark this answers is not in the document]
jalencato: Apply the change.

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hard_negative_transformation.py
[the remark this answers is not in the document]
thvasilo: Is it necessary for have a value for every node? How are these ingested downstream? 

If my understanding is correct, if we have 100M nodes for a type, we'd end up with a 100Mx100M matrix for negatives?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hard_negative_transformation.py
```diff
@@ -0,0 +1,123 @@
...
+        # Edge type should be (src_ntype:relation_type:dst_ntype)
+        # Only support hard negative for destination nodes. Get the node type of destination nodes.
+        # TODO: support hard negative for source nodes.
+        _, _, dst_type = self.hard_node_mapping_dict["edge_type"].split(":")
+        mapping_prefix = self.hard_node_mapping_dict["mapping_path"]
+        format_name = self.hard_node_mapping_dict["format_name"]
+        hard_negative_node_mapping = self.spark.read.parquet(
+            f"{mapping_prefix}{dst_type}/{format_name}/"
+        )
+        max_size = (
```
thvasilo, line 86: Add a comment here explaining what max_size (the maximum number of negatives in the input data)

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_hard_negative_transformation.py
```diff
@@ -0,0 +1,123 @@
...
+            transformed_df.select(F.size(F.col(input_col)).alias(f"{input_col}_size"))
+            .agg(F.max(f"{input_col}_size"))
+            .collect()[0][0]
+        )
+
+        # TODO: Use panda series to possibly improve the efficiency
+        # Explode the original list and join node id mapping dataframe
+        transformed_df = transformed_df.withColumn(ORDER_INDEX, F.monotonically_increasing_id())
+        transformed_df = transformed_df.withColumn(
+            EXPLODE_HARD_NEGATIVE_VALUE, F.explode(F.col(input_col))
```
thvasilo, line 96: Could results in extremely large DFs, let's add a comment here. The expected row count will be `num_nodes*avg(len_of_negatives)`

python/graphstorm/gpartition/post_hard_negative.py
```diff
@@ -0,0 +1,109 @@
...
+
+import json
+import os
+
+from dgl.data.utils import load_tensors, save_tensors
+from graphstorm.model.utils import load_dist_nid_map
+from graphstorm.gconstruct.utils import get_gnid2pnid_map
+
+
+def load_hard_negative_config(gsprocessing_config):
```
thvasilo, line 25: Let's start adding type annotations for new code (using typing.Dict etc.)

python/graphstorm/gconstruct/utils.py
```diff
@@ -1118,6 +1118,28 @@ def get_hard_edge_negs_feats(hard_edge_neg_ops):
 
     return hard_edge_neg_feats
 
+def get_gnid2pnid_map(ntype, node_mapping, gnid2pnid_mapping):
```
thvasilo, line 1121: Add type annotations and return docstring

python/graphstorm/gconstruct/utils.py
```diff
@@ -1118,6 +1118,28 @@ def get_hard_edge_negs_feats(hard_edge_neg_ops):
...
+    """ Get global nid to partitioned nid mapping.
+
+        Parameters
+        ----------
+        ntype: str
+            Path to the directory storing the partitioned graph.
```
thvasilo, line 1127: Is this correct?

python/graphstorm/gconstruct/utils.py
```diff
@@ -1118,6 +1118,28 @@ def get_hard_edge_negs_feats(hard_edge_neg_ops):
...
+        node_mapping: dict
```
thvasilo, line 1128: Maybe have a more descriptive name here? The two mappings passed are the same mappings reversed right? `global_nid <-> partition_nid`

python/graphstorm/gconstruct/utils.py
```diff
@@ -1118,6 +1118,28 @@ def get_hard_edge_negs_feats(hard_edge_neg_ops):
...
+            Dict of mapping. {ntype: partitioned nid to global nid mapping}
```
thvasilo, line 1129: Add a few more details for the contents? How are the mappings represented (torch Tensor where row ids mean this and row values mean that)

python/graphstorm/gconstruct/utils.py
jalencato: I would say let us keep the naming for ``node_mapping``, because it is exactly the file name we are loading. But I will add one additional comment here to note what it is here.

Issue #1081, opened 2024-10-30
Add support for mountpoint-s3 on SageMaker tasks


thvasilo:
Because DistDGL and by extension GraphStorm has an assumption of a shared filesystem to function properly, in our SageMaker implementations need to implement various downloads and uploads to "fake" the existence of a shared filesystem, by downloading data locally to specific locations per instance.

This introduces a maintenance burden as we can't make the same environment assumptions for our SageMaker vs. EC2 with EFS execution, and introduces a lot of [glue code](https://proceedings.neurips.cc/paper_files/paper/2015/file/86df7dcfd896fcaf2674f757a2463eba-Paper.pdf), to make the two system compatible.

[Mountpoint for S3](https://github.com/awslabs/mountpoint-s3) is an AWS project that allows entire S3 buckets to mounted onto EC2 instances and treated a (mostly) regular filesystem. If we are able to use S3 buckets as virtual shared filesystems for SageMaker we should be able to simplify and align the codebase. We note the use-cases suggested by the mountpoint-s3 project align with ours:

> Mountpoint for Amazon S3 is optimized for applications that need high read throughput to large objects, potentially from many  clients at once, and to write new objects sequentially from a single client at a time. This means it's a great fit for applications that use a file interface to:
>     - read large objects from S3, potentially from many instances concurrently, without downloading them to local storage first
>     - access only some S3 objects out of a larger data set, but can't predict which objects in advance
>     - upload their output to S3 directly, or upload files from local storage with tools like cp
> 
> but probably not the right fit for applications that:
> 
>     - use file operations that S3 doesn't natively support, like directory renaming or symlinks
>     - ( make edits to existing files (don't work on your Git repository or run vim in Mountpoint 😄)

We propose starting with a POC that modifies our SageMaker images and entry points to use mountpoint-s3, _but does not affect the user-facing launch scripts_, providing a backwards-compatible solution for our users.

Our first target will be adding GraphBolt support to SageMaker DistPartition, which is currently not possible, because DistDGL to GraphBolt partition conversion assumes that the leader instance has access to the entire distributed graph on disk. Following that, we can migrate our other SageMaker tasks to mountpoint-s3, where shared filesystems are normally required:

- [ ] DistPartition, remove download/upload of data from S3 
- [ ] DistTraining
- [ ] DistInference

thvasilo:
[...]

PR #1083, opened 2024-11-01
[SageMaker] [GraphBolt] Add support for launching GraphBolt jobs on SageMaker

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Add a new SageMaker job to convert DistPart data to GraphBolt. This is our only option currently as there's no way to directly use S3 as a writable, shared file system in SageMaker, see https://github.com/awslabs/graphstorm/issues/1081 for details.
* The `sagemaker/launch_graphbolt_convert.py` will launch the SageMaker job, that downloads the entire partitioned graph to one instance, then runs the GB conversion, one partition at a time. Because DGL writes the new fused CSC graph representation in the same directory as the input data, we can't use one of SageMaker's FastFile modes to stream the data, as that creates read-only filesystems.
* [Optional] We also include an example of how one could use a SageMaker Pipeline to run the GSPartition and GBConvert jobs in sequence, but this can be removed (because SageMaker Pipelines are persistent once created).
* Added unit test mechanism to test sagemaker scripts, we start with testing our parsing logic. To make the scripts available to the runner's python runtime we add the `graphstorm/sagemaker/launch` directory to the runner's `PYTHONPATH`.

EDIT: One note about the PR: The changes to the partition launch that use a SageMaker Pipeline are for demonstration purposes, I think I'll remove them alltogether and just have separate partition/gbconvert jobs. But we might want to have an example of how to programmatically build an SM pipeline as an example, e.g. from gsprocessing to training (as SM jobs)

### DIFF
--- a/.github/workflow_scripts/pytest_check.sh
+++ b/.github/workflow_scripts/pytest_check.sh
@@ -1,9 +1,20 @@
-# Move to parent directory
-cd ../../
-
+#!/bin/env bash
+# Move to repository root
 set -ex
 
-FORCE_CUDA=1 python3 -m pip install -e '.[test]'  --no-build-isolation
+cd ../../
+
+
+GS_HOME=$(pwd)
+# Add SageMaker launch scripts to make the scripts testable
+export PYTHONPATH="${PYTHONPATH}:${GS_HOME}/sagemaker/launch/"
+
 python3 -m pip install pytest
+FORCE_CUDA=1 python3 -m pip install -e '.[test]'  --no-build-isolation
+
+# Run SageMaker tests
+python3 -m pytest -x ./tests/sagemaker-tests -s
+
+# Run main library unit tests (Requires multi-gpu instance)
 sh ./tests/unit-tests/prepare_test_data.sh
 export NCCL_IB_DISABLE=1; export NCCL_SHM_DISABLE=1; NCCL_NET=Socket NCCL_DEBUG=INFO python3 -m pytest -x ./tests/unit-tests -s
--- a/docker/sagemaker/Dockerfile.sm
+++ b/docker/sagemaker/Dockerfile.sm
@@ -20,15 +20,16 @@ FROM branch-${DEVICE} AS final
 
 LABEL maintainer="Amazon AI Graph ML team"
 
-# Install related Python packages
+# Install required Python packages,
+# we set the versions to match those of the base conda environment when possible
 RUN pip3 install \
-        boto3 \
-        numba==0.58.1 \
+        boto3==1.34.112 \
+        numba==0.59.1 \
         numpy==1.26.4 \
         ogb==1.3.6 \
-        pyarrow==18.0.0 \
-        scikit-learn \
-        scipy \
+        pyarrow==16.1.0 \
+        scikit-learn==1.5.0 \
+        scipy==1.13.1 \
         transformers==4.28.1 \
     && rm -rf /root/.cache
 
@@ -56,5 +57,5 @@ ENV PYTHONPATH="/root/dgl/tools/:${PYTHONPATH}"
 
 WORKDIR /opt/ml/code
 
-ENTRYPOINT ["bash", "-m", "start_with_right_hostname.sh"]
+ENTRYPOINT ["bash", "-m", "/usr/local/bin/start_with_right_hostname.sh"]
 CMD ["/bin/bash"]
--- a/docker/sagemaker/Dockerfile.sm
+++ b/docker/sagemaker/Dockerfile.sm
@@ -24,9 +24,9 @@ LABEL maintainer="Amazon AI Graph ML team"
 RUN pip3 install \
         boto3 \
         numba==0.58.1 \
-        numpy==1.26.1 \
+        numpy==1.26.4 \
         ogb==1.3.6 \
-        pyarrow \
+        pyarrow==18.0.0 \
         scikit-learn \
         scipy \
         transformers==4.28.1 \
--- a/docker/sagemaker/Dockerfile.sm
+++ b/docker/sagemaker/Dockerfile.sm
@@ -49,7 +49,7 @@ ENV PYTHONPATH="/opt/ml/code/graphstorm/python/:${PYTHONPATH}"
 RUN cp /opt/ml/code/graphstorm/sagemaker/run/* /opt/ml/code/
 
 # Download DGL source code
-RUN cd /root; git clone --branch v${DGL_VERSION} https://github.com/dmlc/dgl.git
+RUN cd /root; git clone --branch v${DGL_VERSION} --single-branch https://github.com/dmlc/dgl.git
 # Un-comment if we prefer a local DGL distribution
 # COPY dgl /root/dgl
 ENV PYTHONPATH="/root/dgl/tools/:${PYTHONPATH}"
--- a/docs/source/advanced/using-graphbolt.rst
+++ b/docs/source/advanced/using-graphbolt.rst
@@ -166,3 +166,94 @@ data using our GraphBolt conversion entry point:
     # We'll see the GraphBolt representation has been re-created
     ls /tmp/acm_graphbolt/part0
     edge_feat.dgl  fused_csc_sampling_graph.pt  graph.dgl  node_feat.dgl
+
+Using GraphBolt on SageMaker
+----------------------------
+
+To prepare data to use with GraphBolt on SageMaker
+we need to launch the GraphBolt data conversion step
+as a separate SageMaker job, after
+the partitioned DGL graph files have been created on S3.
+
+After running your distributed partition SageMaker job as normal using
+``sagemaker/launch_partition.py``, you next need to launch the
+``sagemaker/launch_graphbolt_convert.py`` script, passing as input
+the S3 URI where you created the DistDGL partition data,
+**plus the suffix dist_graph** as that's where GSPartition creates the partition files.
+
+For example, if you used ``--output-data-s3 s3://my-bucket/my-part-graph`` for
+``sagemaker/launch_partition.py`` you need to use ``--graph-data-s3 s3://my-bucket/my-part-graph/dist_graph``
+for ``sagemaker/launch_graphbolt_convert.py``.
+
+So where before the SageMaker job sequence for distributed processing and training
+was ``GSProcessing -> GSPartition -> GSTraining`` to use GraphBolt we need to add
+a step after partitioning and before training:
+``GSProcessing -> GSPartition -> GraphBoltConvert -> GSTraining``.
+
+.. code-block:: bash
+
+    cd graphstorm/sagemaker
+    sagemaker/launch_partition.py \
+        --graph-data-s3 "s3-uri-where-gsprocessing-data-exist" \
+        --output-data-s3 "s3-uri-where-gspartition-data-will-be"
+        # Add other required parameters like --partition-algorithm, --num-instances etc.
+
+    # Once the above job succeeds we run the below to convert the data to GraphBolt.
+    # Note the /dist_graph suffix!
+    sagemaker/launch_graphbolt_convert.py \
+        --graph-data-s3 "s3-uri-where-gspartition-data-will-be/dist_graph" \
+        --metadata-filename "metadata.json" # Or <graph-name>.json for gconstruct-ed partitions
+
+
+If your data are small enough to process on a single SageMaker instance
+using ``GConstuct``, you can simply pass the ``--use-graphbolt true`` argument
+to the ``GConstruct`` SageMaker launch script and that will create the
+necessary GraphBolt files as well.
+So the job sequence there remains ``GConstruct -> GSTraining``.
+
+.. code-block:: bash
+
+    sagemaker/launch_gconstruct.py \
+        --graph-data-s3 "s3-uri-where-raw-data-exist" \
+        --output-data-s3 "s3-uri-where-gspartition-data-will-be" \
+        --graph-config-file "gconstruct-config.json" \
+        --use-graphbolt true
+
+If you initially used GConstruct to create the non-GraphBolt DistDGL files,
+you'll need to pass in the additional argument ``--metadata-filename``
+to ``launch_graphbolt_convert.py``.
+Use ``<graph-name>.json`` where the graph name should correspond to the
+one you used with GConstruct as shown below:
+
+.. code-block:: bash
+
+    sagemaker/launch_gconstruct.py \
+        --graph-name my-graph \
+        --graph-data-s3 "s3-uri-where-raw-data-exist" \
+        --output-data-s3 "s3-uri-where-gspartition-data-will-be" \
+        --graph-config-file "gconstruct-config.json" # We don't add --use-graphbolt true
+
+    # Once the above job succeeds we run the below to convert the data to GraphBolt
+    sagemaker/launch_graphbolt_convert.py \
+        --graph-data-s3 "s3-uri-where-gspartition-data-will-be"
+        --metadata-filename "my-graph.json" # Should be <graph-name>.json
+
+
+Once the data have been converted to the GraphBolt format you can run your training
+and inference jobs as before, passing the additional
+argument ``--use-graphbolt`` to the SageMaker launch scripts
+to indicate that we want to use GraphBolt during training/inference:
+
+.. code-block:: bash
+
+    sagemaker/launch_train.py \
+        --graph-name my-graph \
+        --graph-data-s3 "s3-uri-where-gspartition-data-will-be" \
+        --yaml-s3 "s3-path-to-train-yaml" \
+        --use-graphbolt true
+
+
+Internally we use the unknown arguments mechanism of Python's
+``argparse`` module to forward
+the ``--use-graphbolt`` argument to the downstream job, along
+with other GraphStorm arguments.
--- a/docs/source/advanced/using-graphbolt.rst
+++ b/docs/source/advanced/using-graphbolt.rst
@@ -186,7 +186,7 @@ the partitioned DGL graph files have been created on S3.
 After running your distributed partition SageMaker job as normal using
 ``sagemaker/launch_partition.py``, you next need to launch the
 ``sagemaker/launch_graphbolt_convert.py`` script, passing as input
-the S3 URI where the DistDGL partition data is stored by ``launch_partition.py``,
+the S3 URI, where the DistDGL partition data is stored by ``launch_partition.py``,
 **plus the suffix `dist_graph`** as that's where GSPartition creates the partition files.
 
 For example, if you used ``--output-data-s3 s3://my-bucket/my-part-graph`` for
--- a/docs/source/advanced/using-graphbolt.rst
+++ b/docs/source/advanced/using-graphbolt.rst
@@ -170,7 +170,15 @@ data using our GraphBolt conversion entry point:
 Using GraphBolt on SageMaker
 ----------------------------
 
-To prepare data to use with GraphBolt on SageMaker
+Before being able to train on SageMaker
+we need to ensure our data on S3 have been
+converted to the GraphBolt format.
+When using GConstruct to process our data
+we can include the GraphBolt data conversion in the GConstruct
+step as we'll show below.
+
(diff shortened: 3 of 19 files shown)


docs/source/advanced/using-graphbolt.rst
```diff
@@ -166,3 +166,94 @@ data using our GraphBolt conversion entry point:
     # We'll see the GraphBolt representation has been re-created
     ls /tmp/acm_graphbolt/part0
     edge_feat.dgl  fused_csc_sampling_graph.pt  graph.dgl  node_feat.dgl
+
+Using GraphBolt on SageMaker
+----------------------------
+
+To prepare data to use with GraphBolt on SageMaker
```
jalencato, line 173: That should be necessary whether you are using the distributed pipeline or gconstruct?

python/graphstorm/gpartition/convert_to_graphbolt.py
```diff
@@ -27,24 +27,67 @@
...
+            part_config,
+            store_eids=True,
+            graph_formats="coo",
+        )
+    else:
+        dgl_distributed.dgl_partition_to_graphbolt(  # pylint: disable=unexpected-keyword-arg
+            part_config,
+            store_eids=True,
+            graph_formats="coo",
+            njobs=njobs,
```
jalencato, line 80: Do we test that with DGL >= 2.4.0?

sagemaker/launch/common_parser.py
```diff
@@ -49,7 +49,7 @@ def get_common_parser() -> argparse.ArgumentParser:
     common_args.add_argument("--instance-type", type=str,
         help="instance type for the SageMaker job")
     common_args.add_argument("--instance-count", type=int,
-        default=2,
+        default=1,
```
jalencato, line 52: In any case we will use the common parser default? I assume we will only call the script via launch_TASK.py.

sagemaker/launch/launch_train.py
```diff
@@ -109,7 +109,7 @@ def run_job(input_args, image, unknowargs):
         py_version="py3",
         base_job_name=prefix,
         hyperparameters=params,
-        sagemaker_session=sess,
+        # sagemaker_session=sess,
```
jalencato, line 112: Why we comment here?

sagemaker/launch/launch_partition.py
```diff
@@ -55,33 +58,92 @@ def run_job(input_args, image):
...
+        role_arn=role,
+        description="GraphStorm DistPartition"
+    )
+
+    execution = pipeline.start()
+
+    print(execution.describe())
+
+    # if not input_args.async_execution:
+    #     execution.wait()
```
jalencato, line 144: Delete the comment code.

sagemaker/launch/launch_partition.py
```diff
@@ -1,11 +1,14 @@
...
 import logging
+import os
 from time import strftime, gmtime
 
 import boto3 # pylint: disable=import-error
 from sagemaker.processing import ScriptProcessor
-import sagemaker
+from sagemaker.workflow.pipeline import Pipeline, PipelineDefinitionConfig
+from sagemaker.workflow.pipeline_context import PipelineSession
+from sagemaker.workflow.steps import ProcessingStep, ProcessingInput
```
jalencato, line 11: A quick question about the session here: since I see in the overall description we are moving our partition pipeline to the sagemaker pipeline. Will it affect our previous usage? Like if we can still go the SageMaker processing job console to check the job status. Will it affect any user experience?

jalencato: I see all the previous ScriptProcessor parts are removed.

thvasilo: I added this as a demonstration, will probably remove. from the pr description:

>One note about the PR: The changes to the partition launch that use a SageMaker Pipeline are for demonstration purposes, I think I'll remove them alltogether and just have separate partition/gbconvert jobs. But we might want to have an example of how to programmatically build an SM pipeline as an example, e.g. from gsprocessing to training (as SM jobs)

tests/sagemaker-tests/test_sagemaker_args.py
thvasilo, line 1: Note that I used Amazon Q to seed these tests and then edited by hand.

sagemaker/launch/launch_gconstruct.py
```diff
@@ -105,9 +104,7 @@ def get_gconstruct_parser():
 
     gconstruct_arguments = parser.add_argument_group("GraphStorm Gconstruct arguments")
 
-    gconstruct_arguments.add_argument("--input-graph-s3", type=str,
-        required=True, help="S3 location of the input graph data")
-    gconstruct_arguments.add_argument("--output-graph-s3", type=str,
+    gconstruct_arguments.add_argument("--output-graph-s3", "--output-data-s3", type=str,
```
classicsong, line 107: Is this a breaking change? If yes, please flag it and update the PR description.

sagemaker/launch/launch_train.py
```diff
@@ -75,24 +80,12 @@ def run_job(input_args, image, unknowargs):
...
-                break
-            sub_params.append(unknowargs[i])
-        params[unknowargs[unknow_idx][2:]] = ' '.join(sub_params)
-        unknow_idx = i
-
-    print(f"Parameters {params}")
-    print(f"GraphStorm Parameters {unknowargs}")
+    unknown_args_dict = parse_unknown_gs_args(unknowargs)
+    if len(unknown_args_dict) > 0:
+        params.update(unknown_args_dict)
```
classicsong, line 85: If unknown_args_dict is always a dict, we can directly do `params.update(unknown_args_dict)`.
`params.update({})` will not change params.

sagemaker/launch/launch_train.py
classicsong: Same question.

sagemaker/run/gconstruct_entry.py
```diff
@@ -32,12 +32,15 @@ def parse_construct_args():
...
         required=True, help="Graph name")
-    return parser
+
+    return parser.parse_known_args()
 
 
 if __name__ =='__main__':
-    parser = parse_construct_args()
-    args, unknownargs = parser.parse_known_args()
+    args, unknownargs = parse_construct_args()
```
classicsong, line 40: Why change the code?

tests/sagemaker-tests/test_sagemaker_args.py
```diff
@@ -0,0 +1,62 @@
...
+@pytest.mark.parametrize("input_args, expected_output", [
+    (['--num', '1', '2', '3', '--str', 'a', 'b', 'c'],
+     {'num': '1 2 3', 'str': 'a b c'}),
+    (['--mixed', 'str', '123', 'true', '--flag'],
+     {'mixed': 'str 123 true', 'flag': ''}),
+    (['--empty', '', '--zero', '0', '--false', 'false'],
+     {'empty': '', 'zero': '0', 'false': 'false'}),
+    (['--special-chars', '!@#$%^&*()'],
+     {'special-chars': '!@#$%^&*()'}),
+    (['--unicode', 'こんにちは', 'world'],
```
classicsong, line 58: Not sure why we need this test.

tests/sagemaker-tests/test_sagemaker_args.py
```diff
@@ -0,0 +1,62 @@
...
+    args = ['--arg1', 'value1', '--flag', '--arg2', 'value2a', 'value2b']
+    expected = {'arg1': 'value1', 'flag': '', 'arg2': 'value2a value2b'}
+    assert dict(parse_unknown_gs_args(args)) == expected
+
+def test_quoted_args():
+    args = ['--arg', '"quoted value"', '--another-arg', "'single quoted'"]
+    expected = {'arg': '"quoted value"', 'another-arg': "'single quoted'"}
+    assert dict(parse_unknown_gs_args(args)) == expected
+
+@pytest.mark.parametrize("input_args, expected_output", [
```
classicsong, line 49: I think the above tests have covered all the possible cases.

sagemaker/launch/launch_partition.py
classicsong: If this is just for demonstration purpose, I would suggest we do not directly merge it into main branch. We can create a dev branch first.

classicsong: The code here can be a good example of using SageMaker Pipeline to automate the SageMaker tasks.

.github/workflow_scripts/pytest_check.sh
```diff
@@ -1,9 +1,20 @@
...
-FORCE_CUDA=1 python3 -m pip install -e '.[test]'  --no-build-isolation
+GS_HOME=$(pwd)
+# Add SageMaker launch scripts to make the scripts testable
+export PYTHONPATH="${PYTHONPATH}:${GS_HOME}/sagemaker/launch/"
+
 python3 -m pip install pytest
+FORCE_CUDA=1 python3 -m pip install -e '.[test]'  --no-build-isolation
+
+# Run SageMaker tests
+python3 -m pytest -x ./tests/sagemaker-tests -s
```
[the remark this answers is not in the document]
thvasilo: I was thinking to have them before because they take trivial time so for someone only touching SageMaker code they can fail early and not wait an hour for the main tests to run?

docker/sagemaker/Dockerfile.sm
```diff
@@ -24,9 +24,9 @@ LABEL maintainer="Amazon AI Graph ML team"
 RUN pip3 install \
         boto3 \
         numba==0.58.1 \
-        numpy==1.26.1 \
+        numpy==1.26.4 \
```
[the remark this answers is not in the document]
thvasilo: Not necessary, but this is the version that gets installed by default now, what happens is that during graphstorm installation 1.26.1 gets removed and 1.26.4 gets installed. So we just make sure it's installed in advance here.

python/graphstorm/gpartition/convert_to_graphbolt.py
thvasilo: I have tested it in our local images, for SageMaker it's an early placement, so we have the functionality once we move the SM images to 2.4.0.

sagemaker/launch/common_parser.py
thvasilo: All `launch_*.py` tasks use the common parser, so they use the default here. Since now we have at least one job that requires a single instance I thought we should change the default.

Ideally I'd prefer this to be a required argument but we can discuss that elsewhere.

sagemaker/launch/launch_gconstruct.py
thvasilo: It's not breaking, we're just adding an alias to the call to make the gconstruct arguments consistent with the other launch scripts. Other scripts use `graph-data-s3` and `output-data-s3`, only gconstruct uses `input-graph-s3` and `output-graph-s3`. So adding an alias here and in the common parser for the input so both are supported.

sagemaker/launch/launch_partition.py
thvasilo: Yes, I'll create another revision of this PR that does that, and leave this change in a dev branch on my fork. In the future we can add a separate launch script that creates an end2end SageMaker pipeline.

sagemaker/launch/launch_train.py
thvasilo: This allows us to launch and test SageMaker jobs using SageMaker [local mode](https://sagemaker.readthedocs.io/en/stable/overview.html#local-mode) by simply setting the instance-type to `local`. I used this functionality to quickly test my changes locally before launching actual SM jobs.

sagemaker/run/gconstruct_entry.py
thvasilo: The intention is to match function outcome to its name, `parse_construct_args` implies the function parses arguments, but previously it was returning a parser. In addition, the `parser` var in the function was shadowing the `parser` var inside the main function so pylance was reporting a warning.

docs/source/advanced/using-graphbolt.rst
jalencato: We can call it out that if user wants to use graphbolt, they can do all of them in one job for GConstruct, but one additional step for distributed partition pipeline.

.github/workflow_scripts/pytest_check.sh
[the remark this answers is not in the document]
classicsong: But most of the time, we do not change the SageMaker code.
Does unit test take an hour?

tests/sagemaker-tests/test_sagemaker_args.py
```diff
@@ -1,62 +1,88 @@
...
+    args = ["--num-epochs 1 --use-graphbolt true"]
+    expected = {"num-epochs": "1", "use-graphbolt": "true"}
+    assert dict(parse_unknown_gs_args(args)) == expected
+
+
+def test_parse_complex_string_with_quotes():
+    args = [
+        (
+            "--target-etype query,clicks,asin query,search,asin "
+            "--feat-name ntype0:feat0 ntype1:feat1"
```
classicsong, line 81: There is no quotes in the input string.

classicsong (commented):
How do you launch the convert2graphbolt task?

sagemaker/launch/common_parser.py
```diff
@@ -91,22 +91,34 @@ def parse_estimator_kwargs(arg_string: str) -> Dict[str, Any]:
...
+    Returns
+    -------
+    Dict[str, str]
+        Dictionary of parsed arguments {'arg_name': 'arg_value'}
     """
     unknown_args_dict = {}
     current_arg_name = None
 
+    # Handle case where all args were parsed as a single string
+    if len(unknown_args) == 1 and unknown_args[0].count("--") > 1:
```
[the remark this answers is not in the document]
thvasilo: Ah good point, I'll update the code and add a test.

tests/sagemaker-tests/test_sagemaker_args.py
thvasilo: The quoting refers to the CLI call in bash, like `python launch_*.py --instance-type [ml...] '--feat-name ntype0:feat0 ntype1:feat1'`

This happened e.g. in my training automation script where my call was


```bash
TRAIN_YAML_S3="s3://my-bucket/yaml-files/acm/acm_nc.yaml"
GSF_ARGS="--num-epochs 1 --use-graphbolt true"

python3 $SCRIPT_DIR/launch/launch_train.py \
    --entry-point run/train_entry.py \
    --yaml-s3 ${TRAIN_YAML_S3} \
    # Other launch args...
    "${GSF_ARGS}" # Note the double quotes
```

Generally it's good practice to use the double quotes in bash vars where we might have spaces and special characters: https://www.shellcheck.net/wiki/SC2086 so we should support this case.

jalencato (commented):
In overall I think we need to run separate sagemaker regression test before we actually merge the code to make sure there is no other backward compatibility problem.

python/graphstorm/sagemaker/sagemaker_partition.py
```diff
@@ -255,7 +255,7 @@ def run_partition(job_config: PartitionJobConfig):
     for key, val in os.environ.items():
         logging.debug("%s: %s", key, val)
 
-    leader_addr = socket.gethostbyname('algo-1')
+    leader_addr = socket.gethostbyname(sorted(hosts)[0])
```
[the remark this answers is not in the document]
thvasilo: It doesn't really matter which one we choose as long as all instances choose the same. This guarantees this will happen, previously `'algo-1'` was causing errors in local mode because instances are named differently there.

sagemaker/launch/launch_graphbolt_convert.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    Get GraphStorm partition task parser.
+    """
+    parser = get_common_parser()
+
+    partition_args = parser.add_argument_group("GraphStorm Partition Arguments")
+
+    partition_args.add_argument("--entry-point", type=str,
+        default="run/gb_convert_entry.py",
+        help="PATH-TO graphstorm/sagemaker/run/partition_entry.py")
```
classicsong, line 95:
```suggestion
        help="PATH-TO graphstorm/sagemaker/run/gb_convert_entry.py")
```

sagemaker/launch/launch_graphbolt_convert.py
```diff
@@ -0,0 +1,124 @@
...
+
+    partition_args.add_argument("--gb-convert-njobs", type=int, default=1,
+        help="Number of parallel processes to use for GraphBolt partition conversion.")
+
+    partition_args.add_argument("--metadata-filename", type=str, default="metadata.json",
+        help="Partition metadata file name. Default: 'metadata.json'. If GConstruct was "
+            "used should be set to <graph-name>.json")
+
```
classicsong, line 103: remove empty line

sagemaker/launch/launch_graphbolt_convert.py
```diff
@@ -0,0 +1,124 @@
...
+
+    partition_args.add_argument('--log-level', default='INFO',
```
classicsong, line 105: You use print in this file. Shall we change them to logging.xxx?

sagemaker/launch/launch_graphbolt_convert.py
```diff
@@ -0,0 +1,124 @@
...
+def get_partition_parser():
+    """
+    Get GraphStorm partition task parser.
+    """
+    parser = get_common_parser()
+
+    partition_args = parser.add_argument_group("GraphStorm Partition Arguments")
+
+    partition_args.add_argument("--entry-point", type=str,
+        default="run/gb_convert_entry.py",
```
classicsong, line 94: I did not see `gb_convert_entry.py`

sagemaker/launch/launch_graphbolt_convert.py
```diff
@@ -0,0 +1,124 @@
...
+    Parameters
+    ----------
+    input_args:
+        Input arguments
+    image: str
+        ECR image uri
+    """
+    timestamp = strftime("%Y-%m-%d-%H-%M-%S", gmtime())
+    # SageMaker base job name
+    sm_task_name = input_args.task_name if input_args.task_name else timestamp
```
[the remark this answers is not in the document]
thvasilo: Actually we do this with base_job_name on line 55. I'll make a change so we use the user-provided string if provided, otherwise just have `gbconvert` as base name, and then SM automatically adds a timestap to ensure uniqueness

docs/source/advanced/using-graphbolt.rst
```diff
@@ -166,3 +166,103 @@ data using our GraphBolt conversion entry point:
...
+If you initially used GConstruct to create the non-GraphBolt DistDGL files,
+you'll need to pass in the additional argument ``--metadata-filename``
+to ``launch_graphbolt_convert.py``.
+Use ``<graph-name>.json`` where the graph name should be the
+one you used with GConstruct as shown below:
+
+.. code-block:: bash
+
+    # NOTE: we provide 'my-graph' as the graph name
+    sagemaker/launch_gconstruct.py \
```
jalencato, line 239: shall we add python3 here? Same for all the commands here. Otherwise user need to type their own python3 in their command.

sagemaker/run/gb_convert_entry.py
```diff
@@ -0,0 +1,75 @@
...
+
+    # Read processing config file
+    with open("/opt/ml/config/processingjobconfig.json", "r", encoding="utf-8") as f:
+        processing_config = json.load(f)
+
+    processing_inputs: list[dict] = processing_config["ProcessingInputs"]
+
+    # Find the S3 input path from the job's configuration
+    for input_entry in processing_inputs:
+        if input_entry["InputName"] == "dist_graph_s3_input":
```
classicsong, line 57: What if the user changes the s3 path?

sagemaker/run/gb_convert_entry.py
```diff
@@ -0,0 +1,75 @@
...
+            local_partition_path = input_entry["S3Input"]["LocalPath"]
+            s3_partition_path = input_entry["S3Input"]["S3Uri"]
```
classicsong, line 59: Are we expecting the users will not change the data in the middle?

classicsong (approved):

PR #1090, opened 2024-11-13
Fix GConstruct file parsing when input is a list with a single wildcard path

thvasilo:
*Issue #, if available:*

* Reported by customer

*Description of changes:*

* Previously if a user passed in a list of input files with a single wildcard path the path was treated as a regular (non-wildcard) path, leading to file not found error.
* We make the file parsing consistent by using the `expand_wildcard` and always passing a list of str to it.
* Add a test case where we used to fail

### DIFF
--- a/python/graphstorm/gconstruct/file_io.py
+++ b/python/graphstorm/gconstruct/file_io.py
@@ -21,6 +21,7 @@ import glob
 import json
 import os
 import logging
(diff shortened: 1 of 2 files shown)


jalencato (approved):
LGTM

classicsong (approved):

PR #1091, opened 2024-11-13
[GSProcessing] Fix parsing of file paths that start with "./"

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Some GConstruct or GSProcessing config files from customers could have file paths that start with `./` which works when input is local but break for S3. We fix this by removing the `./` prefix if it exists when parsing the input files.
* Add change to config we use for testing to verify the change works (although we'd probably need a more S3-specific test)

### DIFF
--- a/graphstorm-processing/tests/resources/small_heterogeneous_graph/gsprocessing-config.json
+++ b/graphstorm-processing/tests/resources/small_heterogeneous_graph/gsprocessing-config.json
@@ -17,7 +17,7 @@
                 "data": {
                     "format": "csv",
                     "files": [
-                        "nodes/movie.csv"
+                        "./nodes/movie.csv"
                     ],
                     "separator": ","
                 },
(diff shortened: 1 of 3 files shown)


graphstorm-processing/tests/resources/small_heterogeneous_graph/gsprocessing-config.json
```diff
@@ -104,7 +104,7 @@
                 "data": {
                     "format": "csv",
                     "files": [
-                        "edges/movie-included_in-genre.csv"
+                        "./edges/movie-included_in-genre.csv"
```
jalencato, line 107: Shall we leave one as `./` and keep another as the same?

thvasilo: We have other paths in the doc that don't include a `./` like 

```
 "data": {
                    "format": "csv",
                    "files": [
                        "nodes/user.csv"
                    ],
                    "separator": ","
                },
```

in lines 28-34

PR #1092, opened 2024-11-14
[Bugfix] Fix the bug when _get_sparse_emb_range return a start larger than num-of-embs

classicsong:
*Issue #, if available:*
In _get_sparse_emb_range, It is possible that start is also larger than num_embs. For example, when `num_embs = 11`, `world_size = 8`, and `rank is 7`, `start = 7 * 2`, which is `14 > 11`.

*Description of changes:*

### DIFF
--- a/python/graphstorm/model/utils.py
+++ b/python/graphstorm/model/utils.py
@@ -146,6 +146,13 @@ def _get_sparse_emb_range(num_embs, rank, world_size):
         world_size : int
             World size in a distributed environment. This tells the size of a distributed cluster
             (How many processes in a cluster).
+
+        Return
+        ------
+        Tuple of (int, int)
+            A tuple storing the start idx (inclusive) and end
+            idx (exclusive) of the target emb range regarding
+            to the rank and world_size.
     """
     assert rank < world_size, \
         "local rank {rank} shold be smaller than world size {world_size}"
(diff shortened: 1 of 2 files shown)


tests/unit-tests/test_model_save_load.py
```diff
@@ -57,6 +57,19 @@ def test_get_sparse_emb_range():
     assert start == 12
     assert end == 15
 
+    # num_embs = 15
```
thvasilo, line 60: Remove?

thvasilo (approved):

python/graphstorm/model/utils.py
```diff
@@ -156,7 +156,12 @@ def _get_sparse_emb_range(num_embs, rank, world_size):
     else:
         start = rank * math.ceil(num_embs / world_size)
         end = (rank + 1) * math.ceil(num_embs / world_size)
-        end = num_embs if rank + 1 == world_size else end
+        # It is possible that start is also larger than num_embs
+        # For example, when num_embs = 11, world_size = 8
+        # and rank is 7
+        # start = 7 * 2, which is 14 > 11.
+        start = num_embs if start > num_embs else start
+        end = num_embs if end > num_embs else end
```
thvasilo, line 164: This might result in a skewed assignment but I guess that's OK.

classicsong: Yeah. it is not a big problem.

PR #1099, opened 2024-11-21
[Bug Fix] GSProcessing Value Assign Disorder

jalencato:
*Issue #, if available:*

*Description of changes:*


thvasilo:
Changing to a draft until we verify root cause

jalencato:
@thvasilo May double check the PR again as the PR now have been work successfully with the test dataset. The trick is that we also should not only maintain the output df order but also input df order.

thvasilo:
CI seems to have issues with latest poetry, let's merge https://github.com/awslabs/graphstorm/pull/1131 first and run tests again @jalencato

classicsong:
Shall we close this PR?

PR #1100, opened 2024-11-24
Add unit tests for load_sparse_emb and save_sparse_emb

classicsong:
*Issue #, if available:*
Add unit tests for load_sparse_emb and save_sparse_emb

*Description of changes:*

### DIFF
--- a/tests/unit-tests/test_model_save_load.py
+++ b/tests/unit-tests/test_model_save_load.py
@@ -31,6 +31,9 @@ from graphstorm.model.utils import _get_sparse_emb_range
 from graphstorm.model.utils import pad_file_index
 from graphstorm import get_node_feat_size
 
+from graphstorm.model.utils import load_sparse_emb, save_sparse_emb
+import graphstorm.model.utils as utils
+
 from data_utils import generate_dummy_dist_graph
 
 def test_get_sparse_emb_range():
@@ -201,7 +204,62 @@ def test_sparse_embed_load_corner_cases():
     th.distributed.destroy_process_group()
     dgl.distributed.kvstore.close_kvstore()
 
+class DummySparseEmb:
+    """ Dummy class for sparse embedding
+    """
+    def __init__(
+        self,
+        shape: tuple,
+        dtype: th.dtype = th.float,
+    ):
+        self._tensor = th.zeros(shape, dtype=dtype)
+
+    @property
+    def num_embeddings(self):
+        return len(self._tensor)
+
+    def __setitem__(self, idx: th.Tensor, val: th.Tensor):
+        self._tensor[idx] = val
+
+    def __getitem__(self, idx: th.Tensor) -> th.Tensor:
+        return self._tensor[idx]
+
+@pytest.mark.parametrize("emb_size", [(140, 10), (1, 10), (13, 10)])
+@pytest.mark.parametrize("world_size", [80, 1, 12])
+def test_save_load_sparse_emb(emb_size, world_size, monkeypatch):
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        emb_path = os.path.join(tmpdirname, "path")
+
+
+        sparse_emb = DummySparseEmb(emb_size)
+        sparse_emb[th.arange(emb_size[0])] = th.rand(emb_size)
+        ntype = "test"
+        world_size = 80
+        def mock_get_world_size():
+            return world_size
+
+        for rank in range(world_size):
+            def mock_get_rank():
+                return rank
+
+            monkeypatch.setattr(utils , "get_world_size", mock_get_world_size)
+            monkeypatch.setattr(utils , "get_rank", mock_get_rank)
+            utils.save_sparse_emb(emb_path, sparse_emb, ntype)
+
+        load_sparse_emb = DummySparseEmb(emb_size)
+        load_emb_path = os.path.join(emb_path, ntype)
+        for rank in range(world_size):
+            def mock_get_rank():
+                return rank
+
+            monkeypatch.setattr(utils, "get_world_size", mock_get_world_size)
+            monkeypatch.setattr(utils, "get_rank", mock_get_rank)
+            utils.load_sparse_emb(load_sparse_emb, load_emb_path)
+
+        assert_equal(load_sparse_emb._tensor.numpy(), sparse_emb.numpy())
+
 if __name__ == '__main__':
+    test_save_load_sparse_emb()
     test_get_sparse_emb_range()
     test_sparse_embed_save(4)
     test_sparse_embed_load(3, 8)
--- a/tests/unit-tests/test_model_save_load.py
+++ b/tests/unit-tests/test_model_save_load.py
@@ -19,6 +19,7 @@ import os
 import dgl
 import pytest
 import tempfile
+from typing import Tuple
 
 import torch as th
 from numpy.testing import assert_equal
@@ -31,7 +32,6 @@ from graphstorm.model.utils import _get_sparse_emb_range
 from graphstorm.model.utils import pad_file_index
 from graphstorm import get_node_feat_size
 
-from graphstorm.model.utils import load_sparse_emb, save_sparse_emb
 import graphstorm.model.utils as utils
 
 from data_utils import generate_dummy_dist_graph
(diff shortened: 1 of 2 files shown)


tests/unit-tests/test_model_save_load.py
```diff
@@ -201,7 +204,78 @@ def test_sparse_embed_load_corner_cases():
...
+@pytest.mark.parametrize("world_size", [80, 1, 12])
+def test_save_load_sparse_emb(emb_size, world_size, monkeypatch):
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        emb_path = os.path.join(tmpdirname, "path")
+
+
+        sparse_emb = DummySparseEmb(emb_size)
+        sparse_emb[th.arange(emb_size[0])] = th.rand(emb_size)
+        ntype = "test"
+        world_size = 80
```
thvasilo, line 237: Won't this override the parametrized value?

tests/unit-tests/test_model_save_load.py
```diff
@@ -201,7 +204,78 @@ def test_sparse_embed_load_corner_cases():
...
+            monkeypatch.setattr(utils, "get_rank", mock_get_rank)
+            utils.load_sparse_emb(load_sparse_emb, load_emb_path)
+
+        assert_equal(load_sparse_emb._tensor.numpy(), sparse_emb._tensor.numpy())
+
+
+        load_sparse_emb2 = DummySparseEmb(emb_size)
+        load_emb_path = os.path.join(emb_path, ntype)
+
+        world_size = 2
```
thvasilo, line 265: Same here, overriding parametrized value?

tests/unit-tests/test_model_save_load.py
```diff
@@ -201,7 +204,78 @@ def test_sparse_embed_load_corner_cases():
     th.distributed.destroy_process_group()
     dgl.distributed.kvstore.close_kvstore()
 
+class DummySparseEmb:
+    """ Dummy class for sparse embedding
+    """
+    def __init__(
+        self,
+        shape: tuple,
```
thvasilo, line 212: Use `from typing import Tuple` to retain support for Python 3.8, literal type annotations were introduced in 3.9

tests/unit-tests/test_model_save_load.py
```diff
@@ -201,7 +204,78 @@ def test_sparse_embed_load_corner_cases():
...
+        ntype = "test"
+        world_size = 80
+        def mock_get_world_size():
+            return world_size
+
+        for rank in range(world_size):
+            def mock_get_rank():
+                return rank
+
+            monkeypatch.setattr(utils , "get_world_size", mock_get_world_size)
```
thvasilo, line 245: Do we need to set this here or can it be done outside the for loop?

tests/unit-tests/test_model_save_load.py
```diff
@@ -201,7 +204,78 @@ def test_sparse_embed_load_corner_cases():
...
+            monkeypatch.setattr(utils , "get_rank", mock_get_rank)
+            utils.save_sparse_emb(emb_path, sparse_emb, ntype)
```
thvasilo, line 247: I noticed there's a different implementation when `if isinstance(sparse_emb, WholeGraphDistTensor)` in the implementation of `save_sparse_emb` are we covering both cases with the test?

tests/unit-tests/test_model_save_load.py
```diff
@@ -201,7 +204,78 @@ def test_sparse_embed_load_corner_cases():
...
+
+        load_sparse_emb = DummySparseEmb(emb_size)
```
thvasilo, line 249: This will shadow the imported `load_sparse_emb` function. Can we use a different name to not confuse the two?

tests/unit-tests/test_model_save_load.py
```diff
@@ -201,7 +204,78 @@ def test_sparse_embed_load_corner_cases():
...
+
+        world_size = 2
+        for rank in range(world_size):
+            def mock_get_rank():
+                return rank
+
+            monkeypatch.setattr(utils, "get_world_size", mock_get_world_size)
+            monkeypatch.setattr(utils, "get_rank", mock_get_rank)
+            utils.load_sparse_emb(load_sparse_emb2, load_emb_path)
+            print("aa")
```
thvasilo, line 273: Remove

tests/unit-tests/test_model_save_load.py
```diff
@@ -201,7 +204,78 @@ def test_sparse_embed_load_corner_cases():
...
+        load_emb_path = os.path.join(emb_path, ntype)
+
+        world_size = 2
+        for rank in range(world_size):
+            def mock_get_rank():
+                return rank
+
+            monkeypatch.setattr(utils, "get_world_size", mock_get_world_size)
+            monkeypatch.setattr(utils, "get_rank", mock_get_rank)
+            utils.load_sparse_emb(load_sparse_emb2, load_emb_path)
```
thvasilo, line 272: Do we need to do the mocking twice? I thought doing it above should work for both cases

tests/unit-tests/test_model_save_load.py
```diff
@@ -31,6 +31,9 @@
 from graphstorm.model.utils import pad_file_index
 from graphstorm import get_node_feat_size
 
+from graphstorm.model.utils import load_sparse_emb, save_sparse_emb
```
thvasilo, line 34: These are unused

tests/unit-tests/test_model_save_load.py
```diff
@@ -201,7 +204,78 @@ def test_sparse_embed_load_corner_cases():
...
+        sparse_emb = DummySparseEmb(emb_size)
+        sparse_emb[th.arange(emb_size[0])] = th.rand(emb_size)
+        ntype = "test"
+        world_size = 80
+        def mock_get_world_size():
+            return world_size
+
+        for rank in range(world_size):
+            def mock_get_rank():
+                return rank
```
thvasilo, line 243: Are we sure we're getting different values for rank here? A variable defined in a function closure will get the same value for all closures in the loop, see https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/cell-var-from-loop.html

tests/unit-tests/test_model_save_load.py
```diff
@@ -201,7 +204,78 @@ def test_sparse_embed_load_corner_cases():
     th.distributed.destroy_process_group()
     dgl.distributed.kvstore.close_kvstore()
 
+class DummySparseEmb:
```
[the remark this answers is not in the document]
classicsong: No.
Using DistEmbedding requires to launch dist server which is too much to the unit test.
The code is to test the logic of GraphStorm, so we can mock the embedding interface.

tests/unit-tests/test_model_save_load.py
classicsong: good catch

tests/unit-tests/test_model_save_load.py
```diff
@@ -201,7 +204,78 @@ def test_sparse_embed_load_corner_cases():
...
+    def __setitem__(self, idx: th.Tensor, val: th.Tensor):
+        self._tensor[idx] = val
+
+    def __getitem__(self, idx: th.Tensor) -> th.Tensor:
+        return self._tensor[idx]
+
+@pytest.mark.parametrize("emb_size", [(140, 10), (1, 10), (13, 10)])
+@pytest.mark.parametrize("world_size", [80, 1, 12])
+def test_save_load_sparse_emb(emb_size, world_size, monkeypatch):
+    with tempfile.TemporaryDirectory() as tmpdirname:
```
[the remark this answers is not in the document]
classicsong: I think either way is fine.

tests/unit-tests/test_model_save_load.py
classicsong: It can be done outside the loop.
Let me move it.

tests/unit-tests/test_model_save_load.py
classicsong: I did not add the test for WholeGraphDistTensor.
Testing wholegraph tensor needs to enable WholeGraph which is not supported in unit test.

tests/unit-tests/test_model_save_load.py
classicsong: Yes. It is going to test the case when saving and loading are using different number of processes.

Issue #1093, opened 2024-11-14
Add "standard" transformation for GConstruct 


thvasilo:
GSProcessing supports a "standard" transformation where each value in a column is divided by the sum of the column. We should provide the same transformation for GConstruct.

PR #1101, opened 2024-11-26
Add standard transform for gconstruct

classicsong:
*Issue #, if available:*
#1093 

*Description of changes:*
Add standard transformation for gconstruct.

### DIFF
--- a/docs/source/cli/graph-construction/single-machine-gconstruct.rst
+++ b/docs/source/cli/graph-construction/single-machine-gconstruct.rst
@@ -147,6 +147,15 @@ GraphStorm provides a set of transformation operations for different types of fe
                   "max_bound": 2.,
                   "min_bound": -2.}
 
+* **Numerical standard transformation** normalizes numerical input features with `val = val / sum`, where `val` is the feature value and `sum` is a summation of all the values in the feature. The ``name`` field in the feature transformation dictionary is ``standard``. The dictionary can contain one optional field, i.e., ``sum``, which is summation of all the values in the feature col from the previous transformation.
+
+  Example:
+
+  .. code:: json
+
+    "transform": {"name": "standard",
+                  "sum": 100.1,}
+
 * **Numerical Rank Gauss transformation** normalizes numerical input features with rank gauss normalization. It maps the numeric feature values to gaussian distribution based on ranking. The method follows the description in the normalization section of `the Porto Seguro's Safe Driver Prediction kaggle competition <https://www.kaggle.com/c/porto-seguro-safe-driver-prediction/discussion/44629#250927>`_. The ``name`` field in the feature transformation dictionary is ``rank_gauss``. The dict can contains two optional fields, i.e., ``epsilon`` which is used to avoid ``INF`` float during computation and ``uniquify`` which controls whether deduplicating input features before computing rank gauss norm.
 
   Example:
--- a/docs/source/cli/graph-construction/single-machine-gconstruct.rst
+++ b/docs/source/cli/graph-construction/single-machine-gconstruct.rst
@@ -147,7 +147,7 @@ GraphStorm provides a set of transformation operations for different types of fe
                   "max_bound": 2.,
                   "min_bound": -2.}
 
-* **Numerical standard transformation** normalizes numerical input features with `val = val / sum`, where `val` is the feature value and `sum` is a summation of all the values in the feature. The ``name`` field in the feature transformation dictionary is ``standard``. The dictionary can contain one optional field, i.e., ``sum``, which is summation of all the values in the feature col from the previous transformation.
+* **Numerical standard transformation** normalizes numerical input features with `val = val / sum`, where `val` is the feature value and `sum` is a summation of all the values in the feature. The ``name`` field in the feature transformation dictionary is ``standard``. The dictionary can contain one optional field, i.e., ``sum``, which is summation of all the values in the feature col from the previous transformation. (By default, the ``gconstruct.construct_graph`` command will save the ``sum`` value for each standard transformation.)
 
   Example:
 
--- a/python/graphstorm/gconstruct/transform.py
+++ b/python/graphstorm/gconstruct/transform.py
@@ -23,6 +23,7 @@ import sys
 import abc
 import json
 import warnings
+from typing import Any, Optional
 
 import numpy as np
 import torch as th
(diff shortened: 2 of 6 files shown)


python/graphstorm/gconstruct/transform.py
```diff
@@ -665,6 +666,128 @@ def call(self, feats):
...
+        The feature name used in the constructed graph.
+    sum: np.array or list of floats
+        The summation of all the values in the feature col
+        from the last transformation.
+        If it is None, we will compute it on the fly.
+        Default: None
+    out_dtype:
+        The dtype of the transformed feature.
+        Default: None, we will not do data type casting.
+        Default: None
```
jalencato, line 688: Shall we specify all valid output type here?

tests/unit-tests/gconstruct/test_transform.py
```diff
@@ -1147,7 +1148,122 @@ def test_hard_edge_dst_negative_transform(id_dtype):
     ground_truth[1][-2] = -1
     assert_equal(neg["hard_neg"][:,:10], ground_truth)
 
+@pytest.mark.parametrize("input_dtype", [np.cfloat, np.float32, np.float16])
```
jalencato, line 1151: Do we have a test case for summation == 0?

docs/source/cli/graph-construction/single-machine-gconstruct.rst
```diff
@@ -147,6 +147,15 @@ GraphStorm provides a set of transformation operations for different types of fe
                   "max_bound": 2.,
                   "min_bound": -2.}
 
+* **Numerical standard transformation** normalizes numerical input features with `val = val / sum`, where `val` is the feature value and `sum` is a summation of all the values in the feature. The ``name`` field in the feature transformation dictionary is ``standard``. The dictionary can contain one optional field, i.e., ``sum``, which is summation of all the values in the feature col from the previous transformation.
```
jalencato, line 150: Why is the sum optional here? Why not always determined by the summation value of the data input.

python/graphstorm/gconstruct/transform.py
```diff
@@ -665,6 +666,128 @@ def call(self, feats):
...
+            # It will load all data into main memory.
+            feats = feats.to_numpy()
+
+        if feats.dtype not in [np.float64, np.float32, np.float16, np.int64, \
+                              np.int32, np.int16, np.int8]:
+            try:
+                # if input dtype is not float or integer, we need to cast the data
+                # into float32
+                feats = feats.astype(np.float32)
+            except: # pylint: disable=bare-except
```
jalencato, line 784: I think ``except TypeError:`` will fix the pylint error here without disabling.

jalencato (commented):
Let us also fix the lint error proposed by the CI.

tests/unit-tests/gconstruct/test_transform.py
classicsong: These is one in L1221

docs/source/cli/graph-construction/single-machine-gconstruct.rst
classicsong: The distribution will change between training graph and inference graph if we do not save the summation value.

python/graphstorm/gconstruct/transform.py
classicsong: Thanks

python/graphstorm/gconstruct/transform.py
classicsong: No. Actually, we do not check the output type.
Maybe add the check in another PR.

thvasilo (changes requested):

docs/source/cli/graph-construction/single-machine-gconstruct.rst
classicsong: We do save the original value.
The implementation is compatible with GSProcessing.

classicsong: Updated the doc.

python/graphstorm/gconstruct/transform.py
```diff
@@ -665,6 +666,129 @@ def call(self, feats):
 
         return {self.feat_name: feats}
 
+class NumericalStandardTransform(TwoPhaseFeatTransform):
+    r""" Numerical value with standard normalization.
```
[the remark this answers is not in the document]
classicsong: I follow the naming of GSProcessing.
Let's keep the naming same first.

BTW, do you have a good naming for this transformation?

python/graphstorm/gconstruct/transform.py
```diff
@@ -665,6 +666,129 @@ def call(self, feats):
...
+    def __init__(self,
+                 col_name : str,
+                 feat_name : str,
+                 summation : Optional[Any] = None,
+                 out_dtype: Optional[str] = None,
+                 transform_conf: Optional[dict] = None):
+        self._summation = np.array(summation, dtype=np.float32) if summation is not None else None
+        self._conf = transform_conf
+
+        out_dtype = np.float32 if out_dtype is None else out_dtype
```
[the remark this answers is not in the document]
classicsong: ``self.out_dtype = ...`` is called in the parent class of NumericalStandardTransform.
I have updated the impl to set the default value of out_dtype as np.float32. (As it is a numerical feature transform)

python/graphstorm/gconstruct/transform.py
```diff
@@ -665,6 +666,129 @@ def call(self, feats):
...
+        # If sum has already been set.
+        # Skip the pre-process step
+        if self._summation is not None:
+            return {self.feat_name: None}
+
+        if isinstance(feats, ExtMemArrayWrapper):
+            # TODO(xiangsx): This is not memory efficient.
+            # It will load all data into main memory.
+            feats = feats.to_numpy()
+
```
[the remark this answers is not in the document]
classicsong: You mean after L718?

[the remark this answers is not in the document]
classicsong: I don't think we need to add an assert here.
We already have the ``assert isinstance(feats, (np.ndarray, ExtMemArrayWrapper))`` and handle the ``isinstance(feats, ExtMemArrayWrapper)`` case. There is no need to do assert again.

python/graphstorm/gconstruct/transform.py
```diff
@@ -665,6 +666,129 @@ def call(self, feats):
...
+        if feats.dtype not in [np.float64, np.float32, np.float16, np.int64, \
+                              np.int32, np.int16, np.int8]:
```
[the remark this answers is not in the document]
classicsong: I would perfer to change ``np.float64, np.float32, np.float16, np.int64, np.int32, np.int16, np.int8`` into a constant in another PR.

The ``\`` is necessary otherwise, the length of the code will exceed 100.

python/graphstorm/gconstruct/transform.py
```diff
@@ -665,6 +666,129 @@ def call(self, feats):
...
+            return
+
+        # We have to aggregative sum value from workers.
+        summations = []
+        for sum_i in info:
+            summations.append(sum_i)
+        summations = np.stack(summations)
+        summation = np.sum(summations, axis=0)
+
+        assert not np.any(summation == 0), \
```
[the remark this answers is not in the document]
classicsong: But mathematically, it is invalid. We should raise an exception.
And user should not use standard normalization in such cases.

python/graphstorm/gconstruct/transform.py
```diff
@@ -665,6 +666,129 @@ def call(self, feats):
...
+        assert isinstance(feats, (np.ndarray, ExtMemArrayWrapper)), \
+            f"Feature {self._feat_name} of NumericalMinMaxTransform " \
+            "must be numpy array or ExtMemArray"
+
+        if isinstance(feats, ExtMemArrayWrapper):
+            # TODO(xiangsx): This is not memory efficient.
+            # It will load all data into main memory.
+            feats = feats.to_numpy()
+
+        if feats.dtype not in [np.float64, np.float32, np.float16, np.int64, \
```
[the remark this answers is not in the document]
classicsong: There is a assert in L768

python/graphstorm/gconstruct/transform.py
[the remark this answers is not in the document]
thvasilo: No better naming options, I haven't really encountered this transformation outside of NeptuneML. We can discuss whether we want to change this implementation to the popular version of the standard transform.

thvasilo (commented):
LGTM, but let's have a discussion whether we want to change the transform to the more widely used standardization (zero mean, unit variance)

thvasilo (approved):
LGTM, but let's have an offline discussion whether we want to change the transform to the more widely used standardization (zero mean, unit variance)

jalencato (approved):
LGTM

classicsong:
> LGTM, but let's have an offline discussion whether we want to change the transform to the more widely used standardization (zero mean, unit variance)

Sounds good

PR #1102, opened 2024-11-28
Add support for ParMETIS to local Dockerfile

### LINKED ISSUE #873: [Refactor] Merge the parmetis dockerfile into the local graphstorm dockerfile

thvasilo:
*Issue #, if available:*

*Description of changes:*

Fixes #873 

* We add support for ParMETIS to the main local Dockerfile. This way we don't have to maintain two separate Dockerfiles.
* Add an optional positional parameter to determine whether to include ParMETIS dependencies when building local image


docker/local/Dockerfile.local
```diff
@@ -75,18 +81,50 @@ RUN TORCH_MAJOR_MINOR=$(echo $TORCH_VERSION | cut -c1-3) && \
...
+# Install Parmetis
+RUN cd /root && \
+    git clone --single-branch --branch main https://github.com/KarypisLab/PM4GNN.git && \
+    cd PM4GNN && \
+    make config cc=mpicc prefix=/root/local && \
+    make install
 
+ENV PATH=$PATH:/root/local/bin
+ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/root/local/lib/
+RUN cp /root/local/bin/pm_dglpart /root/local/bin/pm_dglpart3
```
classicsong, line 121: Why do this?

docker/local/Dockerfile.local
```diff
@@ -101,11 +139,21 @@ RUN mkdir -p ${SSHDIR} \
 
 EXPOSE ${SSH_PORT}
 
+ENV PYTHONPATH="/root/dgl/tools/:${PYTHONPATH}"
```
classicsong, line 142: Can we move the DGL ENV setting to L56 (Together with git clone dgl)?

docker/local/Dockerfile.local
thvasilo: I think this is needed for the parmetis executable and portable library to be available to bash and any running process. From https://github.com/KarypisLab/ParMETIS?tab=readme-ov-file#building-and-installing-parmetis

docker/local/Dockerfile.local
```diff
@@ -1,4 +1,5 @@
 ARG DEVICE=gpu
```
jalencato, line 1: Have you got any number about time to build the docker image here? I think a big burden here is that the time for building parmetis docker image will be too long, and also make the docker image itself too large.

docker/local/Dockerfile.local
```diff
@@ -101,11 +140,18 @@ RUN mkdir -p ${SSHDIR} \
...
+ENV PYTHONPATH="/graphstorm/python/:${PYTHONPATH}"
+
+
 # Copy GraphStorm scripts and tools
 COPY code/examples /graphstorm/examples
 COPY code/inference_scripts /graphstorm/inference_scripts
 COPY code/tools /graphstorm/tools
 COPY code/training_scripts /graphstorm/training_scripts
 COPY code/fetch_and_run.sh /graphstorm/fetch_and_run.sh
+RUN chmod +x "/graphstorm/fetch_and_run.sh"
```
jalencato, line 155: Any reason we need to do this now?

docker/build_docker_oss4local.sh
```diff
@@ -33,6 +33,13 @@ else
     DEVICE_TYPE="$4"
 fi
 
+# process argument 5: support for parmetis
+if [ -z "$4" ]; then
+    USE_PARMETIS="false"
```
jalencato, line 38: We better to note the option here in the doc. We have changed the way to build docker image here.

docker/local/Dockerfile.local
thvasilo: ParMETIS dependencies are or added if requested by the user, the build time should be roughly the same as building the ParMETIS image using the specialized Dockerfile. 

Could you clarify if there's a concern I'm missing here? I will add some measurements of build times for the images using the specialized Dockerfile  vs. this one.

thvasilo:
Removed 0.4 tag as we can merge this after the release, not urgent.

docker/README.md
```diff
@@ -7,74 +7,5 @@ recommend that our users use Docker as the base running environment to use Graph
...
-        --exclude "*__pycache__" --exclude "third_party"
-```
-
-and in `local/Dockerfile.local` replace the line `RUN cd /root; git clone --branch v${DGL_VERSION} https://github.com/dmlc/dgl.git`
-with the following lines:
-
-```Dockerfile
-COPY code/dgl /root/dgl
-ENV PYTHONPATH="/root/dgl/python/:${PYTHONPATH}"
-ENV LD_LIBRARY_PATH="/opt/gs-venv/lib/python3.9/site-packages/dgl/:$LD_LIBRARY_PATH"
```
jalencato, line 79: Is the readme hosted on anywhere on readthedoc?

docker/build_graphstorm_image.sh
```diff
@@ -78,6 +79,9 @@ parse_params() {
             SUFFIX="${2-}"
             shift
             ;;
+        -u | --use-parmetis)
```
jalencato, line 82: Maybe -up? u if just for use.

docker/local/Dockerfile.local
```diff
@@ -46,9 +47,17 @@ RUN pip install \
 ARG DGL_VERSION=2.3.0
 ARG DGL_CUDA_VERSION=121
 ARG OGB_VERSION=1.3.6
-ARG TORCH_VERSION=2.3
+ARG TORCH_VERSION=2.3.1
```
jalencato, line 50: I think we used 2.3.0

docs/source/install/env-setup.rst
```diff
@@ -171,6 +171,7 @@ tag and other aspects of the build. We list the full argument list below:
 * ``-x, --verbose``       Print script debug info (set -x)
 * ``-e, --environment``   Image execution environment. Must be one of 'local' or 'sagemaker'. Required.
 * ``-d, --device``        Device type, must be one of 'cpu' or 'gpu'. Default is 'gpu'.
+* ``-u, --use-parmetis``  When this flag is set we add the ParMETIS dependencies to the local image. ParMETIS partitioning is not available on SageMaker.
```
jalencato, line 174: Same here.

docker/local/Dockerfile.local
thvasilo: You mean in the source images? Yeah we can leave it as is for now and update the source images along with this parameter in another PR.

docker/build_graphstorm_image.sh
thvasilo: Hmm, maybe I'll leave this as only the fully spelled out `--use-parmetis` to avoid having to remember too many single-letter flags.

jalencato (approved):
LGTM

PR #1108, opened 2024-12-07
[SageMaker] Add GraphStorm SageMaker Pipeline creation and execution

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Add support for customizable SageMaker pipelines.
* We use one file to do the parameter parsing and validation, one to create the pipeline and one to execute.
* I used GenAI to create the first README, then hand-edited. 

### DIFF
--- a/python/graphstorm/gconstruct/file_io.py
+++ b/python/graphstorm/gconstruct/file_io.py
@@ -88,6 +88,8 @@ def expand_wildcard(data_files: List[str]) -> List[str]:
 
     """
     expanded_files = []
+    if len(data_files) == 1 and os.path.isdir(data_files[0]):
+        data_files = [os.path.join(data_files[0], "*")]
     for item in data_files:
         if '*' in item:
             matched_files = sorted(glob.glob(item))
--- a/python/graphstorm/gconstruct/file_io.py
+++ b/python/graphstorm/gconstruct/file_io.py
@@ -88,8 +88,6 @@ def expand_wildcard(data_files: List[str]) -> List[str]:
 
     """
     expanded_files = []
-    if len(data_files) == 1 and os.path.isdir(data_files[0]):
-        data_files = [os.path.join(data_files[0], "*")]
     for item in data_files:
         if '*' in item:
             matched_files = sorted(glob.glob(item))
--- a/sagemaker/pipeline/README.md
+++ b/sagemaker/pipeline/README.md
@@ -171,7 +171,8 @@ The pipeline's behavior is controlled by various configuration parameters, inclu
 - `--output-prefix-s3`: S3 prefix for the output data. (Required)
 - `--pipeline-name`: Name for the pipeline.
 - `--base-job-name`: Base job name for SageMaker jobs. (Default: 'gs')
-- `--jobs-to-run`: Space-separated string of jobs to run in the pipeline. (Required)
+- `--jobs-to-run`: Space-separated string of jobs to run in the pipeline.
+  Possible values are: "gconstruct", "gsprocessing", "dist_part", "gb_convert", "train", "inference" (Required).
 - `--log-level`: Logging level for the jobs. (Default: INFO)
 - `--step-cache-expiration`: Expiration time for the step cache. (Default: 30d)
 - `--update-pipeline`: Update an existing pipeline instead of creating a new one. (Flag)
--- a/sagemaker/pipeline/README.md
+++ b/sagemaker/pipeline/README.md
@@ -1,0 +1,195 @@
+# GraphStorm SageMaker Pipeline
+
+This project provides a set of tools to create and execute SageMaker pipelines for GraphStorm, a library for large-scale graph neural networks. The pipeline automates the process of graph construction, partitioning, training, and inference using Amazon SageMaker.
+
+## Table of Contents
+
+1. [Overview](#overview)
+2. [Prerequisites](#prerequisites)
+3. [Project Structure](#project-structure)
+4. [Installation](#installation)
+5. [Usage](#usage)
+   - [Creating a Pipeline](#creating-a-pipeline)
+   - [Executing a Pipeline](#executing-a-pipeline)
+6. [Pipeline Components](#pipeline-components)
+7. [Configuration](#configuration)
+8. [Advanced Usage](#advanced-usage)
+9. [Troubleshooting](#troubleshooting)
+
+## Overview
+
+This project simplifies the process of running GraphStorm workflows on Amazon SageMaker. It provides scripts to:
+
+1. Define and create SageMaker pipelines for GraphStorm tasks
+2. Execute these pipelines with customizable parameters
+3. Manage different stages of graph processing, including construction, partitioning, training, and inference
+
+## Prerequisites
+
+- Python 3.8+
+- AWS account with appropriate permissions. See the official
+  [SageMaker AI](https://docs.aws.amazon.com/sagemaker/latest/dg/build-and-manage-access.html) docs
+  for detailed permissions needed to create and run SageMaker Pipelines.
+- Familiarity with SageMaker AI and
+  [SageMaker Pipelines](https://docs.aws.amazon.com/sagemaker/latest/dg/pipelines.html).
+- Basic understanding of graph neural networks and GraphStorm
+
+## Project Structure
+
+The project consists of three main Python scripts:
+
+1. `create_sm_pipeline.py`: Defines the structure of the SageMaker pipeline
+2. `pipeline_parameters.py`: Manages the configuration and parameters for the pipeline
+3. `execute_pipeline.py`: Executes created pipelines
+
+## Access code and install dependencies
+
+1. Clone the GraphStorm repository:
+   ```
+   git clone https://github.com/awslabs/graphstorm.git
+   cd graphstorm/sagemaker/pipeline
+   ```
+
+2. Install the required dependencies:
+   ```
+   pip install sagemaker boto3
+   ```
+
+## Usage
+
+### Creating a Pipeline
+
+To create a new SageMaker pipeline for GraphStorm:
+
+```bash
+python create_sm_pipeline.py \
+    --role arn:aws:iam::123456789012:role/SageMakerRole \
+    --region us-west-2 \
+    --graphstorm-pytorch-image-url 123456789012.dkr.ecr.us-west-2.amazonaws.com/graphstorm:sm-cpu \
+    --instance-count 2 \
+    --jobs-to-run gconstruct train inference \
+    --graph-name my-graph \
+    --graph-construction-config-filename my_gconstruct_config.json \
+    --input-data-s3 s3://input-bucket/data \
+    --output-prefix-s3 s3://output-bucket/results \
+    --train-inference-task node_classification \
+    --train-yaml-s3 s3://config-bucket/train.yaml
+```
+
+This command creates a new pipeline with the specified configuration. The pipeline will
+include one GConstruct job, one training job and one inference job.
+It will use the configuration defined in `s3://input-bucket/data/my_gconstruct_config.json`
+to construct the graph and the train config file at `s3://config-bucket/train.yaml`
+to run training and inference.
+
+The `--instance-count` parameter determines the number of workers and partitions we will create and use
+during partitioning/training. It is also aliased to `--num-parts`.
+
+You can customize various aspects of the pipeline using additional command-line arguments. Refer to the script's help message for a full list of options:
+
+```bash
+python create_sm_pipeline.py --help
+```
+
+### Executing a Pipeline
+
+To execute a created pipeline:
+
+```bash
+python execute_pipeline.py \
+    --pipeline-name my-graphstorm-pipeline \
+    --region us-west-2
+```
+
+You can override various pipeline parameters during execution:
+
+```bash
+python execute_pipeline.py \
+    --pipeline-name my-graphstorm-pipeline \
+    --region us-west-2 \
+    --instance-count 4 \
+    --gpu-instance-type ml.g4dn.12xlarge
+```
+
+For a full list of execution options:
+
+```bash
+python execute_pipeline.py --help
+```
+
+For more fine-grained execution options see the
+[SageMaker AI documentation](https://docs.aws.amazon.com/sagemaker/latest/dg/pipelines-selective-ex.html).
+
+## Pipeline Components
+
+The GraphStorm SageMaker pipeline typically includes the following steps:
+
+1. **Graph Construction**: Builds the graph from input data
+2. **Graph Partitioning**: Partitions the graph for distributed processing
+3. **Training**: Trains the graph neural network model
+4. **Inference**: Runs inference on the trained model
+
+Each step is configurable and can be customized based on your specific requirements.
+
+## Configuration
+
+The pipeline's behavior is controlled by various configuration parameters, including:
+
+- AWS configuration (region, roles, image URLs)
+- Instance configuration (instance types, counts)
+- Task configuration (graph name, input/output locations)
+- Training and inference configurations
+
+Refer to the `PipelineArgs` class in `pipeline_parameters.py` for a complete list of configurable options.
+
+## Advanced Usage
+
+### Using GraphBolt
+
+To use GraphBolt for improved performance:
+
+```bash
+python create_sm_pipeline.py \
+    ... \
+    --use-graphbolt true
+```
+
+### Custom Job Sequences
+
+You can customize the sequence of jobs in the pipeline using the `--jobs-to-run` argument when creating the pipeline. For example:
+
+```bash
+python create_sm_pipeline.py \
+    ... \
+    --jobs-to-run gsprocessing dist_part gb_convert train inference \
+    --use-graphbolt true
+```
+
+will create a pipeline that uses GSProcessing to process and prepare the data for partitioning,
+use GSPartition to partition the data, convert the partitioned data to the GraphBolt format,
+then run a train and an inference job in sequence.
+
+
+### Asynchronous Execution
+
+To start a pipeline execution without waiting for it to complete:
+
+```bash
+python execute_pipeline.py \
+    --pipeline-name my-graphstorm-pipeline \
+    --region us-west-2 \
+    --async-execution
+```
+
+## Troubleshooting
+
+- Ensure all required AWS permissions are correctly set up
+- Check SageMaker execution logs for detailed error messages
+- Verify that all S3 paths are correct and accessible. Note trailing `/` that could cause issues.
+- Ensure that the specified EC2 instance types are available in your region
+
+See also [Troubleshooting Amazon SageMaker Pipelines](https://docs.aws.amazon.com/sagemaker/latest/dg/pipelines-troubleshooting.html)
+
+For more detailed information about GraphStorm, refer to the [GraphStorm documentation](https://graphstorm.readthedocs.io/).
+
+If you encounter any issues or have questions, please open an issue in the project's [GitHub repository](https://github.com/awslabs/graphstorm/issues).
--- a/sagemaker/pipeline/README.md
+++ b/sagemaker/pipeline/README.md
@@ -40,9 +40,12 @@ The project consists of three main Python scripts:
 
 1. `create_sm_pipeline.py`: Defines the structure of the SageMaker pipeline
 2. `pipeline_parameters.py`: Manages the configuration and parameters for the pipeline
-3. `execute_pipeline.py`: Executes created pipelines
+3. `execute_sm_pipeline.py`: Executes created pipelines
 
-## Access code and install dependencies
+## Installation
+
+To construct and execute GraphStorm SageMaker pipelines you need the code
+available and a Python environment with the SageMaker SDK and `boto3` installed.
 
 1. Clone the GraphStorm repository:
    ```
@@ -63,15 +66,15 @@ To create a new SageMaker pipeline for GraphStorm:
 
 ```bash
 python create_sm_pipeline.py \
-    --role arn:aws:iam::123456789012:role/SageMakerRole \
+    --execution-role arn:aws:iam::123456789012:role/SageMakerRole \
     --region us-west-2 \
-    --graphstorm-pytorch-image-url 123456789012.dkr.ecr.us-west-2.amazonaws.com/graphstorm:sm-cpu \
+    --graphstorm-pytorch-cpu-image-url 123456789012.dkr.ecr.us-west-2.amazonaws.com/graphstorm:sagemaker-cpu \
     --instance-count 2 \
     --jobs-to-run gconstruct train inference \
     --graph-name my-graph \
     --graph-construction-config-filename my_gconstruct_config.json \
     --input-data-s3 s3://input-bucket/data \
-    --output-prefix-s3 s3://output-bucket/results \
+    --output-prefix s3://output-bucket/results \
     --train-inference-task node_classification \
     --train-yaml-s3 s3://config-bucket/train.yaml
 ```
@@ -83,7 +86,7 @@ to construct the graph and the train config file at `s3://config-bucket/train.ya
 to run training and inference.
 
 The `--instance-count` parameter determines the number of workers and partitions we will create and use
-during partitioning/training. It is also aliased to `--num-parts`.
+during partitioning/training.
 
 You can customize various aspects of the pipeline using additional command-line arguments. Refer to the script's help message for a full list of options:
 
@@ -96,7 +99,7 @@ python create_sm_pipeline.py --help
 To execute a created pipeline:
 
 ```bash
-python execute_pipeline.py \
+python execute_sm_pipeline.py \
     --pipeline-name my-graphstorm-pipeline \
     --region us-west-2
 ```
@@ -104,7 +107,7 @@ python execute_pipeline.py \
 You can override various pipeline parameters during execution:
 
 ```bash
-python execute_pipeline.py \
+python execute_sm_pipeline.py \
     --pipeline-name my-graphstorm-pipeline \
     --region us-west-2 \
     --instance-count 4 \
@@ -114,20 +117,22 @@ python execute_pipeline.py \
 For a full list of execution options:
 
 ```bash
-python execute_pipeline.py --help
+python execute_sm_pipeline.py --help
 ```
 
-For more fine-grained execution options see the
+For more fine-grained execution options, like selective execution,
 [SageMaker AI documentation](https://docs.aws.amazon.com/sagemaker/latest/dg/pipelines-selective-ex.html).
 
 ## Pipeline Components
 
-The GraphStorm SageMaker pipeline typically includes the following steps:
+The GraphStorm SageMaker pipeline can include the following steps:
 
-1. **Graph Construction**: Builds the graph from input data
-2. **Graph Partitioning**: Partitions the graph for distributed processing
-3. **Training**: Trains the graph neural network model
-4. **Inference**: Runs inference on the trained model
+1. **Graph Construction (GConstruct)**: Builds the partitioned graph from input data in a single instance.
+2. **Graph Processing (GSProcessing)**: Processes the graph data using PySpark, preparing it for distributed partitioning.
+3. **Graph Partitioning (DistPart)**: Partitions the graph using multiple instances.
+4. **GraphBolt Conversion**: Converts the partitioned data (usually generated from DistPart) to GraphBolt format.
+5. **Training**: Trains the graph neural network model.
+6. **Inference**: Runs inference on the trained model.
 
 Each step is configurable and can be customized based on your specific requirements.
 
@@ -165,27 +170,40 @@ python create_sm_pipeline.py \
     --use-graphbolt true
 ```
 
-will create a pipeline that uses GSProcessing to process and prepare the data for partitioning,
-use GSPartition to partition the data, convert the partitioned data to the GraphBolt format,
+This will create a pipeline that uses GSProcessing to process and prepare the data for partitioning,
+use DistPart to partition the data, convert the partitioned data to the GraphBolt format,
 then run a train and an inference job in sequence.
-
+You can use this job sequence when your graph is too large to partition on one instance using
+GConstruct (1+ TB is the suggested threshold to move to distributed partitioning).
 
 ### Asynchronous Execution
 
 To start a pipeline execution without waiting for it to complete:
 
 ```bash
-python execute_pipeline.py \
+python execute_sm_pipeline.py \
     --pipeline-name my-graphstorm-pipeline \
     --region us-west-2 \
     --async-execution
 ```
 
+### Local Execution
+
+For testing purposes, you can execute the pipeline locally:
+
+```bash
+python execute_sm_pipeline.py \
+    --pipeline-name my-graphstorm-pipeline \
+    --local-execution
+```
+
+Note that local execution requires a GPU if the pipeline is configured to use GPU instances.
+
 ## Troubleshooting
 
 - Ensure all required AWS permissions are correctly set up
 - Check SageMaker execution logs for detailed error messages
-- Verify that all S3 paths are correct and accessible. Note trailing `/` that could cause issues.
+- Verify that all S3 paths are correct and accessible
 - Ensure that the specified EC2 instance types are available in your region
 
 See also [Troubleshooting Amazon SageMaker Pipelines](https://docs.aws.amazon.com/sagemaker/latest/dg/pipelines-troubleshooting.html)
--- a/sagemaker/pipeline/README.md
+++ b/sagemaker/pipeline/README.md
@@ -32,7 +32,7 @@ This project simplifies the process of running GraphStorm workflows on Amazon Sa
   for detailed permissions needed to create and run SageMaker Pipelines.
 - Familiarity with SageMaker AI and
   [SageMaker Pipelines](https://docs.aws.amazon.com/sagemaker/latest/dg/pipelines.html).
-- Basic understanding of graph neural networks and GraphStorm
+- Basic understanding of graph neural networks and [GraphStorm](https://graphstorm.readthedocs.io/en/latest/index.html).
 
 ## Project Structure
 
@@ -124,7 +124,7 @@ For a full list of execution options:
 python execute_sm_pipeline.py --help
 ```
 
-For more fine-grained execution options, like selective execution, please refer to 
+For more fine-grained execution options, like selective execution, please refer to
 [SageMaker AI documentation](https://docs.aws.amazon.com/sagemaker/latest/dg/pipelines-selective-ex.html).
 
 ## Pipeline Components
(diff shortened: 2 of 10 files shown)


python/graphstorm/gconstruct/file_io.py
```diff
@@ -88,6 +88,8 @@ def expand_wildcard(data_files: List[str]) -> List[str]:
 
     """
     expanded_files = []
+    if len(data_files) == 1 and os.path.isdir(data_files[0]):
+        data_files = [os.path.join(data_files[0], "*")]
```
thvasilo, line 92: This adds support for path prefixes to gconstruct (`path/to/files/`), allowing the exact same config to be used for both gconstruct and gsporcessing

classicsong: I would suggest we create a standalone PR for this fix and add some unit tests.

sagemaker/pipeline/README.md
```diff
@@ -0,0 +1,213 @@
...
+    --graph-name my-graph \
+    --graph-construction-config-filename my_gconstruct_config.json \
+    --input-data-s3 s3://input-bucket/data \
+    --output-prefix s3://output-bucket/results \
+    --train-inference-task node_classification \
+    --train-yaml-s3 s3://config-bucket/train.yaml
+```
+
+This command creates a new pipeline with the specified configuration. The pipeline will
+include one GConstruct job, one training job and one inference job.
```
classicsong, line 83: You also need to introduce the `execution-role` and `graphstorm-pytorch-cpu-image-url` which are very important.

sagemaker/pipeline/README.md
```diff
@@ -0,0 +1,213 @@
...
+```bash
+python create_sm_pipeline.py --help
+```
+
+### Executing a Pipeline
+
+To execute a created pipeline:
+
+```bash
+python execute_sm_pipeline.py \
```
classicsong, line 102: Have you defined the pipeline-name in L68?

sagemaker/pipeline/README.md
```diff
@@ -0,0 +1,213 @@
...
+    --pipeline-name my-graphstorm-pipeline \
+    --region us-west-2
+```
+
+You can override various pipeline parameters during execution:
```
classicsong, line 107:
```suggestion
You can override the default pipeline parameters during execution:
```

sagemaker/pipeline/README.md
```diff
@@ -0,0 +1,213 @@
...
+    --gpu-instance-type ml.g4dn.12xlarge
+```
+
+For a full list of execution options:
+
+```bash
+python execute_sm_pipeline.py --help
+```
+
+For more fine-grained execution options, like selective execution,
```
classicsong, line 123:
```suggestion
For more fine-grained execution options, like selective execution, please refer to 
```

sagemaker/pipeline/README.md
```diff
@@ -0,0 +1,213 @@
...
+[SageMaker AI documentation](https://docs.aws.amazon.com/sagemaker/latest/dg/pipelines-selective-ex.html).
+
+## Pipeline Components
+
+The GraphStorm SageMaker pipeline can include the following steps:
+
+1. **Graph Construction (GConstruct)**: Builds the partitioned graph from input data in a single instance.
+2. **Graph Processing (GSProcessing)**: Processes the graph data using PySpark, preparing it for distributed partitioning.
```
classicsong, line 131:
```suggestion
2. **Graph Processing (GSProcessing)**: Processes the graph data using PySpark, preparing it for distributed graph partitioning.
```

sagemaker/pipeline/README.md
```diff
@@ -0,0 +1,213 @@
...
+## Configuration
+
+The pipeline's behavior is controlled by various configuration parameters, including:
+
+- AWS configuration (region, roles, image URLs)
+- Instance configuration (instance types, counts)
+- Task configuration (graph name, input/output locations)
+- Training and inference configurations
+
+Refer to the `PipelineArgs` class in `pipeline_parameters.py` for a complete list of configurable options.
```
classicsong, line 148: We need to list the options here when we officially announce the support of SageMaker pipline.

sagemaker/pipeline/README.md
```diff
@@ -0,0 +1,213 @@
...
+
+```bash
+python create_sm_pipeline.py \
+    ... \
+    --jobs-to-run gsprocessing dist_part gb_convert train inference \
+    --use-graphbolt true
+```
+
+This will create a pipeline that uses GSProcessing to process and prepare the data for partitioning,
+use DistPart to partition the data, convert the partitioned data to the GraphBolt format,
```
classicsong, line 174:
```suggestion
uses DistPart to partition the data, converts the partitioned data to the GraphBolt format,
```

sagemaker/pipeline/README.md
```diff
@@ -0,0 +1,213 @@
...
+then run a train and an inference job in sequence.
```
classicsong, line 175:
```suggestion
then runs a train and an inference job in sequence.
```

sagemaker/pipeline/README.md
```diff
@@ -0,0 +1,213 @@
...
+You can use this job sequence when your graph is too large to partition on one instance using
+GConstruct (1+ TB is the suggested threshold to move to distributed partitioning).
```
classicsong, line 177: 1TB+ of what? edges or nodes?

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+
+@dataclass
+class InstanceConfig:
+    """Configuration for SageMaker instances"""
+
+    train_infer_instance_count: int
+    cpu_instance_type: str
+    gpu_instance_type: str
+    graph_construction_instance_type: str
+    gsprocessing_instance_count: int
```
classicsong, line 59: Do we need the gsprocessing_instance_type?

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+
+    def __post_init__(self):
+        assert (
+            self.cpu_instance_type or self.gpu_instance_type
+        ), "At least one instance type (CPU or GPU) should be specified."
+
+        if self.train_on_cpu:
+            assert (
+                self.cpu_instance_type
+            ), "Need to provide a CPU instance type when training on CPU"
```
classicsong, line 71: Add an else to check the self.gpu_instance_type

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+
+        if not self.graph_construction_instance_type:
+            self.graph_construction_instance_type = self.cpu_instance_type
```
classicsong, line 74: Then we also need to check that self.cpu_instance_type is provided.

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+                "No graph processing instance type specified, using the CPU instance type: %s",
+                self.cpu_instance_type,
+            )
+
+
+@dataclass
+class TaskConfig:
+    """Pipeline/task-level configuration"""
+
+    base_job_name: str
```
classicsong, line 85: What does base_job_name mean?
Can you add a docstr for each field? (Same for other classes.)

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+        # Ensure jobs are in the right order
+        self.jobs_to_run.sort(key=lambda x: JOB_ORDER[x])
+
+        # We should only run either gconstruct or gsprocessing
+        if "gconstruct" in self.jobs_to_run and "gsprocessing" in self.jobs_to_run:
+            raise ValueError(
+                "Should not try to run both GConstruct and GSProcessing steps, "
+                f"got job sequence: {self.jobs_to_run}"
+            )
+        # TODO: Should we allow running GConstruct with dist_part?
```
classicsong, line 109: No

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+    input_json_filename: str
+    output_json_filename: str
+
+
+@dataclass
+class TrainingConfig:
+    """Configuration for the training step"""
+
+    model_output_path: str
+    train_inference_task: str
```
classicsong, line 146: Why we need train_inference_task here?

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+        self.use_graphbolt_str = self.use_graphbolt_str.lower()
+
+
+@dataclass
+class InferenceConfig:
+    """Configuration for the inference step"""
+
+    save_embeddings: bool
+    save_predictions: bool
+    inference_model_snapshot: str
```
classicsong, line 161: inference_model_snapshot path or url?

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+    instance_config: InstanceConfig
+    task_config: TaskConfig
+    partition_config: PartitionConfig
+    training_config: TrainingConfig
+    inference_config: InferenceConfig
+    script_paths: ScriptPaths
+    step_cache_expiration: str
+    update: bool
+
+    def __eq__(self, other):
```
classicsong, line 216: Why we want to overwrite ``__eq__``?

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+        if not isinstance(other, PipelineArgs):
+            return False
+        return asdict(self) == asdict(other)
+
+    def __hash__(self):
```
classicsong, line 221: Why we want to overwrite `__hash__`

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+        # these do not affect the outcome of the execution
+        obj_dict.pop("step_cache_expiration")
+        obj_dict.pop("update")
+
+        # Convert the dictionary to a JSON string and create an md5 hash
+        json_str = json.dumps(obj_dict, sort_keys=True)
+        return hashlib.md5(json_str.encode()).hexdigest()
+
+    def __post_init__(self):
+        if not self.instance_config.train_on_cpu:
```
classicsong, line 256: Why not move the check into instance_config's `__post_init__`?

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+            assert self.instance_config.gpu_instance_type, (
+                "GPU instance type must be specified if not training on CPU, "
+                f"got {self.instance_config.train_on_cpu=} "
+                f"{self.instance_config.gpu_instance_type=}"
+            )
+            assert self.aws_config.graphstorm_pytorch_cpu_image_url, (
```
classicsong, line 262: graphstorm_pytorch_gpu_image_url ?

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+                "Must use provide GPU image when training on GPU. "
+                "use --graphstorm-pytorch-gpu-image-url"
+            )
```
classicsong, line 265: What about graphstorm_pytorch_cpu_image_url?

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+        ):
+            assert self.graph_construction_config.config_filename, (
+                "Need to provide a GConstruct/GSProcessing config file "
+                "when running graph construction."
+            )
+
+        # TODO: When using GConstruct+DistPart/GBConvert (possible?) provide
+        # the correct partition input JSON filename?
+        # if "gconstruct" in self.task_config.jobs_to_run:
+        #     if "dist_part" in self.task_config.jobs_to_run or "gb_convert" in self.task_config.jobs_to_run:
```
classicsong, line 280: I think we do need this.

classicsong: If someone can not run gconstruct and partition with a single instance, he/she needs to use gsprocessing+distpart.

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+            )
+            logging.warning(
+                "No GSProcessing instance count specified, using the training instance count: %s",
+                self.instance_config.gsprocessing_instance_count,
+            )
+
+        # GConstruct uses 'metis', so just translate that if needed
+        if (
+            self.partition_config.partition_algorithm.lower() == "parmetis"
+            and "gconstruct" in self.task_config.jobs_to_run
```
classicsong, line 334: first check "gconstruct" in self.task_config.jobs_to_run and then the partition_algorithm.

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+        type=int,
+        help="Number of worker instances/partitions for partition, training, inference.",
+        required=True,
+    )
+    optional_args.add_argument(
+        "--cpu-instance-type",
+        type=str,
+        help=(
+            "CPU instance type. "
+            "Always used in DistPart step and if '--train-on-cpu' is provided, "
```
classicsong, line 462: What about gconstruct?

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+    optional_args.add_argument(
+        "--gsprocessing-instance-count",
+        type=int,
+        default=None,
+        help="Number of GSProcessing instances. Default is equal to number of training instances.",
+    )
+    optional_args.add_argument(
+        "--volume-size-gb",
+        type=int,
+        help="Additional volume size for instances in GB.",
```
classicsong, line 500:
```suggestion
        help="Additional volume size for SageMaker instances in GB.",
```

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+    required_args.add_argument(
+        "--output-prefix-s3",
+        type=str,
+        required=True,
+        help="S3 prefix for the output data. Required.",
+    )
+    optional_args.add_argument(
+        "--pipeline-name",
+        type=str,
+        help="Name for the pipeline. Needs to be unique per account and region.",
```
classicsong, line 523: What if it is not provided?

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+    )
+    optional_args.add_argument(
+        "--base-job-name",
+        type=str,
+        default="gs",
+        help="Base job name for SageMaker jobs. Default: 'sm'",
```
classicsong, line 529:
```suggestion
        help="Base job name for SageMaker jobs. Default: 'gs'",
```

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+    )
+    required_args.add_argument(
+        "--jobs-to-run",
+        nargs="+",
+        required=True,
+        help=(
+            "Space-separated string of jobs to run in the pipeline, "
+            "e.g. 'gconstruct train inference'. "
```
classicsong, line 537:
```suggestion
            "e.g. '--jobs-to-run gconstruct train inference'. "
```

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+        action="store_true",
+        help="Update an existing pipeline instead of creating a new one.",
+    )
+
+    # Graph construction configuration
+    optional_args.add_argument(
+        "--graph-construction-config-filename",
+        type=str,
+        default="",
+        help="Filename for the graph construction config. "
```
classicsong, line 565: S3 url?

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+    optional_args.add_argument(
+        "--inference-script",
+        type=str,
+        default=f"{SCRIPT_PATH}/../run/infer_entry.py",
+        help="Path to inference SageMaker entry point script.",
+    )
+    optional_args.add_argument(
+        "--gconstruct-script",
+        type=str,
+        default=f"{SCRIPT_PATH}/../run/gconstruct_entry.py",
```
classicsong, line 689: Where is the `../run/gconstruct_entry.py`?

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,764 @@
...
+    optional_args.add_argument(
+        "--dist-part-script",
+        type=str,
+        default=f"{SCRIPT_PATH}/../run/partition_entry.py",
+        help="Path to DistPartition SageMaker entry point script.",
+    )
+    optional_args.add_argument(
+        "--gb-convert-script",
+        type=str,
+        default=f"{SCRIPT_PATH}/../run/gb_convert_entry.py",
```
classicsong, line 671: Where is the `../run/gb_convert_entry.py`

sagemaker/pipeline/create_sm_pipeline.py
```diff
@@ -0,0 +1,693 @@
...
+            "InferenceConfigFile", inference_yaml_default
+        )
+        self.inference_model_snapshot_param = self._create_string_parameter(
+            "InferenceModelSnapshot", args.inference_config.inference_model_snapshot
+        )
+
+    def _create_pipeline_steps(
+        self, args: PipelineArgs
+    ) -> List[Union[ProcessingStep, TrainingStep]]:
+        """Creates the pipeline steps and reeturns them as a list
```
classicsong, line 238:
```suggestion
        """Creates the pipeline steps and returns them as a list
```

sagemaker/pipeline/create_sm_pipeline.py
```diff
@@ -0,0 +1,693 @@
...
+        ----------
+        args : PipelineArgs
+            Pipeline arguments
+
+        Returns
+        -------
+        List[Union[ProcessingStep, TrainingStep]]
+            A list of pipeline steps to be executed in sequence
+        """
+        pipeline_steps: List[Union[ProcessingStep, TrainingStep]] = []
```
classicsong, line 250: What about inference step?

sagemaker/pipeline/create_sm_pipeline.py
```diff
@@ -0,0 +1,693 @@
...
+                boto_session=boto_session, sagemaker_client=sagemaker_client
+            )
+
+    def create_pipeline(self) -> Pipeline:
+        """Create a SageMaker pipeline for GraphStorm.
+
+        The pipeline consists of the following steps:
+        1. Partition graph
+        2. Train graph
+        3. Encode node IDs
```
classicsong, line 114: Why there is a step of Encode node IDs?

sagemaker/pipeline/create_sm_pipeline.py
```diff
@@ -0,0 +1,693 @@
...
+            return PipelineSession(
+                boto_session=boto_session, sagemaker_client=sagemaker_client
+            )
+
+    def create_pipeline(self) -> Pipeline:
+        """Create a SageMaker pipeline for GraphStorm.
+
+        The pipeline consists of the following steps:
+        1. Partition graph
+        2. Train graph
```
classicsong, line 113: What does train graph mean?

sagemaker/pipeline/create_sm_pipeline.py
```diff
@@ -0,0 +1,693 @@
...
+            "--num-parts",
+            self.instance_count_param.to_string(),
+            "--part-method",
+            self.partition_algorithm_param,
+        ]
+
+        # TODO: Make this a pipeline parameter?
+        if args.graph_construction_config.graph_construction_args:
+            gconstruct_arguments.extend(
+                args.graph_construction_config.graph_construction_args.split(" ")
```
classicsong, line 339: Have you checked if there is nargs='+' in the gconstruct arg list?

sagemaker/pipeline/create_sm_pipeline.py
```diff
@@ -0,0 +1,693 @@
...
+        # GSProcessing should always be the first step and start with the source data
+        gsprocessing_config = Join(
+            on="/",
+            values=[
+                self.input_data_param,
+                self.graphconstruct_config_param,
+            ],
+        )
+
+        print(gsprocessing_config.expr)
```
classicsong, line 388: remove print or use logging

sagemaker/pipeline/create_sm_pipeline.py
```diff
@@ -0,0 +1,693 @@
...
+        gb_convert_step = ProcessingStep(
+            name="GraphBoltConversion",
+            code=args.script_paths.gb_convert_script,
+            processor=gb_part_processor,
+            inputs=[
+                ProcessingInput(
+                    input_name="dist_graph_s3_input",
+                    destination="/opt/ml/processing/dist_graph/",
+                    source=self.next_step_data_input,
+                    s3_input_mode="File",
```
classicsong, line 521: Are we loading S3 files by our own script?

sagemaker/pipeline/create_sm_pipeline.py
```diff
@@ -0,0 +1,693 @@
...
+        # Training step
+        train_estimator = PyTorch(
+            entry_point=os.path.basename(args.script_paths.train_script),
+            source_dir=os.path.dirname(args.script_paths.train_script),
+            image_uri=self.train_infer_image,
+            role=args.aws_config.execution_role,
+            instance_count=self.instance_count_param,
+            instance_type=self.train_infer_instance,
+            py_version="py3",
+            hyperparameters={
```
classicsong, line 550: use `train_params` dict as in `_create_inference_step`?

sagemaker/pipeline/create_sm_pipeline.py
```diff
@@ -0,0 +1,693 @@
...
+    )
+
+    pipeline_generator = GraphStormPipelineGenerator(pipeline_args)
+
+    pipeline = pipeline_generator.create_pipeline()
+
+    if pipeline_args.update:
+        # TODO: If updating ensure pipeline exists first to get more informative error
+        pipeline.update(role_arn=pipeline_args.aws_config.execution_role)
+        print(f"Pipeline '{pipeline.name}' updated successfully.")
```
classicsong, line 686: Ues logging.

sagemaker/pipeline/execute_sm_pipeline.py
```diff
@@ -0,0 +1,275 @@
...
+            "By default we look for '<pipeline-name>-pipeline-args.json' in the working dir."
+        ),
+    )
+
+    overrides = parser.add_argument_group(
+        "Pipeline overrides", "Override default pipeline parameters at execution time."
+    )
+
+    # Optional override parameters
+    overrides.add_argument("--instance-count", type=int, help="Override instance count")
```
classicsong, line 76: How to control the instance cnt for train/infer and gsprocessing separately?

sagemaker/pipeline/execute_sm_pipeline.py
```diff
@@ -0,0 +1,275 @@
...
+                raise RuntimeError(
+                    "Need host with NVidia GPU to run training on GPU! "
+                    "Try re-deploying the pipeline with --train-on-cpu set."
+                )
+        # Use local pipeline and session
+        local_session = LocalPipelineSession()
+        pipeline_generator = GraphStormPipelineGenerator(
+            pipeline_deploy_args, input_session=local_session
+        )
+        # Set shared memory to half the host's size, as SM does
```
classicsong, line 165:
```suggestion
        # Set shared memory to half of the host's size, as SM does
```

sagemaker/pipeline/execute_sm_pipeline.py
```diff
@@ -0,0 +1,275 @@
...
+    else:
+        assert args.region, "Need to provide --region for remote SageMaker execution"
+        boto_session = boto3.Session(region_name=args.region)
+        # Use remote pipeline and session
+        remote_session = PipelineSession(boto_session)
+        pipeline = Pipeline(name=args.pipeline_name, sagemaker_session=remote_session)
+
+    # Prepare parameter overrides
+    execution_params = {}
+    if args.instance_count is not None:
```
classicsong, line 182:
```suggestion
    if args.instance_count:
```
?

sagemaker/pipeline/execute_sm_pipeline.py
```diff
@@ -0,0 +1,275 @@
...
+
+        if pipeline_deploy_args.get_hash_hex() != deploy_time_hash:
+            new_prefix = os.path.join(
+                pipeline_deploy_args.task_config.output_prefix,
+                args.pipeline_name,
+                pipeline_deploy_args.get_hash_hex(),
+            )
+            warnings.warn(
+                "The pipeline execution arguments have been modified "
+                "compared to the deployment parameters. "
```
classicsong, line 252:
```suggestion
```

sagemaker/pipeline/execute_sm_pipeline.py
```diff
@@ -0,0 +1,275 @@
...
+
+    # If no parameters are provided, use an empty dict to use all defaults
+    execution = pipeline.start(
+        parameters=execution_params or {},
+    )
+
+    if args.local_execution:
+        sys.exit(0)
+
+    print(f"Pipeline execution started: {execution.describe()}")
```
classicsong, line 264: use logging

sagemaker/pipeline/create_sm_pipeline.py
```diff
@@ -0,0 +1,693 @@
...
+            "PartitionAlgorithm", args.partition_config.partition_algorithm
+        )
+        # TODO: Probably should not be a parameter
+        self.graph_name_param = self._create_string_parameter(
+            "GraphName", args.task_config.graph_name
+        )
+        self.num_trainers_param = self._create_int_parameter(
+            "NumTrainers", args.training_config.num_trainers
+        )
+        # TODO: Maybe should not be configurable because it requires changing job sequence?
```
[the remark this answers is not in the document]
thvasilo: I mean use-graphbolt should probably not be a runtime parameter for the pipeline, because specifically in the distributed processing job sequence we need to add another pipeline step, going from `GSProcessing->GSPartition->...` to `GSProcessing->GSPartition->GBConvert->...`.

For the GConstruct job sequence we can use `--use-graphbolt` as a runtime parameter, because the steps remain the same, we just change their parameters.

sagemaker/pipeline/create_sm_pipeline.py
```diff
@@ -0,0 +1,693 @@
...
+        # If inference yaml is not provided, re-use the training one
+        inference_yaml_default = (
+            args.inference_config.inference_yaml_file
+            or args.training_config.train_yaml_file
+        )
+        self.inference_config_file_param = self._create_string_parameter(
+            "InferenceConfigFile", inference_yaml_default
+        )
+        self.inference_model_snapshot_param = self._create_string_parameter(
+            "InferenceModelSnapshot", args.inference_config.inference_model_snapshot
```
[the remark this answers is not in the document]
thvasilo: It's the epoch we should use for inference.

We have an issue with inference requiring a specific epoch to load the model from. Because this depends on the training YAML we can't derive this during pipeline deployment, so we have to ask the user to provide it explicitly here.

sagemaker/pipeline/create_sm_pipeline.py
thvasilo: These are SageMaker Pipeline classes, see https://docs.aws.amazon.com/sagemaker/latest/dg/build-and-manage-steps.html#build-and-manage-properties.

For inference we still use a SageMaker Training job.

sagemaker/pipeline/create_sm_pipeline.py
thvasilo: Good point, I'll see if we have to adjust the way we parse these args, maybe use the method we use in https://github.com/awslabs/graphstorm/blob/391ff1636ebed735bed17ba57ee7d2f8b894f52b/sagemaker/launch/common_parser.py#L70-L91

sagemaker/pipeline/create_sm_pipeline.py
thvasilo: Not for GBConvert, we use the SageMaker service to download the graph input at job launch time.

sagemaker/pipeline/execute_sm_pipeline.py
thvasilo: We have the pipeline parameter `gsprocessing_instance_count` but we only expose it during pipeline deployment. I can make that into an execution-time parameter as well.

sagemaker/pipeline/execute_sm_pipeline.py
thvasilo: Shouldn't we leave this in as further explanation? To clarify this only happens when the parameters differ from the deployment params.

sagemaker/pipeline/pipeline_parameters.py
thvasilo: We use `graph_construction_instance_type` for that. Where possible we share `graph_construction*` params between GSProcessing and GConstruct since we only allow one of these two steps to exist in a pipeline.

sagemaker/pipeline/pipeline_parameters.py
thvasilo: It's the way we provide e.g. `node_classification` to the train/inference launch scripts, that requires the `--task-type` to be set.

sagemaker/pipeline/pipeline_parameters.py
thvasilo: The model prefix path is determined by the pipeline, and the user only chooses which epoch to use. 
I think `inference_model_snapshot` is descriptive enough for that? e.g. `epoch-15` is a snapshot, the user doesn't really provide the path/url to the model.

```
inference_model_path = Join(
            on="/",
            values=[
                self.model_input_path,
                self.inference_model_snapshot_param,
            ],
        )
```

sagemaker/pipeline/pipeline_parameters.py
thvasilo: Just to ensure we control the implementation. Creation of `__hash__` is a bit convoluted for data classes so I want to be explicit.

From https://docs.python.org/3/library/dataclasses.html

> If eq and frozen are both true, by default @dataclass will generate a __hash__() method for you. If eq is true and frozen is false, __hash__() will be set to None, marking it unhashable (which it is, since it is mutable). If eq is false, __hash__() will be left untouched meaning the __hash__() method of the superclass will be used (if the superclass is [object](https://docs.python.org/3/library/functions.html#object), this means it will fall back to id-based hashing).

sagemaker/pipeline/pipeline_parameters.py
thvasilo: Good point, `assert self.instance_config.gpu_instance_type` should be moved, `assert self.aws_config.graphstorm_pytorch_gpu_image_url` needs to be here to have access to the AWS config.

sagemaker/pipeline/pipeline_parameters.py
thvasilo: Yes in that case they'd need to re-deploy their pipeline to add the necessary steps. I'll come up with a test that works here.

sagemaker/pipeline/pipeline_parameters.py
thvasilo: That one uses `"--graph-construction-instance-type`

sagemaker/pipeline/pipeline_parameters.py
thvasilo: We use the function `_get_pipeline_name(args: PipelineArgs) -> str` which covers both cases.

sagemaker/pipeline/pipeline_parameters.py
thvasilo: It's only the filename here, we derive the full S3 path from the input prefix:

```
gc_proc_input = ProcessingInput(
            source=self.input_data_param, destination=gc_local_input_path
        )

[...]
gconstruct_arguments = [
            "--graph-config-path",
            self.graphconstruct_config_param,
            "--input-path",
            gc_local_input_path,
            ...
]
```

sagemaker/pipeline/pipeline_parameters.py
thvasilo: Was previously missing but added it's at https://github.com/awslabs/graphstorm/blob/main/sagemaker/run/gb_convert_entry.py

sagemaker/pipeline/pipeline_parameters.py
thvasilo: https://github.com/awslabs/graphstorm/blob/main/sagemaker/run/gconstruct_entry.py

Starting at `sagemaker/pipeline` which is where `SCRIPT_PATH` points to, we go up one level to `sagemaker/` then down to `sagemaker/run` where the entry points exist.

classicsong: OK

sagemaker/pipeline/create_sm_pipeline.py
```diff
@@ -0,0 +1,693 @@
...
+        self.pipeline_params: List[Union[ParameterString, ParameterInteger]] = []
+        self._create_pipeline_parameters(args)
+
+        self.next_step_data_input = self.args.task_config.input_data_s3
+        self.cache_config = CacheConfig(
+            enable_caching=True, expire_after=self.args.step_cache_expiration
+        )
+
+        # Build up the output prefix
+        # We use a hash of the execution parameters dict and
```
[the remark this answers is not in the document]
thvasilo: It's the default value for the execution subpath param in 

```
self.execution_subpath_param = self._create_string_parameter(
            "ExecutionSubpath", args.get_hash_hex()
        )
```

sagemaker/pipeline/create_sm_pipeline.py
thvasilo: Only `--output-format` seems to be an nargs='+' argument in GConstruct, so I'd say it's safe to keep this arg as is for this release, and decide whether we want to support multiple output formats for pipelines.

sagemaker/pipeline/pipeline_parameters.py
```diff
@@ -0,0 +1,880 @@
...
+                "Should not try to run both GConstruct and GSProcessing steps, "
+                f"got job sequence: {self.jobs_to_run}"
+            )
+
+        # When running gsprocessing ensure we run dist_part as well
+        if "gsprocessing" in self.jobs_to_run and "dist_part" not in self.jobs_to_run:
+            raise ValueError(
+                "When running GSProcessing need to run 'dist_part' as the following job, "
+                f"got job sequence: {self.jobs_to_run}"
+            )
```
thvasilo, line 172: Maybe we want to allow pipelines that only contain a GSProcessing step?

sagemaker/pipeline/README.md
```diff
@@ -145,7 +149,62 @@ The pipeline's behavior is controlled by various configuration parameters, inclu
...
+- `--gsprocessing-instance-count`: Number of GSProcessing instances.
+- `--volume-size-gb`: Additional volume size for SageMaker instances in GB. (Default: 100)
+
+### Task Configuration
+- `--graph-name`: Name of the graph. (Required)
+- `--input-data-s3`: S3 path to the input graph data. (Required)
+- `--output-prefix-s3`: S3 prefix for the output data. (Required)
+- `--pipeline-name`: Name for the pipeline.
+- `--base-job-name`: Base job name for SageMaker jobs. (Default: 'gs')
+- `--jobs-to-run`: Space-separated string of jobs to run in the pipeline. (Required)
```
classicsong, line 174: Is there any place, a user can find the names of jobs supported in the pipeline?

sagemaker/pipeline/README.md
```diff
@@ -145,7 +149,62 @@ The pipeline's behavior is controlled by various configuration parameters, inclu
...
+
+### Partition Configuration
+- `--partition-algorithm`: Partitioning algorithm. (Default: random)
+- `--partition-output-json`: Name for the output JSON file that describes the partitioned data. (Default: metadata.json)
+- `--partition-input-json`: Name for the JSON file that describes the input data for partitioning. (Default: updated_row_counts_metadata.json)
+
+### Training Configuration
+- `--model-output-path`: S3 path for model output.
+- `--num-trainers`: Number of trainers to use during training/inference. (Default: 4)
+- `--train-inference-task`: Task type for training and inference. (Required)
```
classicsong, line 191:
```suggestion
- `--train-inference-task-type`: Task type for training and inference. (Required)
```
The name train-inference-task is confusing.

sagemaker/pipeline/README.md
```diff
@@ -171,10 +230,11 @@ python create_sm_pipeline.py \
...
 
 This will create a pipeline that uses GSProcessing to process and prepare the data for partitioning,
-use DistPart to partition the data, convert the partitioned data to the GraphBolt format,
-then run a train and an inference job in sequence.
+uses DistPart to partition the data, converts the partitioned data to the GraphBolt format,
+then runs a train and an inference job in sequence.
 You can use this job sequence when your graph is too large to partition on one instance using
-GConstruct (1+ TB is the suggested threshold to move to distributed partitioning).
+GConstruct. 10B+ edges is the suggested threshold to move to distributed partitioning, or if your
+features are larger than 1TB.
```
classicsong, line 237:
```suggestion
features are larger than 1TByte.
```

sagemaker/pipeline/create_sm_pipeline.py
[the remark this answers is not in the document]
classicsong: I c.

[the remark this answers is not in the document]
classicsong: Can we add a warning here?

sagemaker/pipeline/create_sm_pipeline.py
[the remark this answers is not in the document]
classicsong: Maybe we should add some comments here to provide an example.

sagemaker/pipeline/create_sm_pipeline.py
classicsong: Can you add a comment here, saying Inference will also run as a SageMaker Training job.

sagemaker/pipeline/create_sm_pipeline.py
classicsong: Please add a TODO here.

sagemaker/pipeline/create_sm_pipeline.py
classicsong: Will it cause scalability issues?

sagemaker/pipeline/pipeline_parameters.py
classicsong: gconstruct is simpler. We can leave this question to the next release.

sagemaker/pipeline/README.md
thvasilo: Good point, currently they need to run `--help` in the creation, I'll document them here as well.

sagemaker/pipeline/create_sm_pipeline.py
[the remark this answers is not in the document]
thvasilo: We could print warning if the pipeline is using `GSProcessing->GSPartition->...` that it's not possible to go from `--use-graphbolt false` to `--use-graphbolt false`.

Although I'm always on the side of not overriding user input when that would change user intent. So in this case I'd vote for not exposing `--use-graphbolt` at execution time, but only on deployment time, even if that is possible when using GConstruct.

classicsong (approved):
LGTM

PR #1116, opened 2024-12-18
[SageMaker] Add SageMaker GraphBolt internal implementation

thvasilo:
*Issue #, if available:*

*Description of changes:*

* The original GraphBolt on SageMaker PR was missing the internal (GSF library) implementation, we only provided the launch script and entry point. This adds the missing module in the GraphStorm library.

### DIFF
--- a/python/graphstorm/sagemaker/sagemaker_gb_convert.py
+++ b/python/graphstorm/sagemaker/sagemaker_gb_convert.py
@@ -1,0 +1,95 @@
+""" Copyright 2023 Contributors
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+    SageMaker implementation of DistDGL to GraphBolt conversion.
+"""
+
+import importlib.metadata
+import logging
+import os
+import time
+from packaging import version
+
+import boto3
+import sagemaker
+from sagemaker.s3 import S3Uploader
+
+from graphstorm.gpartition.convert_to_graphbolt import run_gb_conversion
+
+
+def run_gb_convert(s3_output_path: str, local_dist_part_config: str, njobs: int):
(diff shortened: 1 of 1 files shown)


python/graphstorm/sagemaker/sagemaker_gb_convert.py
```diff
@@ -0,0 +1,95 @@
...
+    ), f"{local_dist_part_config=} does not exist."
+
+    dgl_version = importlib.metadata.version("dgl")
+    if version.parse(dgl_version) < version.parse("2.1.0"):
+        raise ValueError(
+            "GraphBolt conversion requires DGL version >= 2.1.0, "
+            f"but DGL version was {dgl_version}. "
+        )
+
+    boto_session = boto3.Session(region_name=os.environ['AWS_REGION'])
```
jalencato, line 62: Do we have boto3 dependencies in our docker?

python/graphstorm/sagemaker/sagemaker_gb_convert.py
```diff
@@ -0,0 +1,95 @@
...
+    gb_start = time.time()
+    run_gb_conversion(local_dist_part_config, njobs)
+    logging.info("GraphBolt conversion took %f sec.", time.time() - gb_start)
+
+    # Iterate through the partition data and upload only the modified/new
+    # files to the corresponding path on S3
+    upload_start = time.time()
+    for root, _, files in os.walk(os.path.dirname(local_dist_part_config)):
+        for file in files:
+            if file.endswith(("fused_csc_sampling_graph.pt", ".json")):
```
jalencato, line 76: How about we add a check here about if there is no ``fused_csc_sampling_graph.pt`` here? Just for consistency.

jalencato (commented):
Do we have a test on SageMaker for this PR's newly introduced function?

python/graphstorm/sagemaker/sagemaker_gb_convert.py
thvasilo: `sagemaker` depends on `boto` and `sagemaker` is installed on all SM images we use so we should be OK here.

I am planning a small update to our SM images to pin sagemaker, boto and botocore to specific versions.

thvasilo:
> Do we have a test on SageMaker for this PR's newly introduced function?

You mean a regression test? I don't think we have any unit test for any of the sagemaker code.

jalencato:
> > Do we have a test on SageMaker for this PR's newly introduced function?
> 
> You mean a regression test? I don't think we have any unit test for any of the sagemaker code.

I mean we add a new regression test after we merge this PR, as no current unit test covers the function here.

jalencato (approved):
LGTM.

PR #1117, opened 2024-12-19
[GSProcessing] Update GSProcessing custom split config parsing and documentation to match GConstruct.

thvasilo:
NOTE: GSProcessing now requires list[str] for its custom split file config input.

*Issue #, if available:*

*Description of changes:*

* Update GSProcessingdocumentation for custom split files to match implementation.
* Make GSProcessing config parsing be equivalent to GConstruct by allowing only one of train/val/test to be defined. Previously GSProcessing required all, but GConstruct did not.
* Make the custom file input for GSProcessing to _require_ list of files. When possible we want GSProcessing input config to have _one_ way to define things, allowing easier verification and reducing use confusion.
* Also used GenAI to generate some tests for `gsprocessing.config` classes and functions that were previously untested.

### DIFF
--- a/graphstorm-processing/graphstorm_processing/config/config_conversion/gconstruct_converter.py
+++ b/graphstorm-processing/graphstorm_processing/config/config_conversion/gconstruct_converter.py
@@ -61,7 +61,7 @@ class GConstructConfigConverter(ConfigConverter):
                         # check if split_pct is valid
                         assert (
                             math.fsum(label_splitrate) == 1.0
-                        ), f"sum of the label split rate should be ==1.0, got {label_splitrate=}"
+                        ), f"sum of the label split rate should be == 1.0, got {label_splitrate}"
                         label_dict["split_rate"] = {
                             "train": label_splitrate[0],
                             "val": label_splitrate[1],
--- a/graphstorm-processing/graphstorm_processing/config/config_conversion/gconstruct_converter.py
+++ b/graphstorm-processing/graphstorm_processing/config/config_conversion/gconstruct_converter.py
@@ -50,7 +50,7 @@ class GConstructConfigConverter(ConfigConverter):
         labels_list = []
         if labels in [[], [{}]]:
             return []
-        for label in labels:
+        for label in labels:  # pylint: disable=too-many-nested-blocks
             try:
                 label_column = label["label_col"] if "label_col" in label else ""
                 label_type = label["task_type"]
@@ -61,7 +61,7 @@ class GConstructConfigConverter(ConfigConverter):
                         # check if split_pct is valid
                         assert (
                             math.fsum(label_splitrate) == 1.0
-                        ), "sum of the label split rate should be ==1.0"
+                        ), f"sum of the label split rate should be ==1.0, got {label_splitrate=}"
                         label_dict["split_rate"] = {
                             "train": label_splitrate[0],
                             "val": label_splitrate[1],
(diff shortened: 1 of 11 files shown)


graphstorm-processing/graphstorm_processing/data_transformations/dist_label_loader.py
```diff
@@ -73,19 +73,19 @@ class CustomSplit:
 
     Parameters
     ----------
-    train : str
-        Path of the training mask parquet file.
+    train : list[str]
```
jalencato, line 76: Why is train contains list[str], not all train/test/valid?

graphstorm-processing/tests/test_converter.py
```diff
@@ -24,7 +24,7 @@
 @pytest.fixture(name="converter")
 def fixture_create_converter() -> GConstructConfigConverter:
     """Creates a new converter object for each test."""
-    yield GConstructConfigConverter()
+    return GConstructConfigConverter()
```
jalencato, line 27: Any reason we change here?

graphstorm-processing/graphstorm_processing/config/config_conversion/gconstruct_converter.py
```diff
@@ -61,19 +61,39 @@ def _convert_label(labels: list[dict]) -> list[dict]:
                         # check if split_pct is valid
                         assert (
                             math.fsum(label_splitrate) == 1.0
-                        ), "sum of the label split rate should be ==1.0"
+                        ), f"sum of the label split rate should be ==1.0, got {label_splitrate=}"
```
jalencato, line 64:
```suggestion
                        ), f"sum of the label split rate should be == 1.0, got {label_splitrate}"
```

graphstorm-processing/graphstorm_processing/config/config_conversion/gconstruct_converter.py
```diff
@@ -61,19 +61,39 @@ def _convert_label(labels: list[dict]) -> list[dict]:
...
+                        "At least one of ['train', 'valid', 'test'] "
+                        "needs to exist in custom split configs."
+                    )
+
+                    # Fill in missing values if needed
+                    for entry in ["train", "valid", "test", "column"]:
+                        entry_val = label_custom_split_filenames.get(entry, None)
+                        if entry_val:
+                            if isinstance(entry_val, str):
+                                label_custom_split_filenames[entry] = [entry_val]
```
jalencato, line 85: In this way do we want to support string input for gsprocessing config?

thvasilo: Yes, since GConstruct supports both, we use this to convert to list[str] which is our expectation for GSProcessing.

graphstorm-processing/graphstorm_processing/data_transformations/dist_label_loader.py
thvasilo: Yeah need to fix to docstring.

graphstorm-processing/tests/test_converter.py
thvasilo: This matches the return type of the function, `yield GConstructConfigConverter()` returns a `Generator[GConstructConfigConverter]`, `return GConstructConfigConverter()` returns `GConstructConfigConverter` which is the function signature.

In pytest we use `yield` fixtures when we want to include cleanup/teardown code in the fixture _after_ the `yield`, see https://docs.pytest.org/en/6.2.x/fixture.html#teardown-cleanup-aka-fixture-finalization

Here that's not needed.

graphstorm-processing/pyproject.toml
```diff
@@ -41,6 +41,7 @@ pre-commit = "^3.3.3"
 types-mock = "^5.1.0.1"
 pylint = "~2.17.5"
 diff-cover = "^9.0.0"
+pytest-cov = "^6.0.0"
```
thvasilo, line 44: Adding this to get coverage reports easier using `pytest --cov=graphstorm_processing tests/`

Specifically, I wanted to be able to see the dynamic contexts, which for each line of code shows us which test covers that line, see https://coverage.readthedocs.io/en/latest/contexts.html#dynamic-contexts

https://pytest-cov.readthedocs.io/en/stable/readme.html#usage

graphstorm-processing/graphstorm_processing/config/config_conversion/gconstruct_converter.py
jalencato: Right. But I suppose the code here only convert the gconstruct config, do we also support the config in gsprocessing config?

graphstorm-processing/pyproject.toml
jalencato: Do we need to add dependencies requirements also in the doc?

thvasilo: These are dev-only dependencies and are covered when a developer runs `poetry install --with dev`. We include this in the dev guide: https://graphstorm.readthedocs.io/en/latest/cli/graph-construction/distributed/gsprocessing/developer-guide.html#install-dependencies-through-poetry

graphstorm-processing/graphstorm_processing/config/config_conversion/gconstruct_converter.py
thvasilo: I would prefer having a single way to do things to avoid confusing users. This goes along with the Python principle of "There should be one-- and preferably only one --obvious way to do it." https://peps.python.org/pep-0020/

python/graphstorm/gconstruct/file_io.py
```diff
@@ -118,7 +118,7 @@ def read_index_parquet(data_file, column):
 
     if len(column) == 1:
         res_array = df[column[0]].to_numpy()
-    elif len(df.columns) == 2:
+    elif len(column) == 2:
```
jalencato, line 121: What is this change for?

jalencato (approved):
Approved with a small comment.

python/graphstorm/gconstruct/file_io.py
thvasilo: Make this check consistent with the previous one, 

```python
    if len(column) == 1:
        res_array = df[column[0]].to_numpy()
    elif len(column) == 2:
```

PR #1120, opened 2024-12-23
Fix/readme paper100m

YukeWang96:
There are some issues when running the following commands from https://graphstorm.readthedocs.io/en/latest/cli/model-training-inference/distributed/cluster.html#train-a-large-graph-ogbn-papers100m

image

image

We later figure out that it is because of the last m in papers100m datasets should be lower case instead of M.

### DIFF
--- a/docs/source/cli/model-training-inference/distributed/cluster.rst
+++ b/docs/source/cli/model-training-inference/distributed/cluster.rst
@@ -194,7 +194,7 @@ Distribute Partitioned Graphs and Configurations to all Instances
 ...................................................................
 In this step, users need to copy these partitioned files to the shared file system of the GraphStorm cluster. And the IP list file creation and 2222 port open operations are identical to the above OGBN-MAG section.
 
-For the OGBN-Papers100M data, we use a YAML file, ``ogbn_papers100M_nc_p3.yaml``, that has the contents below.
+For the OGBN-Papers100M data, we use a YAML file, ``ogbn_papers100m_nc_p3.yaml``, that has the contents below.
 
 .. code-block:: yaml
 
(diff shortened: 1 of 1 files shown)


docs/source/cli/model-training-inference/distributed/cluster.rst
```diff
@@ -180,21 +180,21 @@ Run the below command to download and partition the OGBN-Papers100M data for a n
...
                                                 --filepath /data \
                                                 --num-parts 3 \
                                                 --train-pct 0.1 \
                                                 --balance-train \
                                                 --balance-edges \
-                                                --output /data/ogbn_papers100M_3p \
+                                                --output /data/ogbn_papers100m_3p \
 
-Given the size of OGBN-Papers100M, the download and partition process could run more than 5 hours and consume around 700GB memory in peak. After the command completes, the partitioned OGBN-Papers100M graphs are stored in the ``/data/ogbn_papers100M_3p`` folder whose structure is the same as the OGBN-MAG's.
+Given the size of OGBN-Papers100M, the download and partition process could run more than 5 hours and consume around 700GB memory in peak. After the command completes, the partitioned OGBN-Papers100M graphs are stored in the ``/data/ogbn_papers100m_3p`` folder whose structure is the same as the OGBN-MAG's.
```
classicsong, line 191:
```suggestion
Given the size of OGBN-Papers100m, the download and partition process could run more than 5 hours and consume around 700GB memory in peak. After the command completes, the partitioned OGBN-Papers100m graphs are stored in the ``/data/ogbn_papers100m_3p`` folder whose structure is the same as the OGBN-MAG's.
```

classicsong (approved):
Please update the PR according to the comments.

PR #1128, opened 2025-01-06
[Bug Fix] Fix Edge Feature None block error

jalencato:
*Issue #, if available:*

*Description of changes:*

Previously (before adding edge feature in message passing), we do not require any block information, so the code works fine. Now as we require them for edge feats, we need to make a small fix about it. The error is only triggered in large dataset with workers > 24. Already test successfully with amazon-review dataset with this PR.


classicsong (approved):
LGTM

PR #1129, opened 2025-01-08
Fix typo in sagemaker launcher

classicsong:
*Issue #, if available:*
There is a typo in SUPPORTED_INFER_TASKS.

*Description of changes:*
Change `comput_emb` to `compute_emb`


thvasilo (commented):
LGTM, typo fix. Ping @jalencato to also take a look, we should have a mechanism to catch this kind of error, i.e. CI for sagemaker, possibly with local jobs.

Also w.r.t. to workflows, we don't need to run the regular GS CI for changes under the `sagemaker/` path.

PR #1132, opened 2025-01-12
[SageMaker] Fix download location for model inference

### LINKED ISSUE #949: SageMaker inference download locations are misconfigured, and models are downloaded twice
GSF tries to download the models into /opt/ml/gsgnn_model, as seen here https://github.com/thvasilo/graphstorm/blob/8e7c4c2e10accb114f2beccaa36ec3094d01241c/python/graphstorm/sagemaker/sagemaker_infer.py#L173

One a job with large model (learnable embeddings included) we see this in the logs in terms of disk space:

```
Filesystem      Size  Used Avail Use% Mounted on
overlay         120G   31G   90G  26% /
tmpfs            64M     0   64M   0% /dev
tmpfs           374G     0  374G   0% /sys/fs/cgroup
/dev/nvme0n1p1   70G   47G   24G  67% /usr/sbin/docker-init
/dev/nvme2n1   1008G  178G  779G  19% /tmp
shm             372G     0  372G   0% /dev/shm
/dev/nvme1n1    120G   31G   90G  26% /etc/hosts
tmpfs           374G     0  374G   0% /proc/acpi
tmpfs           374G     0  374G   0% /sys/firmware
```

The partition mounted under `/`, and I think that includes `/opt` , will only have 90GB available.

To be able to download larger datasets/models we need to be used the partition mounted under `/tmp` .

Also, in our inference launch script we define

https://github.com/thvasilo/graphstorm/blob/8e7c4c2e10accb114f2beccaa36ec3094d01241c/sagemaker/launch/launch_infer.py#L120

That will download the model data from the provided S3 path into /opt/ml/input/data/<channel_name> which by default for models will be /opt/ml/input/data/model  (see the [Estimator](https://sagemaker.readthedocs.io/en/stable/api/training/estimators.html#sagemaker.estimator.EstimatorBase) docs)

But then [here](https://github.com/thvasilo/graphstorm/blob/8e7c4c2e10accb114f2beccaa36ec3094d01241c/python/graphstorm/sagemaker/sagemaker_infer.py#L259), we try to download the model again, this time into /opt/ml/gsgnn_model

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Remove double download of model files and ensure models are downloaded under /tmp
* Tested by running a node classification job with ogb-arxiv data

Fixes #949

### DIFF
--- a/sagemaker/launch/launch_infer.py
+++ b/sagemaker/launch/launch_infer.py
@@ -118,7 +118,6 @@ def run_job(input_args, image, unknownargs):
         role=role,
         instance_count=instance_count,
         instance_type=instance_type,
-        model_uri=model_artifact_s3,
         py_version="py3",
         base_job_name=prefix,
         hyperparameters=params,
--- a/sagemaker/launch/launch_train.py
+++ b/sagemaker/launch/launch_train.py
@@ -92,15 +92,15 @@ def run_job(input_args, image, unknowargs):
 
     est = PyTorch(
         entry_point=os.path.basename(entry_point),
-        source_dir=os.path.dirname(entry_point),
+        hyperparameters=params,
         image_uri=container_image_uri,
-        role=role,
         instance_count=instance_count,
         instance_type=instance_type,
         output_path=model_artifact_s3,
         py_version="py3",
-        base_job_name=prefix,
-        hyperparameters=params,
+        role=role,
+        sagemaker_session=sess,
+        source_dir=os.path.dirname(entry_point),
         tags=[{"Key":"GraphStorm", "Value":"oss"},
               {"Key":"GraphStorm_Task", "Value":"Training"}],
         **estimator_kwargs
--- a/sagemaker/launch/launch_train.py
+++ b/sagemaker/launch/launch_train.py
@@ -66,8 +66,6 @@ def run_job(input_args, image, unknowargs):
 
     container_image_uri = image
 
-    prefix = f"gs-train-{graph_name}"
-
     params = {
         "graph-data-s3": graph_data_s3,
         "graph-name": graph_name,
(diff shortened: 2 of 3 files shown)


sagemaker/launch/launch_train.py
```diff
@@ -92,15 +92,15 @@ def run_job(input_args, image, unknowargs):
...
+        hyperparameters=params,
         image_uri=container_image_uri,
-        role=role,
         instance_count=instance_count,
         instance_type=instance_type,
         output_path=model_artifact_s3,
         py_version="py3",
-        base_job_name=prefix,
-        hyperparameters=params,
+        role=role,
```
classicsong, line 101: Do we still need the local variable `prefix`?

sagemaker/launch/launch_infer.py
```diff
@@ -118,7 +118,6 @@ def run_job(input_args, image, unknownargs):
         role=role,
         instance_count=instance_count,
         instance_type=instance_type,
-        model_uri=model_artifact_s3,
```
classicsong, line 121: This is why it downloads the model twice, right? Once by SageMaker and once by download_model(model_artifact_s3, model_path, sagemaker_session).

Can you add the above root cause in PR description?

classicsong (approved):

sagemaker/launch/launch_infer.py
jalencato: What does the model_uri=model_artifact_s3 here actually download? Does it only download the sagemaker artifact?

thvasilo: AFAIK SageMaker will download whatever is under model_artifact_s3. It expects an SM-formatted model file, that is a `model.tar.gz`.

There's some of that logic in the SM SDK: https://github.com/aws/sagemaker-python-sdk/blob/a58654ecbd5997aa59e4fcd8389715fe04c71f5d/src/sagemaker/estimator.py#L1881

Issue #1072, opened 2024-10-21
graphstorm images utilisation in SageMaker TuningJobs


milianru:
Hello,

I was facing some issues with colliding model artifacts using the graphstorm training image in a TuningJob.

I was wondering, if the graphstorm docker images are in general intended for usage in Sagemaker TuningJobs? And if so, is there an example on how to orchestrate model artifacts of each training job?

When we constructed a `HyperparameterTuner` with a `PyTorch`  estimator using the training images, the result was first unexpected. All training jobs wrote their artifacts to `--model-artifact-s3`, without building up the S3 prefixes for each training run - as I knew it from other training images.

After diving a bit into the code and the AWS documentation I found that Sagemaker moves everything stored under '/opt/ml/code` to the output prefix specified by the estimator object. In contrast to this common interface the graphstorm container implement the logic controlled by '--model-artifact-s3`, which works perfectly fine for a single training job, but results in model artifacts collisions for a TuningJob. 

We found a way to workaround the issue by creating a custom training entry script. It which basically extended the regular train_entry.py with a final copy of the training artifacts to `/opt/ml/model`. Having the model artifacts on the containter stored there, allowed the TuniningJob to orchestrate all training artifacts successfully.

I was wondering, whether the model storage in `graphstorm/sagemaker/sagemaker_train.py` under `/tmp/gsgnn_model/`  is prefered over `/opt/ml/model` for a different reason ([here](https://github.com/awslabs/graphstorm/blob/main/python/graphstorm/sagemaker/sagemaker_train.py#L167))?  If this is the case, would it be possible to define a new argument for the local storage directory or similar?

In case this is something that you don't have on your agenda, but you see the purpose, I could also assist with a PR.

thvasilo:
Hi @milianru I think this actually a shortcoming of our SageMaker training output implementation. 

If you're willing to contribute a fix I can shepherd the PR and try to get it merged for the next release.

classicsong:
Hi, @milianru, for the place where GraphStorm stores the model artifacts, we use `/tmp/gsgnn_model/` because we will upload the model to an S3 bucket (defined by `--model-artifact-s3`). GraphStorm does not rely on SageMaker to upload the model artifact (as sometime it will be very large).

You are free to adujst the training entry script if it works :).
Another way is to see if HyperparameterTuner can automatically generate arguments for `--model-artifact-s3`.

milianru:
Hi @thvasilo,
[...]

PR #1133, opened 2025-01-13
[SageMaker] Add launcher and support for SageMaker HPO jobs

thvasilo:
*Issue #, if available:*

Fixes https://github.com/awslabs/graphstorm/issues/1072

*Description of changes:*

* We change train_entry and sagmaker_train so that when the user does not provide a value for the `--model-artifact-s3` argument, we rely on the SageMaker service to upload model artifacts. This makes it possible to run HPO and other jobs that rely on using the "official" SageMaker model paths.
* Add example launch script for HPO jobs and documentation.

### DIFF
--- a/docs/source/cli/model-training-inference/distributed/sagemaker.rst
+++ b/docs/source/cli/model-training-inference/distributed/sagemaker.rst
@@ -252,8 +252,8 @@ For example your JSON file can look like:
             "IntegerParameterRanges": [
                 {
                     "Name": "hidden_size",
-                    "MaxValue": "64",
-                    "MinValue": "256",
+                    "MinValue": "64",
+                    "MaxValue": "256",
                     "ScalingType": "Auto"
                 }
             ]
--- a/docs/source/cli/model-training-inference/distributed/sagemaker.rst
+++ b/docs/source/cli/model-training-inference/distributed/sagemaker.rst
@@ -221,7 +221,7 @@ with SageMaker AI,
 which allows you to optimize the hyper-parameters
 of your model with an easy-to-use interface.
 
-The ``launch/launch_hyperparameter_tuning.py`` script can act as a thin
+The ``sagemaker/launch/launch_hyperparameter_tuning.py`` script can act as a thin
 wrapper for SageMaker's `HyperParameterTuner <https://sagemaker.readthedocs.io/en/stable/api/training/tuner.html>`_.
 
 You define the hyper-parameters of interest by passing a filepath to a JSON file,
@@ -279,7 +279,7 @@ e.g. ``"accuracy"``. See the entry for ``eval_metric`` in :ref:`Evaluation Metri
 for a full list of supported metrics.
 
 ``--metric-dataset`` defines which dataset to collect metrics from, and
-can be either ``"test"`` or ``"val"`` to collect metrics during test or validation
+can be either ``"test"`` or ``"val"`` to collect metrics from test or validation set
 respectively. Finally use ``--objective-type`` to set the type of the objective,
 which can be either ``"Maximize"`` or ``"Minimize"``.
 
--- a/docs/source/cli/model-training-inference/distributed/sagemaker.rst
+++ b/docs/source/cli/model-training-inference/distributed/sagemaker.rst
@@ -274,14 +274,16 @@ string that directly corresponds to one of SageMaker's
 `scaling types <https://docs.aws.amazon.com/sagemaker/latest/dg/automatic-model-tuning-define-ranges.html#scaling-type>`_.
 By default scaling type will be ``'Auto'``.
 
-Use ``--metric-name`` to define the name of metric to use as a tuning objective,
+Use ``--metric-name`` to provide the name of a GraphStorm metric to use as a tuning objective,
 e.g. ``"accuracy"``. See the entry for ``eval_metric`` in :ref:`Evaluation Metrics <eval_metrics>`
 for a full list of supported metrics.
 
-``--metric-dataset`` defines which dataset to collect metrics from, and
+``--eval-mask`` defines which dataset to collect metrics from, and
 can be either ``"test"`` or ``"val"`` to collect metrics from test or validation set
 respectively. Finally use ``--objective-type`` to set the type of the objective,
 which can be either ``"Maximize"`` or ``"Minimize"``.
+See the `SageMaker documentation <https://docs.aws.amazon.com/sagemaker/latest/dg/automatic-model-tuning.html>`_
+for more details
 
 Finally you can use ``--strategy`` to select the optimization strategy
 from one of "Bayesian", "Random", "Hyperband", "Grid". See the
@@ -306,7 +308,7 @@ Example HPO call:
         --max-parallel-jobs 4 \
         --hyperparameter-ranges my_param_ranges.json \
         --metric-name "accuracy" \
-        --metric-dataset "val" \
+        --eval-mask "val" \
         --objective-type "Maximize" \
         --strategy "Bayesian"
 
--- a/docs/source/cli/model-training-inference/distributed/sagemaker.rst
+++ b/docs/source/cli/model-training-inference/distributed/sagemaker.rst
@@ -279,7 +279,7 @@ e.g. ``"accuracy"``. See the entry for ``eval_metric`` in :ref:`Evaluation Metri
 for a full list of supported metrics.
 
 ``--eval-mask`` defines which dataset to collect metrics from, and
-can be either ``"test"`` or ``"val"`` to collect metrics from test or validation set
+can be either ``"test"`` or ``"val"`` to collect metrics from test or validation set,
 respectively. Finally use ``--objective-type`` to set the type of the objective,
 which can be either ``"Maximize"`` or ``"Minimize"``.
 See the `SageMaker documentation <https://docs.aws.amazon.com/sagemaker/latest/dg/automatic-model-tuning.html>`_
--- a/docs/source/cli/model-training-inference/distributed/sagemaker.rst
+++ b/docs/source/cli/model-training-inference/distributed/sagemaker.rst
@@ -224,32 +224,55 @@ of your model with an easy-to-use interface.
 The ``launch/launch_hyperparameter_tuning.py`` script can act as a thin
 wrapper for SageMaker's `HyperParameterTuner <https://sagemaker.readthedocs.io/en/stable/api/training/tuner.html>`_.
 
-You define the hyper-parameters of interest by passing a python dictionary as a string,
-where the keys of the dictionary are the names of the parameters to tune,
-and each value is a dictionary that defines the type and tuning range
-of the parameter.
+You define the hyper-parameters of interest by passing a filepath to a JSON file,
+or a python dictionary as a string,
+where the structure of the dictionary is the same as for SageMaker's
+`Dynamic hyper-parameters <https://docs.aws.amazon.com/sagemaker/latest/dg/automatic-model-tuning-define-ranges.html#automatic-model-tuning-define-ranges-dynamic>`.
+For example your JSON file can look like:
+
+.. code:: bash
+
+    # Content of my_param_ranges.json
+    {
+        "ParameterRanges": {
+            "CategoricalParameterRanges": [
+                {
+                    "Name": "model_encoder_type",
+                    "Values": ["rgcn", "hgt"]
+                }
+            ],
+            "ContinuousParameterRanges": [
+                {
+                    "Name": "lr",
+                    "MinValue": "1e-5",
+                    "MaxValue" : "1e-2",
+                    "ScalingType": "Auto"
+                }
+            ],
+            "IntegerParameterRanges": [
+                {
+                    "Name": "hidden_size",
+                    "MaxValue": "64",
+                    "MinValue": "256",
+                    "ScalingType": "Auto"
+                }
+            ]
+        }
+    }
+
+Which you can then use to launch an HPO job:
 
 .. code:: bash
 
     # Example hyper-parameter ranges
     python launch/launch_hyperparameter_tuning.py \
-        --hyperparameter-ranges '{"lr": {"type": "continuous", "min": 1e-5, "max": 1e-2, "scaling_type": "Auto"}, "num_layers": {"type": "integer", "min": 1, "max": 3}, "model_encoder_type": {"type": "categorical", "values": ["rgcn", "rgat"]}}'
+        --hyperparameter-ranges my_param_ranges.json
+        # Other launch parameters...
 
-The supported parameter types are ``continuous``
-for real-values parameters, ``integer`` for numerical integer parameters,
-and ``categorical`` for discrete-value string parameters.
-These directly correspond to Sagemaker's
-`Dynamic hyper-parameters <https://docs.aws.amazon.com/sagemaker/latest/dg/automatic-model-tuning-define-ranges.html#automatic-model-tuning-define-ranges-dynamic>`~
-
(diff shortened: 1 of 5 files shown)


docs/source/cli/model-training-inference/distributed/sagemaker.rst
```diff
@@ -213,6 +213,80 @@ from ``${DATASET_S3_PATH}`` as input and create a DistDGL graph with
...
+You define the hyper-parameters of interest by passing a python dictionary as a string,
+where the keys of the dictionary are the names of the parameters to tune,
+and each value is a dictionary that defines the type and tuning range
+of the parameter.
+
+.. code:: bash
+
+    # Example hyper-parameter ranges
+    python launch/launch_hyperparameter_tuning.py \
+        --hyperparameter-ranges '{"lr": {"type": "continuous", "min": 1e-5, "max": 1e-2, "scaling_type": "Auto"}, "num_layers": {"type": "integer", "min": 1, "max": 3}, "model_encoder_type": {"type": "categorical", "values": ["rgcn", "rgat"]}}'
```
classicsong, line 236: Should a json or yaml configuration file be a better interface for users to use?

docs/source/cli/model-training-inference/distributed/sagemaker.rst
```diff
@@ -213,6 +213,80 @@ from ``${DATASET_S3_PATH}`` as input and create a DistDGL graph with
...
+
+The supported parameter types are ``continuous``
+for real-values parameters, ``integer`` for numerical integer parameters,
+and ``categorical`` for discrete-value string parameters.
+These directly correspond to Sagemaker's
+`Dynamic hyper-parameters <https://docs.aws.amazon.com/sagemaker/latest/dg/automatic-model-tuning-define-ranges.html#automatic-model-tuning-define-ranges-dynamic>`~
+
+Continuous and integer parameters need to define a ``"min"`` and ``"max"`` value to use
```
classicsong, line 244: I don't think we need L244-L246. Users can refer to the SageMaker doc for more information.
Otherwise, we have to track the changes from SageMaker.

docs/source/cli/model-training-inference/distributed/sagemaker.rst
```diff
@@ -213,6 +213,80 @@ from ``${DATASET_S3_PATH}`` as input and create a DistDGL graph with
...
+during tuning, while categorical variables need to provide a list of strings
+as ``"values"`` to choose from.
+
+For continuous and integer parameters you can also provide a ``scaling_type``
+string that directly corresponds to one of SageMaker's
+`scaling types <https://docs.aws.amazon.com/sagemaker/latest/dg/automatic-model-tuning-define-ranges.html#scaling-type>`_,
+i.e. can be 'Auto', 'Linear', 'Logarithmic', or 'ReverseLogarithmic'.
```
classicsong, line 251:
```suggestion
```

python/graphstorm/sagemaker/sagemaker_train.py
```diff
@@ -164,7 +164,16 @@ def run_train(args, unknownargs):
         os.makedirs(restore_model_path, exist_ok=True)
     else:
         restore_model_path = None
-    output_path = "/tmp/gsgnn_model/"
+
+    if args.model_artifact_s3:
+        # If user provided an S3 output destination as an input arg, the script itself
```
classicsong, line 169:
```suggestion
        # If a user provides an S3 output destination as an input arg, the script itself
```

python/graphstorm/sagemaker/sagemaker_train.py
```diff
@@ -164,7 +164,16 @@ def run_train(args, unknownargs):
...
+        # will upload the model artifacts after training, so we save under /tmp.
+        output_path = "/tmp/gsgnn_model/"
+    else:
+        # If the user did not provide an output destination as an arg, we rely on SageMaker to
```
classicsong, line 173:
```suggestion
        # If the user does not provide an output destination as an arg, we rely on SageMaker to
```

python/graphstorm/sagemaker/sagemaker_train.py
```diff
@@ -164,7 +164,16 @@ def run_train(args, unknownargs):
...
+        # do the model upload so we save the model to the pre-determined path /opt/ml/model
+        output_path = "/opt/ml/model"
```
classicsong, line 175: This mode won't work with learnable embeddings. The embeddings are saved on multiple machines.

python/graphstorm/sagemaker/sagemaker_train.py
```diff
@@ -164,7 +164,16 @@ def run_train(args, unknownargs):
         os.makedirs(restore_model_path, exist_ok=True)
     else:
         restore_model_path = None
-    output_path = "/tmp/gsgnn_model/"
+
+    if args.model_artifact_s3:
```
classicsong, line 168: Maybe we should add an assertion here.

sagemaker/run/train_entry.py
```diff
@@ -37,10 +37,11 @@ def get_train_parser():
...
         help="S3 location of training yaml file. "
-             "Do not store it with partitioned graph",
-             required=True)
-    parser.add_argument("--model-artifact-s3", type=str,
-        help="S3 location to store the model artifacts.")
+             "Do not store it with partitioned graph")
+    parser.add_argument("--model-artifact-s3", type=str, default=None,
+        help="S3 location to store the model artifacts. If None, we rely on SageMaker "
+        "to upload model artifacts, so the launching Estimator needs to have 'output_path' set. "
+        "Default: None")
```
classicsong, line 44: Maybe we should set it to `required=True`.

docs/source/cli/model-training-inference/distributed/sagemaker.rst
thvasilo: Yeah good point because formatting JSON in the command line can be tricky. I'll make this support either a file or a JSON string.

sagemaker/launch/launch_hyperparameter_tuning.py
```diff
@@ -0,0 +1,209 @@
...
+        hyperparameter_ranges=hyperparameter_ranges,
+        objective_type=args.objective_type,
+        max_jobs=args.max_jobs,
+        max_parallel_jobs=args.max_parallel_jobs,
+        metric_definitions=[
+            {
+                "Name": full_metric_name,
+                "Regex": (
+                    f"INFO:root:best_{args.metric_dataset}_score: "
+                    f"{{'{args.metric_name}': ([0-9\\.]+)}}"
```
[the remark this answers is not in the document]
thvasilo: I think we need a guaranteed metric log format so that it is consistent across metrics, SageMaker relies on that for HPO and general metric reporting. I'll take a look at the codebase and raise an issue if needed.

AFAIK we already produce metrics in a JSON format during logging?

sagemaker/run/train_entry.py
thvasilo: If we do that we won't be able to set the value to `None`, but we could change it to require a specific string. Not sure which is better UX.

docs/source/cli/model-training-inference/distributed/sagemaker.rst
thvasilo: Ah good point. Right now we use custom parsing of the ranges, i.e. our expected JSON syntax does not match SageMaker 1 to 1. 

Better if we instead directly use SageMaker's format and refer the users to that.

python/graphstorm/sagemaker/sagemaker_train.py
thvasilo: An assertion for the value of args.model_artifact_s3? Like `assert args.model_artifact_s3.startswith("s3://")` ?

sagemaker/launch/launch_hyperparameter_tuning.py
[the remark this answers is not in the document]
thvasilo: I checked `ep_trainer.py`, `np_trainer.py`, `glem_np_trainer.py`, `lp_trainer.py`, `mt_trainer.py` and the at least use this format to report metrics. And all metrics will report a number right? I think that ensures that, but if we add new trainers they would need to follow this format. 
Maybe I can create a follow-up PR that adds a function that all the above trainers use instead of each one re-implementing their own.

sagemaker/launch/launch_hyperparameter_tuning.py
```diff
@@ -0,0 +1,257 @@
...
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+
+    Launch SageMaker HPO jobs.
+
+    Usage:
```
classicsong, line 18:
```suggestion
    Example:
```

sagemaker/launch/launch_hyperparameter_tuning.py
```diff
@@ -0,0 +1,257 @@
...
+    Returns
+    -------
+    dict
+        Hyperparameter dict that can be used to create a HyperparameterTuner
+        object.
+
+    Raises
+    ------
+    ValueError
+        If the JSON dict contains an invalid parameter type.
```
classicsong, line 79: .. versionadded xxx

sagemaker/launch/launch_hyperparameter_tuning.py
```diff
@@ -0,0 +1,257 @@
...
+    }
+    if args.custom_script is not None:
+        params["custom-script"] = args.custom_script
+    if args.model_checkpoint_to_load is not None:
+        params["model-checkpoint-to-load"] = args.model_checkpoint_to_load
+
+    unknown_args_dict = parse_unknown_gs_args(unknownargs)
+    params.update(unknown_args_dict)
+
+    print(f"SageMaker launch parameters {params}")
```
classicsong, line 143: Use logging

sagemaker/launch/launch_hyperparameter_tuning.py
```diff
@@ -0,0 +1,257 @@
...
+            {"Key": "GraphStorm", "Value": "oss"},
+            {"Key": "GraphStorm_Task", "Value": "HPO"},
+        ],
+        **estimator_kwargs,
+    )
+
+    hyperparameter_ranges = parse_hyperparameter_ranges(args.hyperparameter_ranges)
+
+    # Construct the full metric name based on user input
+    full_metric_name = f"best_{args.metric_dataset}_score:{args.metric_name}"
```
classicsong, line 169: Can you double-check whether `args.metric_dataset` and `args.metric_name` is case-sensitive and whether we should deal with it?

sagemaker/launch/launch_hyperparameter_tuning.py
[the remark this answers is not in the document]
classicsong: The SageMaker tracker guarantees to produce the metric logs in a predefined format.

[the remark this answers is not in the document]
classicsong: We only need to make sure the new trainers will also support SageMaker tracker.

sagemaker/launch/launch_hyperparameter_tuning.py
```diff
@@ -0,0 +1,209 @@
...
+        help="JSON string defining hyperparameter ranges",
+    )
+    hpo_group.add_argument(
+        "--metric-name",
+        type=str,
+        required=True,
+        help="Name of the metric to optimize (e.g., 'accuracy', 'amri')",
+    )
+    hpo_group.add_argument(
+        "--metric-dataset",
```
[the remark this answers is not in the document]
classicsong: The metric-dataset is misleading.

docs/source/cli/model-training-inference/distributed/sagemaker.rst
```diff
@@ -213,6 +213,103 @@ from ``${DATASET_S3_PATH}`` as input and create a DistDGL graph with
...
+    python launch/launch_hyperparameter_tuning.py \
+        --hyperparameter-ranges my_param_ranges.json
+        # Other launch parameters...
+
+For continuous and integer parameters you can provide a ``ScalingType``
+string that directly corresponds to one of SageMaker's
+`scaling types <https://docs.aws.amazon.com/sagemaker/latest/dg/automatic-model-tuning-define-ranges.html#scaling-type>`_.
+By default scaling type will be ``'Auto'``.
+
+Use ``--metric-name`` to define the name of metric to use as a tuning objective,
```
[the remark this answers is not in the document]
thvasilo: Use needs to be one of our supported metrics.

sagemaker/launch/launch_hyperparameter_tuning.py
thvasilo: Like if the user provides `Accuracy` right? I'll add `lower()` here

docs/source/cli/model-training-inference/distributed/sagemaker.rst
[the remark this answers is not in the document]
thvasilo: Changed to `Use ``--metric-name`` to provide the name of a GraphStorm metric`

python/graphstorm/sagemaker/utils.py
```diff
@@ -15,18 +15,26 @@
...
+from typing import Optional
 from urllib.parse import urlparse
 
 import boto3
 from botocore.errorfactory import ClientError
 from sagemaker.s3 import S3Downloader
 from sagemaker.s3 import S3Uploader
 
+# Set a fixed seed for reproducibility
+SEED = 42
```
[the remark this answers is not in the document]
thvasilo: Ah I was using a consistent rng before this is a leftover. I'll remove it.

python/graphstorm/sagemaker/utils.py
```diff
@@ -455,3 +463,59 @@ def remove_embs(emb_path):
...
+    if not job_str_identifier:
+        job_str_identifier = os.getenv('SM_USER_ARGS', '')
+
+    # Create a hash of the unique identifier
+    hash_object = hashlib.md5(job_str_identifier.encode())
+    hash_hex = hash_object.hexdigest()
+
+    # Convert first 4 chars of hash to int and scale to valid port range
+    # Using 10000-65000 to avoid privileged ports and common ports
+    base_port = 10000 + (int(hash_hex[:4], 16) % 55000)
```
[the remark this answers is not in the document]
thvasilo: Also from the previous implementation I'll fix these.

python/graphstorm/sagemaker/utils.py
```diff
@@ -455,3 +463,59 @@ def remove_embs(emb_path):
         Local embedding path
     """
     remove_data(emb_path)
+
+def is_port_available(port):
+    """Check if a port is available."""
+    try:
+        with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
```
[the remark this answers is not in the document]
thvasilo: Ah good point, connect_ex will return an error instead of throwing an exception which is more efficient https://docs.python.org/3/library/socket.html#socket.socket.connect_ex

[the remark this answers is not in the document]
thvasilo: There's a subtle difference between the `bind` check and `connect`. `bind` checks if the port is available to create a service on.

`connect/connect_ex` are used to check if a service is already running on the given port.

The difference lies in the state of the port, e.g. for a priviledged port like `80`, `bind` will fail due to permissions, while `connect` will work if there's a service listening on the port.

classicsong (approved):

PR #1134, opened 2025-01-13
[SageMaker] Make default region value be None, so AWS config takes precedence

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Our current default for region in SageMaker jobs will override users' local configuration. By setting this to None by default we respect the local config, while allowing users to specify a different region if needed.


jalencato (approved):
LGTM

PR #1136, opened 2025-01-17
[GSProcessing] Enforce re-order for node label processing during classification

### LINKED ISSUE #1135: GSProcessing Local Run does not output file list in alphabetic order
GSProcessing does not output the file list in order when running locally, our current output is like:

```
"data": [
                    "node_data/type1-label-label/parquet/part-00034-a443894d-5d18-4361-8866-c0ac8f92ebe2-c000.snappy.parquet",
                    "node_data/type1-label-label/parquet/part-00025-a443894d-5d18-4361-8866-c0ac8f92ebe2-c000.snappy.parquet",
                    "node_data/type1-label-label/parquet/part-00003-a443894d-5d18-4361-8866-c0ac8f92ebe2-c000.snappy.parquet",
                    "node_data/type1-label-label/parquet/part-00051-a443894d-5d18-4361-8866-c0ac8f92ebe2-c000.snappy.parquet",

```

### LINKED ISSUE #1138: Fix label/mask order mismatch for node classification.

thvasilo:
*Issue #, if available:*

Fixes #1135 
Fixes #1138

*Description of changes:*

* We guarantee ordering for node label classification by ordering the transformed label DF after processing by the NODE_INT_MAPPING id, and doing the same for masks.
* Because there is no guarantee for order when writing to Parquet from Spark _even for ordered Spark DataFrames_, we collect the labels and masks to a Pandas DF on the Spark leader, and write that using pyarrow.
* NOTE: DistPart requires at least N files to be present for every node/edge type, where N is the number of requested partitions. So when writing back the masks and label we still use the number of incoming partitions to create as many files.

### DIFF
--- a/graphstorm-processing/graphstorm_processing/config/label_config_base.py
+++ b/graphstorm-processing/graphstorm_processing/config/label_config_base.py
@@ -18,6 +18,8 @@ import abc
 import logging
 from typing import Any, Dict, Optional
 
+from graphstorm_processing.constants import VALID_TASK_TYPES
+
 
 class LabelConfig(abc.ABC):
     """Basic class for label config"""
@@ -55,6 +57,9 @@ class LabelConfig(abc.ABC):
             self._mask_field_names = None
 
     def _sanity_check(self):
+        assert (
+            self._task_type in VALID_TASK_TYPES
+        ), f"Invalid task type {self._task_type}, must be one of {VALID_TASK_TYPES}"
         if self._label_column == "":
             assert self._task_type == "link_prediction", (
                 "When no label column is specified, the task type must be link_prediction, "
@@ -82,6 +87,25 @@ class LabelConfig(abc.ABC):
             assert isinstance(self._mask_field_names, list)
             assert all(isinstance(x, str) for x in self._mask_field_names)
             assert len(self._mask_field_names) == 3
+
+    def __repr__(self) -> str:
+        """Formal object representation for debugging"""
+        return (
+            f"{self.__class__.__name__}(label_column={self._label_column!r}, "
+            f"task_type={self._task_type!r}, separator={self._separator!r}, "
+            f"multilabel={self._multilabel!r}, split={self._split!r}, "
+            f"custom_split_filenames={self._custom_split_filenames!r}, "
+            f"mask_field_names={self._mask_field_names!r})"
+        )
+
+    def __str__(self) -> str:
+        """Informal object representation for readability"""
+        task_desc = f"{self._task_type} task"
+        if self._label_column:
+            task_desc += f" on column '{self._label_column}'"
+        if self._multilabel:
+            task_desc += f" (multilabel with separator '{self._separator}')"
+        return task_desc
 
     @property
     def label_column(self) -> str:
--- a/graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_label_transformation.py
+++ b/graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_label_transformation.py
@@ -45,6 +45,7 @@ class DistSingleLabelTransformation(DistributedTransformation):
 
     def apply(self, input_df: DataFrame) -> DataFrame:
         assert self.spark
+        original_cols = {*input_df.columns} - {self.label_column}
         processed_col_name = self.label_column + "_processed"
 
         str_indexer = StringIndexer(
@@ -63,13 +64,15 @@ class DistSingleLabelTransformation(DistributedTransformation):
 
         # Labels that were missing and were assigned the value numLabels by the StringIndexer
         # are converted to None
-        long_class_label = indexed_df.select(F.col(self.label_column).cast("long")).select(
+        long_class_label = indexed_df.select(
             F.when(
                 F.col(self.label_column) == len(str_indexer_model.labelsArray[0]),  # type: ignore
                 F.lit(None),
             )
             .otherwise(F.col(self.label_column))
-            .alias(self.label_column)
+            .cast("long")
+            .alias(self.label_column),
+            *original_cols,
         )
 
         # Get a mapping from original label to encoded value
@@ -112,7 +115,7 @@ class DistMultiLabelTransformation(DistMultiCategoryTransformation):
         super().__init__(cols, separator)
         self.label_column = cols[0]
 
-    def apply(self, input_df: DataFrame) -> DataFrame:
-        multi_cat_df = super().apply(input_df)
+    def apply(self, input_df: DataFrame, return_all_cols=True) -> DataFrame:
+        multi_cat_df = super().apply(input_df, return_all_cols=return_all_cols)
 
         return multi_cat_df
--- a/graphstorm-processing/graphstorm_processing/distributed_executor.py
+++ b/graphstorm-processing/graphstorm_processing/distributed_executor.py
@@ -264,6 +264,7 @@ class DistributedExecutor:
             output_prefix=self.output_prefix,
             precomputed_transformations=self.precomputed_transformations,
         )
+
         self.loader = DistHeterogeneousGraphLoader(
             self.spark,
             loader_config,
@@ -287,17 +288,18 @@ class DistributedExecutor:
             bucket, s3_prefix = s3_utils.extract_bucket_and_key(self.output_prefix)
             s3 = boto3.resource("s3")
 
-            output_files = os.listdir(loader.output_path)
+            output_files = os.listdir(loader.local_meta_output_path)
             for output_file in output_files:
                 s3.meta.client.upload_file(
-                    f"{os.path.join(loader.output_path, output_file)}",
+                    f"{os.path.join(loader.local_meta_output_path, output_file)}",
                     bucket,
                     f"{s3_prefix}/{output_file}",
                 )
 
     def run(self) -> None:
         """
-        Executes the Spark processing job.
+        Executes the Spark processing job, optional repartition job, and uploads any metadata files
+        if needed.
         """
         logging.info("Performing data processing with PySpark...")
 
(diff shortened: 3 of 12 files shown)


graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_label_transformation.py
```diff
@@ -63,13 +64,15 @@ def apply(self, input_df: DataFrame) -> DataFrame:
...
+        long_class_label = indexed_df.select(
             F.when(
                 F.col(self.label_column) == len(str_indexer_model.labelsArray[0]),  # type: ignore
                 F.lit(None),
             )
             .otherwise(F.col(self.label_column))
-            .alias(self.label_column)
+            .cast("long")
+            .alias(self.label_column),
+            *original_cols,
```
jalencato, line 75: So we add all the original columns here for the label value? Will it overkill?

graphstorm-processing/graphstorm_processing/distributed_executor.py
```diff
@@ -287,17 +288,18 @@ def _upload_output_files(self, loader: DistHeterogeneousGraphLoader, force=False
...
-                    f"{os.path.join(loader.output_path, output_file)}",
+                    f"{os.path.join(loader.local_meta_output_path, output_file)}",
                     bucket,
                     f"{s3_prefix}/{output_file}",
                 )
 
     def run(self) -> None:
         """
-        Executes the Spark processing job.
+        Executes the Spark processing job, optional repartition job, and uploads any metadata files
```
jalencato, line 301:
```suggestion
        Executes the Spark processing job, optional repartition job, and uploads transformed metadata files
```

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -559,6 +567,7 @@ def _write_df(
         if out_format == "parquet":
             # Write to parquet
             input_df = self._replace_special_chars_in_cols(input_df)
+            # TODO: Remove overwrite mode
```
jalencato, line 570: Any reason we would like to remove overwrite? I would say overwrite would be most recommended.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -622,6 +627,67 @@ def csv_row(data: Row):
...
+        list[str]
+            The list of files written, paths being relative to ``self.output_prefix``.
+        """
+
+        # TODO: Remove once we support writing multiple files
+        assert out_path.endswith(".parquet")
+        if self.filesystem_type == FilesystemType.LOCAL:
+            os.makedirs(os.path.dirname(out_path), exist_ok=True)
+
+        pq.write_table(
```
jalencato, line 659: Shall we move this part to line 666? And if emr-serverless support directly use pyarrow to write to S3?

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -1949,35 +2053,81 @@ def _create_split_files(
...
+        # Use custom column names if requested
+        if mask_field_names:
+            mask_names = mask_field_names
+        else:
+            mask_names = ("train_mask", "val_mask", "test_mask")
+
+        # TODO: Make this an argument to the write functions to not rely on function scope?
         split_metadata = {}
+
+        def write_masks_numpy(np_mask_arrays: Sequence[np.ndarray]):
```
jalencato, line 2066: What about put these two functions into utils.py?

jalencato (commented):
Do we have any performance number to compare between the built-in spark method and pandaUDF method?

graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_label_transformation.py
thvasilo: Hmm, well here we want to ensure that the order_col is preserved in the output, but `DistSingleLabelTransformation` is not aware of `order_col`. We could modify the constructor to also provide that, and then return just the label and order column.

The argument against would be that these two cols are selected downstream anyway, and the other columns are not materialized, so there shouldn't be any real performance penalty. 

I'm good with either option.

graphstorm-processing/graphstorm_processing/distributed_executor.py
thvasilo: This also uploads `perf_counters.json` , not sure if we would qualify that as a "transformed" metadata file.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: Yes, one thing we've observed (myself and with end users) is that they often run the same job more than once, keeping the same output-prefix (for example an error happens during a run, they fix part of the config and run again).

This results in some input files being duplicated under the same path, e.g. raw_id_mappings will have files for both names. When partition tries to read that it gets 2x the number of rows, which leads to errors. 

It's better if we enforce users to either clean output directories or select new ones to avoid this.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: We can move it yes. 

> emr-serverless support directly use pyarrow to write to S3

Yes we already use pyarrow to write and modify files on EMRS during the re-partition stage that runs on the Spark leader.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: My rule of thumb to make a function public is : "is this function used in two place in the codebase?". 

e.g. that's why I moved `_create_metadata_entry` from an inner function to be part DGHL in this PR.

Right now the answer is no for these functions, if we see a need to re-use them we can pull them out.

graphstorm-processing/graphstorm_processing/config/label_config_base.py
```diff
@@ -18,6 +18,8 @@
 import logging
 from typing import Any, Dict, Optional
 
+from graphstorm_processing.constants import VALID_TASK_TYPES
+
```
jalencato, line 22: Remove additional line.

graphstorm-processing/graphstorm_processing/repartition_files.py
```diff
@@ -194,8 +197,6 @@ def create_new_relative_path_from_existing(
             "path/to/parquet-repartitioned-my-suffix/part-00003-filename.snappy.parquet"
         """
         original_relative_path_obj = Path(original_relative_path)
-        # We expect files to have a path of the form /path/to/parquet/part-00001.snappy.parquet
-        assert original_relative_path_obj.parts[-2] == "parquet"  # TODO: Remove this assumption?
```
jalencato, line 198: Do we meet any issue for it?

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -622,6 +628,121 @@ def csv_row(data: Row):
...
+            rows_this_file = rows_per_file + (1 if file_idx < remainder else 0)
+            end_idx = start_idx + rows_this_file
+
+            # Slice the table
+            file_table = pa_table.slice(start_idx, rows_this_file)
+
+            # Generate Spark-style part filename
+            part_filename = os.path.join(
+                base_path,
+                f"part-{file_idx:05d}-{unique_id}.parquet",
```
jalencato, line 698: So unlike pyspark, pyarrow does not support the built-in naming?

jalencato (approved):
Overall LGTM. Let us keep an eye on further experiment.

graphstorm-processing/tests/test_dist_heterogenous_loader.py
```diff
@@ -515,6 +521,46 @@ def create_edges_df_num_label(spark: SparkSession, missing_data_points: int) ->
...
+        NODE_MAPPING_STR: node_str_ids, unique per node (uuid4),
+        NODE_MAPPING_INT: node_int_ids, unique per node,
+        NUM_LABEL_COL: numeric labels, 0-9 range,
+        STR_LABEL_COL: string labels, 10 random string labels,
+    """
+    node_str_ids = [str(uuid4()) for _ in range(total_data_points)]
+    node_int_ids = np.arange(total_data_points)
+
+    rng = np.random.default_rng(42)
+    # Create random numerical labels with values 0-9
```
jalencato, line 540: Do we cover the case for None value?

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: Yes, we could try using pyarrow.ds.dataset to do the writing to provide that but that actually complicates things. This allows us to have explicit distribution of rows.

graphstorm-processing/tests/test_dist_heterogenous_loader.py
thvasilo: Yes, lines 547-549 set `missing_data_points` number of labels to `None`.

graphstorm-processing/graphstorm_processing/repartition_files.py
thvasilo: This was triggered when I was using a "non standard" path to write files through pyarrow (without `parquet` in the path). I think we can remove it because it just verifies an assumption that this piece of code cannot control, it's hard for a new dev to understand why this should be here.

PR #1140, opened 2025-01-18
[GSProcessing] Fix ParquetRowCounter bug when different types had same-name features

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Previously, if some node/edge types shared the same feature names, we would end up overwriting the original feature name dict because we shared it between types. 
* Now we create copies of the "data" dict for each type, and write the row counts directly to the corresponding type's dictionary, under the "row_counts" key.
* We also make the same fix for the graph structure entries (edges), although it's unlikely this would happen there.
* Finally used GenAI to generate some test cases for the file because it was untested before.

### DIFF
--- a/graphstorm-processing/graphstorm_processing/graph_loaders/row_count_utils.py
+++ b/graphstorm-processing/graphstorm_processing/graph_loaders/row_count_utils.py
@@ -184,8 +184,13 @@ class ParquetRowCounter:
         all_entries_counts = []  # type: List[Sequence[int]]
         for type_value in self.metadata_dict[edge_or_node_type_key]:
             logging.info("Getting counts for %s, %s", top_level_key, type_value)
-            relative_file_list = self.metadata_dict[top_level_key][type_value]["data"]
+            # Get the data dictionary for this type and create a copy
+            # to avoid modifying shared references
+            type_data_dict = self.metadata_dict[top_level_key][type_value].copy()
+            relative_file_list = type_data_dict["data"]
             type_row_counts = self.get_row_counts_for_parquet_files(relative_file_list)
+
+            # Store the row counts directly in the metadata dictionary
             self.metadata_dict[top_level_key][type_value]["row_counts"] = type_row_counts
             all_entries_counts.append(type_row_counts)
 
@@ -229,10 +234,15 @@ class ParquetRowCounter:
                     self.metadata_dict[top_level_key].keys(),
                 )
                 continue
-            for feature_name, feature_data_dict in self.metadata_dict[top_level_key][
-                type_name
-            ].items():
-                relative_file_list = feature_data_dict["data"]  # type: Sequence[str]
+
+            # Create a new dictionary for this type's features
+            type_features = self.metadata_dict[top_level_key][type_name]
+
+            for feature_name, feature_data in type_features.items():
+                # Create a copy of the feature data to avoid modifying shared references
+                feature_data_dict = feature_data.copy()
+                relative_file_list = feature_data_dict["data"]
+
                 logging.info(
                     "Getting counts for %s, type: %s, feature: %s",
                     top_level_key,
@@ -247,7 +257,12 @@ class ParquetRowCounter:
                     feature_name,
                     feature_row_counts,
                 )
+                # Store the row counts in both places
                 feature_data_dict["row_counts"] = feature_row_counts
+                self.metadata_dict[top_level_key][type_name][feature_name][
+                    "row_counts"
+                ] = feature_row_counts
+
                 features_per_type_counts.append(feature_row_counts)
             all_feature_counts.append(features_per_type_counts)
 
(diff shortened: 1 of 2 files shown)


graphstorm-processing/graphstorm_processing/graph_loaders/row_count_utils.py
```diff
@@ -184,8 +184,13 @@ def _add_counts_for_graph_structure(
         all_entries_counts = []  # type: List[Sequence[int]]
         for type_value in self.metadata_dict[edge_or_node_type_key]:
             logging.info("Getting counts for %s, %s", top_level_key, type_value)
-            relative_file_list = self.metadata_dict[top_level_key][type_value]["data"]
+            # Get the data dictionary for this type and create a copy
+            # to avoid modifying shared references
+            type_data_dict = self.metadata_dict[top_level_key][type_value].copy()
```
jalencato, line 189: What is the estimated size for the ``type_data_dict`` here? I suppose it should be meta information so relatively small.

graphstorm-processing/graphstorm_processing/graph_loaders/row_count_utils.py
```diff
@@ -229,10 +234,15 @@ def _add_counts_for_features(self, top_level_key: str, edge_or_node_type_key: st
...
-                type_name
-            ].items():
-                relative_file_list = feature_data_dict["data"]  # type: Sequence[str]
+
+            # Create a new dictionary for this type's features
+            type_features = self.metadata_dict[top_level_key][type_name]
+
+            for feature_name, feature_data in type_features.items():
+                # Create a copy of the feature data to avoid modifying shared references
+                feature_data_dict = feature_data.copy()
```
jalencato, line 243: Still metadata here?

graphstorm-processing/tests/test_row_count_utils.py
```diff
@@ -0,0 +1,294 @@
...
+                    "data": [
+                        "node_data/type2-feature1/part-00000.parquet",
+                        "node_data/type2-feature1/part-00001.parquet",
+                    ],
+                }
+            },
+        },
+        "edge_data": {
+            "type1:edge:type2": {
+                "weight": {
```
jalencato, line 78: What is weight here?

graphstorm-processing/tests/test_row_count_utils.py
```diff
@@ -0,0 +1,294 @@
...
+            },
+            "type2": {
+                "format": {"name": "parquet"},
+                "data": [
+                    "raw_id_mappings/type2/part-00000.parquet",
+                    "raw_id_mappings/type2/part-00001.parquet",
+                ],
+            },
+        },
+        "edges": {
```
jalencato, line 103: I see we do not test for the edge data, though it is unlikely, shall we also try to test it?

graphstorm-processing/graphstorm_processing/graph_loaders/row_count_utils.py
```diff
@@ -405,6 +420,14 @@ def verify_metadata_match(graph_meta: Dict[str, Dict]) -> bool:
         True if all row counts match for each type, False otherwise.
     """
     logging.info("Verifying features and structure row counts match...")
+
+    # Check if required keys exist
+    required_keys = ["edge_data", "edges", "node_data"]
```
jalencato, line 425: Why there are both edges and edge_data here?

jalencato (commented):
A few questions here:

* What shall we expect the row count here? The right count same to the original row count?
* Previously all node/edge types shared the same row count dict including all feature's information, but now it only stored the ones relative to itself. Will it bring any changes?

graphstorm-processing/graphstorm_processing/graph_loaders/row_count_utils.py
thvasilo: Yeah up to a a couple of thousand strings, so a few KB-MB.

graphstorm-processing/graphstorm_processing/graph_loaders/row_count_utils.py
thvasilo: Yes, feature_data refers to to the `data` entry in a feature, a dict with two or three entries, one string for `format`, one list of str for `files` and optionally one string for separator.

graphstorm-processing/graphstorm_processing/graph_loaders/row_count_utils.py
thvasilo: Edges refers to the edges that describe the graph structure (src,dst) while edge_data are features, labels etc. for each edge type.

This is according to the DGL chunked data specification: https://docs.dgl.ai/guide/distributed-preprocessing.html#chunked-graph-format

The linked spec is not 100% up to date however and should be fixed.

graphstorm-processing/tests/test_row_count_utils.py
thvasilo: Yeah good point, I'll add a test for edge_data as well.

thvasilo:
For the high level questions @jalencato 

> What shall we expect the row count here? The right count same to the original row count?

The row counter is used right after gs-processing has finished, and it counts the rows of each generated file for every node feature/mask/label, for each edge structure (src,dst) file, and for each edge feature/mask label.

gs-repartition uses these values as input to determine if re-partitioning is needed for any of the sets of files, as DGL assumes that all row counts for every node/edge type are the same.

In this case, what was happening was that some features that shared a name ended up without _any_ row counts, because their key kept getting overwritten. Only the first-encountered key got to have row_counts, the rest had nothing. The "right" count is whatever that node's feature file counts actually were, before it was missing. 


> Previously all node/edge types shared the same row count dict including all feature's information, but now it only stored the ones relative to itself. Will it bring any changes?

In terms of expected input/output there won't be any changes, this just fixes the case where same-name features were not getting their row counts populated correctly.

jalencato (approved):
LGTM

PR #1144, opened 2025-01-23
Support dumping training logs for TensorBoard visualization toolkit.

classicsong:
*Issue #, if available:*
#988 

*Description of changes:*
Add a TensorBoard tracker that will save training loss, validation scores and test scores into TensorBoard logs.


docs/source/cli/model-training-inference/configuration-run.rst
```diff
@@ -161,10 +161,10 @@ GraphStorm provides a set of parameters to control how and where to save and res
...
+- **task_tracker**: A task tracker used to formalize and report model performance metrics. Now GraphStorm supports ``sagemaker_task_tracker`` and ``tensorboard_task_tracker``. ``sagemaker_task_tracker`` prints evaluation metrics in a formatted way so that a user can capture those metrics through SageMaker. (See Monitor and Analyze Training Jobs Using Amazon CloudWatch Metrics  for more details.) ``tensorboard_task_tracker`` dumps evaluation metrics in a formatted way that can be loaded by TensorBoard. The default path for storing the TensorBoard logs is ``./runs/`` under **workspace**. Users can define their own TensorBoard log directory by setting **task_tracker** as ``tensorboard_task_tracker:LOG_PATH``, where ``LOG_PATH`` will be the TensorBoard log directory. (Note: to use ``tensorboard_task_tracker``, one should install the tensorboard Python package.)
```
jalencato, line 164: May we add external links here for sagemaker_task_tracker and tensorboard_task_tracker?

python/graphstorm/config/argument.py
```diff
@@ -3237,7 +3287,7 @@ def _add_output_args(parser):
 def _add_task_tracker(parser):
     group = parser.add_argument_group(title="task_tracker")
     group.add_argument("--task-tracker", type=str, default=argparse.SUPPRESS,
-            help=f'Task tracker name. Now we only support {GRAPHSTORM_SAGEMAKER_TASK_TRACKER}')
+            help=f'Task tracker name. Now we  support {SUPPORTED_TASK_TRACKER}')
```
jalencato, line 3290:
```suggestion
            help=f'Task tracker name. Now we support {SUPPORTED_TASK_TRACKER}')
```

docs/source/cli/model-training-inference/configuration-run.rst
classicsong: These two are implemented in GraphStorm.

thvasilo:
To support optional dependencies  we do the following @classicsong 

Pandas has their own `pandas.compat._optional.import_optional_dependency` We can use a similar mechanism. 
See https://pandas.pydata.org/pandas-docs/version/1.4/development/contributing_codebase.html#optional-dependencies and https://pandas.pydata.org/pandas-docs/version/1.4/getting_started/install.html#install-optional-dependencies

Optional imports are done inline, using something like 

```python
def generate_numba_apply_func(
    func, nogil=True, nopython=True, parallel=False
) -> Callable[[npt.NDArray, Index, Index], dict[int, Any]]:
    numba = import_optional_dependency("numba")
    jitted_udf = numba.extending.register_jitable(func)
```

So in this case we could do


```python

class GSTensorBoardTracker(GSSageMakerTaskTracker):
    def __init__(self, log_report_frequency, log_dir=None):
        super().__init__(log_report_frequency, log_dir)
        try:
            tensorboard = importlib.import_module("torch.utils.tensorboard")
        except ImportError as err:
            msg =  (
                "GSTensorBoardTracker requires tensorboard to run. "
                "Please install the tensorboard Python package.")
            raise ImportError(msg) from err
        
        self._writer = tensorboard.SummaryWriter(log_dir)
```

This ensures we only try to pull the dependency if we try to instantiate a `GSTensorBoardTracker` object and not
at module import time.

Then we can modify our setup.py to include tensorboard as an extra dependency:


```python
setup(
    # Metadata
    name='graphstorm',
    version=VERSION,
    python_requires='>=3.8',
    description='GraphStorm',
    long_description_content_type='text/markdown',
    license='Apache-2.0',

    # Package info
    packages=find_packages(where="python", exclude=(
        'tests',
    )),
    package_dir={"": "python"},
    package_data={'': [os.path.join('datasets', 'dataset_checksums', '*.txt')]},
    zip_safe=True,
    include_package_data=True,
    install_requires=requirements,
    ext_modules=extensions,
    cmdclass=cmdclass,
    
    extras_require={
        'visualization': [
            'tensorboard', # Add minimum version if needed, e.g. 'tensorboard>=2.11.2'
        ]
    },
)
```

then when users want tensorboard visualization they can install `pip install graphstorm[visualization]` to pull in the extra tensorboard dependency.

python/graphstorm/tracker/tensorboard_tracker.py
```diff
@@ -0,0 +1,194 @@
...
+        force_report: bool
+            If true, report the metric
+        """
+        if force_report or self._do_report(step):
+            if metric_value is not None:
+                if isinstance(metric_value, str):
+                    # Only rank 0 will write log to TensorBoard
+                    if get_rank() == 0:
+                        self._writer.add_scalar(metric_name, metric_value, step)
+                    logging.info("Step %d | %s: %s", step, metric_name, metric_value)
```
thvasilo, line 82: If `metric_value` is a str should we be using `_writer.add_scalar` ? I'm not very familiar with the tracker interface.

python/graphstorm/tracker/tensorboard_tracker.py
```diff
@@ -0,0 +1,194 @@
...
+                elif isinstance(metric_value, numbers.Number):
+                    # Only rank 0 will write log to TensorBoard
+                    if get_rank() == 0:
+                        self._writer.add_scalar(metric_name, metric_value, step)
+                    logging.info("Step %d | %s: %.4f", step, metric_name, metric_value)
+                else:
+                    # Only rank 0 will write log to TensorBoard
+                    if get_rank() == 0:
+                        self._writer.add_text(metric_name, str(metric_value), step)
+                    logging.info("Step %d | %s: %s", step, metric_name, str(metric_value))
```
thvasilo, line 92: What other types apart from number and str can `metric_value` be?

python/graphstorm/tracker/tensorboard_tracker.py
```diff
@@ -0,0 +1,194 @@
...
+                Train metric name
+            metric_value:
+                Train metric value
+            step: int
+                The corresponding step/iteration in the training loop.
+            force_report: bool
+                If true, report the metric
+        """
+        metric_name = f"{metric_name}/Train"
+        self.log_metric(metric_name, metric_value, step, force_report)
```
thvasilo, line 109: For all the variations of this method could we pass the context as a var and have a single function definition? Otherwise we're repeating ourselves.

```
def log_metric(self, context, metric_name, metric_value, step, force_report=False):
        """ Log train metric
            Parameters
            ----------
            context: str
                Context from which we report the metric, e.g. "Train".
            metric_name: str
                Train metric name
            metric_value:
                Train metric value
            step: int
                The corresponding step/iteration in the training loop.
            force_report: bool
                If true, report the metric
        """
        metric_name = f"{metric_name}/{context}"
        self.log_metric(metric_name, metric_value, step, force_report)
```

caller does GSTensorBoardTracker.log_metric("Train", "accuracy", 0.5, 5)

tests/unit-tests/test_task_tracker.py
```diff
@@ -0,0 +1,62 @@
...
+    setattr(config, "_task_tracker", f"{GRAPHSTORM_TENSORBOARD_TASK_TRACKER}:log")
+    tracker = create_builtin_task_tracker(config)
+    assert isinstance(tracker, GSTensorBoardTracker)
+    # check tensorboard writer
+    assert tracker._writer is not None
+    assert tracker._writer.log_dir.startswith("log")
+
+if __name__ == '__main__':
+    test_get_tracker_class()
+    test_create_builtin_task_tracker()
```
thvasilo, line 62: Missing newline at EOF

python/graphstorm/tracker/tensorboard_tracker.py
classicsong: Good catch, we should use add_text

python/graphstorm/tracker/tensorboard_tracker.py
classicsong: If it is a class, dict, etc. Though, we are not expecting users will pass these obj to log_metric.

python/graphstorm/tracker/tensorboard_tracker.py
classicsong: We can.
But I don't think this makes big difference.
Just keep it as the SageMakerTracker now.

docs/source/cli/model-training-inference/configuration-run.rst
```diff
@@ -161,10 +161,10 @@ GraphStorm provides a set of parameters to control how and where to save and res
...
+- **task_tracker**: A task tracker used to formalize and report model performance metrics. Now GraphStorm supports two task trackers: ``sagemaker_task_tracker`` and ``tensorboard_task_tracker``. ``sagemaker_task_tracker`` prints evaluation metrics in a formatted way so that a user can capture those metrics through SageMaker. (See Monitor and Analyze Training Jobs Using Amazon CloudWatch Metrics  for more details.) ``tensorboard_task_tracker`` dumps evaluation metrics in a formatted way that can be loaded by TensorBoard. The default path for storing the TensorBoard logs is ``./runs/`` under **workspace**. Users can define their own TensorBoard log directory by setting **task_tracker** as ``tensorboard_task_tracker:LOG_PATH``, where ``LOG_PATH`` will be the TensorBoard log directory. (Note: to use ``tensorboard_task_tracker``, one should install the tensorboard Python package.)
```
thvasilo, line 164:
```suggestion
- **task_tracker**: A task tracker used to formalize and report model performance metrics. Now GraphStorm supports two task trackers: ``sagemaker_task_tracker`` and ``tensorboard_task_tracker``. ``sagemaker_task_tracker`` prints evaluation metrics in a formatted way so that a user can capture those metrics through SageMaker. (See Monitor and Analyze Training Jobs Using Amazon CloudWatch Metrics  for more details.) ``tensorboard_task_tracker`` dumps evaluation metrics in a formatted way that can be loaded by TensorBoard. The default path for storing the TensorBoard logs is ``./runs/`` under **workspace**. Users can define their own TensorBoard log directory by setting **task_tracker** as ``tensorboard_task_tracker:LOG_PATH``, where ``LOG_PATH`` will be the TensorBoard log directory. (Note: to use ``tensorboard_task_tracker``, one should install the tensorboard Python package using ``pip install tensorboard`` or during graphstorm installation using ``pip install graphstorm[tensorboard]``.)
```

thvasilo (approved):
LGTM one minor comment for docs.

PR #1145, opened 2025-01-23
[Docker] Add vim for local docker image

jalencato:
*Issue #, if available:*

*Description of changes:*

Improve developer experience for adding vim back to local docker image.

### DIFF
--- a/docker/local/Dockerfile.local
+++ b/docker/local/Dockerfile.local
@@ -17,6 +17,7 @@ RUN apt update && apt install -y --no-install-recommends \
     python3-distutils \
     python3-venv \
     unzip \
+    vim \
     && rm -rf /var/lib/apt/lists/*
 
 # Install aws-cli


classicsong:
Why we need vim?

jalencato:
> Why we need vim?

For developer friendly to modify the config insider a docker container.

thvasilo:
I would be against adding dependencies to the prod images purely for dev purposes, because they make the images heavier. 

We could add an additional target called `dev` that installs dev-related deps, similar to how GSProcessing has `prod` and `test` targets: https://github.com/awslabs/graphstorm/blob/5f7abb1bd2532f37a08fe4f39e920350fe602a70/graphstorm-processing/docker/0.4.0/sagemaker/Dockerfile.cpu#L48-L54

jalencato:
> I would be against adding dependencies to the prod images purely for dev purposes, because they make the images heavier.
> 
> We could add an additional target called `dev` that installs dev-related deps, similar to how GSProcessing has `prod` and `test` targets:
> 
> https://github.com/awslabs/graphstorm/blob/5f7abb1bd2532f37a08fe4f39e920350fe602a70/graphstorm-processing/docker/0.4.0/sagemaker/Dockerfile.cpu#L48-L54
> 
> Our default target is `prod` and the build script supports changing the target to `test` to install the additional dependencies

Make sense. I will close this PR. As I agree it is not necessary to make the dev images too complicated.

PR #1150, opened 2025-01-29
[GSProcessing] Make label transformation use Spark for output instead of Pyarrow when using custom masks.

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Previously, during classification labels would be collected to the driver and written using pyarrow regardless of the type of mask that will be generated.
* To improve scaling, since we observed that the collect is not necessary when using custom masks, we revert to writing the label files using Spark, avoiding collecting to the driver and improving scalability.

### DIFF
--- a/graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
+++ b/graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
@@ -1342,7 +1342,10 @@ class DistHeterogeneousGraphLoader(object):
                 order_col = None
 
             self.graph_info["is_multilabel"] = label_conf.multilabel
+
+            # Create loader object. If doing classification  order_col!=None will enforce re-order
             node_label_loader = DistLabelLoader(label_conf, self.spark, order_col)
+
             logging.info(
                 "Processing label data for node type %s, label col: %s...",
                 node_type,
@@ -1350,6 +1353,7 @@ class DistHeterogeneousGraphLoader(object):
             )
 
             transformed_label = node_label_loader.process_label(nodes_df)
+
             self.graph_info["label_map"] = node_label_loader.label_map
 
             label_output_path = (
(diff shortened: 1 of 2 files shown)


graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -1342,14 +1342,18 @@ def _process_node_labels(
                 order_col = None
 
             self.graph_info["is_multilabel"] = label_conf.multilabel
+
+            # Create loader object. If doing classification  order_col!=None will enforce re-order
```
jalencato, line 1346:
```suggestion
            # Create loader object. If doing classification order_col!=None will enforce re-order
```

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -1361,37 +1365,45 @@ def _process_node_labels(
...
+                        transformed_label.select(label_conf.label_column), label_output_path
+                    )
+                else:
+                    input_num_parts = nodes_df.rdd.getNumPartitions()
+                    # If num parts is different for original and transformed, log a warning
+                    transformed_num_parts = transformed_label.rdd.getNumPartitions()
+                    if input_num_parts != transformed_num_parts:
+                        logging.warning(
+                            "Number of partitions for original (%d) and transformed label data "
+                            "(%d) differ.  This may cause issues with the label split files.",
```
jalencato, line 1381: The issue here should be disorder?

jalencato (approved):
LGTM with minor comments.

graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
thvasilo: Yes, usually this is an indication some sort of shuffling took place to repartition the data

PR #1165, opened 2025-02-13
Validate the numerical values of input features in GConstruct.

classicsong:
*Issue #, if available:*
#1166 

*Description of changes:*
Validate the numerical values during data transformation when debug mode is True.
(GConstruct has already handled the None value in CategoricalTransform and TokenTransform.


python/graphstorm/gconstruct/transform.py
```diff
@@ -280,6 +284,32 @@ def as_out_dtype(self, feats):
...
+        ----------
+        feats: np.array
+            Input feature.
+
+        Return
+        ------
+        np.array: the cast feature.
+        """
+        if not (np.issubdtype(feats.dtype, np.integer) \
+            or np.issubdtype(feats.dtype, np.floating)):
```
thvasilo, line 301: Maybe we can use https://numpy.org/devdocs/reference/arrays.scalars.html#numpy.number

python/graphstorm/gconstruct/transform.py
```diff
@@ -280,6 +284,32 @@ def as_out_dtype(self, feats):
...
+            logging.warning("The feature %s has to be floating points or integers,"
+                            "but get %s. Try to cast it into float32",
+                            self.feat_name, feats.dtype)
+            try:
+                # if input dtype is not float or integer, we need to cast the data
+                # into float32
+                feats = feats.astype(np.float32)
+            except: # pylint: disable=bare-except
+                raise ValueError(f"The feature {self.feat_name} has to be integers or floats.")
```
thvasilo, line 310: And the say has to be a number

thvasilo (approved):
LGTM two minor comments

python/graphstorm/gconstruct/transform.py
classicsong: I used number before, but it has some problems, so I chose to use np.integer and np.floating.

PR #1170, opened 2025-02-14
Handle the cases when the input edge file(s) are empty

classicsong:
*Issue #, if available:*
#996 

*Description of changes:*
GConstruct will handle the corner cases when some edge files are empty files.

Note: DGL graph partition forces every edge type should have at least one edges.


jalencato (approved):
LGTM

PR #1174, opened 2025-02-17
[Tests] HGTLayerwithEdgeFeat tests

RonaldBXu:
*Issue #, if available:*

*Description of changes:*
Adding tests for HGTLayerwithEdgeFeat module.

### DIFF
--- a/tests/unit-tests/test_nn_model.py
+++ b/tests/unit-tests/test_nn_model.py
@@ -706,6 +706,116 @@ def test_hgt_with_edge_features(input_dim, output_dim, dev):
                         np.ones([1, output_dim]) * target_val, decimal=5)
 
     # TODO: add cases when edge feature inputs are all 0s.
+    edge_feats_zeros = {
+        ("n0", "r0", "n1"): th.zeros(r0_eid.shape[0], input_dim).to(dev),
+        ("n0", "r1", "n1"): th.zeros(r1_eid.shape[0], input_dim).to(dev)
+    }
+
+    layerwithef = HGTLayerwithEdgeFeat(
+        input_dim, output_dim, 
+        ntypes, etypes,
+        num_heads=1,
+        edge_feat_name={("n0", "r0", "n1"): ['feat'], ("n0", "r1", "n1"): ['feat']},
+        edge_feat_mp_op='add',
+        activation=th.nn.ReLU(),
+        dropout=0.0,
+        norm='')
+    init_hgtlayer(layerwithef)
+    layerwithef = layerwithef.to(dev)
+    layerwithef.eval()
+    add_emb = layerwithef(block, node_feats, edge_feats_zeros)
+    # With all node features as 1s, edge feature inputs as 0s, and all weights are 1s in the shape of (in_dim, out_dim), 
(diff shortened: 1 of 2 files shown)


tests/unit-tests/test_nn_model.py
```diff
@@ -932,11 +1042,105 @@ def test_hgt_with_edge_features(input_dim, output_dim, dev):
...
+        edge_feats = {
+            ("n0", "r1", "n1"): th.rand(r1_eid.shape[0], input_dim).to(dev)
+        }
+
+        emb = layerwithef(block, node_feats, edge_feats)
+        
+        assert 'n0' not in emb
+        assert emb['n1'].shape[0] == len(seeds['n1'])
+        assert emb['n1'].shape[1] == output_dim
+
```
zhjwy9343, line 1100: Could you add aother normal cases, where initialize `HGTLayerwithEdgeFeat` with both two edge types, but only provide one type of edge feature in forward(), which should also work.

tests/unit-tests/test_nn_model.py
```diff
@@ -932,11 +1042,105 @@ def test_hgt_with_edge_features(input_dim, output_dim, dev):
...
+            edge_feat_mp_op=e_f_mp_op,
+            activation=th.nn.ReLU(),
+            dropout=0.0,
+            norm='')
+        init_hgtlayer(layerwithef)
+        layerwithef = layerwithef.to(dev)
+        layerwithef.eval()
+
+        with assert_raises(AssertionError):
+            layerwithef(block, node_feats)
```
zhjwy9343, line 1143: Could you add another abnormal case, where initialize `HGTLayerwithEdgeFeat` with edge feature name, but provide an empty dict as edge feature in forward(). This will trigger an assertion error of no edge feature provided for MP.

tests/unit-tests/test_trainer.py
```diff
@@ -1177,6 +1177,107 @@ def test_hgt_nc4ef():
...
+
+        trainer5.setup_evaluator(evaluator5)
+
+        with assert_raises(AssertionError):
+            trainer5.fit(
+                train_loader=train_dataloader5,
+                val_loader=val_dataloader5,
+                num_epochs=2
+                )
+
```
zhjwy9343, line 1280: Could you add another abnormal case, where set HGT model without edge features for NC, but provide edge features.

tests/unit-tests/test_trainer.py
```diff
@@ -1546,6 +1647,111 @@ def test_hgt_ec4ef():
...
```
zhjwy9343, line 1754: ditto

tests/unit-tests/test_trainer.py
```diff
@@ -1923,6 +2129,117 @@ def test_hgt_lp4ef():
...
```
zhjwy9343, line 2242: ditto

tests/unit-tests/test_trainer.py
zhjwy9343: I see there is already such test case. So ignore this comment.

PR #1178, opened 2025-02-18
Fix overlapping IDs being written when ID map writes to multiple files

### LINKED ISSUE #1177: [Bug] Duplicated id mappings using IdMap.save() for partitioned files
We detected duplicated IDs inside the raw_id_mappings inside partitioned parquet files. It happened only when multiple files are written. Like, 
```
part-00000.parquet
part-00001.parquet
part-00002.parquet
part-00003.parquet
```
The slices created via pyarrow slice, is using an offset + length logic. This leads to increasing slice sizes when using a start-end logic. https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.slice

This happened in the `id_map.py` file line ##239 where `pq.write_table(...)`
```
        while rows_written < table.num_rows:
            start = rows_written
            end = min(rows_written + max_rows_per_file, table.num_rows)
            filename = f"part-{str(file_idx).zfill(5)}.parquet"
            pq.write_table(table.slice(start, end), os.path.join(file_prefix, filename))
            rows_written = end
            file_idx += 1
```

The bug leads to duplications of mappings and increasing file sizes. Example with start-end logic in Graphstorm using max_rows_per_file= 500:
```
start = 0, end=500 -> Actual sliced Rows 0 – 500
start = 500, end = 1000 -> Actual sliced  Rows 500 – 1500  -> contains parts of next slice already
start = 1000 , end = 1500  ->  Actual sliced Rows 1000 - 2500
 ….
```

**Root Cause:**
PyArrow's `slice()` method uses `offset` and `length`, instead of `start` and `end`.

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Fixes writing behavior and adds corresponding unit test
* GenAI used to extend new unit test

Fixes #1177 

### DIFF
--- a/tests/unit-tests/gconstruct/test_remap_result.py
+++ b/tests/unit-tests/gconstruct/test_remap_result.py
@@ -15,11 +15,13 @@
 """
 import glob
 import os
+import math
 import tempfile
 from functools import partial
 from pathlib import Path
 
 import pytest
+import pyarrow as pa
 import pandas as pd
 import torch as th
 import numpy as np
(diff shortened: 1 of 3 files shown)


tests/unit-tests/gconstruct/test_remap_result.py
```diff
@@ -535,11 +537,57 @@ def test_parse_config():
...
+        # string ids will be sequential id_<idx>
+        ids = [f"id_{i}" for i in range(num_rows)]
+
+        # Create the IdMap
+        id_map = IdMap(np.array(ids))
+
+        # Save the id map - will create multiple partitions due to small GIB_BYTES
+        mapping_prefix = os.path.join(tmpdirname, "raw_id_mappings")
+        id_map.save(mapping_prefix)
+
```
zhjwy9343, line 556: Do we know how many files exactly will be saved? If so, we might check the number of files here.

tests/unit-tests/gconstruct/test_remap_result.py
```diff
@@ -535,11 +537,57 @@ def test_parse_config():
...
+            table = pq.read_table(filename)
+            partition_rows = table.to_pydict()
+            all_rows.extend(zip(
+                partition_rows[MAPPING_INPUT_ID],
+                partition_rows[MAPPING_OUTPUT_ID])
+            )
+            file_idx += 1
+
+        # There should be multiple partition files due to small GIB_BYTES
+        assert file_idx > 1, "Expected multiple partition files"
```
zhjwy9343, line 571: Can we have a specific number of files here?

tests/unit-tests/gconstruct/test_remap_result.py
thvasilo: We can get an approximate number if we determine the byte size of each row or the total byte size of the mapping table in memory

zhjwy9343 (approved):
LGTM

PR #1192, opened 2025-02-27
[GSProcessing] Enforce re-order for edge label processing during classification

jalencato:
*Issue #, if available:*

*Description of changes:*


graphstorm-processing/graphstorm_processing/graph_loaders/dist_heterogeneous_loader.py
```diff
@@ -1571,6 +1572,18 @@ def write_edge_structure(
...
+        edge_df_with_int_ids = edge_df_with_int_ids.rdd.zipWithIndex()
+
+        new_schema = original_schema.add(StructField(EDGE_MAPPING_INT, LongType(), False))
+
+        edge_rdd_with_order_ids = edge_df_with_int_ids.map(
+            lambda rdd_row: (list(rdd_row[0]) + [rdd_row[1]])  # type: ignore
+        )
+
+        edge_df_with_int_ids = self.spark.createDataFrame(edge_rdd_with_order_ids, new_schema)
+
```
thvasilo, line 1586: Have we tested just using e.g. the `dst` column for order? That's what DGL does IIRC

jalencato: I can try it.

jalencato: Test with only the `dst` column, in this way the order of the whole dataframe will be undetermined. Let us move with the int id for now.

thvasilo: What do we mean undetermined? What's happening when sorting by `dst` vs. creating a new column?

jalencato: I mean instead of using newly created `EDGE_MAPPING_INT` column, I use `dst` here for the order_col. When doing the `transformed_label = transformed_label.sort(self.order_col).cache()` to maintain the order, then the final output order will be totally undetermined. So I suggest we keep using a new column.

thvasilo (approved):
LGTM now, a couple minor comments.

graphstorm-processing/tests/test_dist_heterogenous_loader.py
```diff
@@ -1483,3 +1483,130 @@ def read_pandas_from_relative_list(file_list) -> pd.DataFrame:
...
+    spark: SparkSession,
+    dghl_loader: DistHeterogeneousGraphLoader,
+):
+    """Test that label and mask order is maintained after label processing.
+    NOTE: Tests DGHL code together with DistLabelLoader code because their results are coupled.
+    """
+    label_col = STR_LABEL_COL
+
+    # Create a Pandas DF with a label column with 10k "zero", 10k "one", 10k None rows
+    num_datapoints = 10**5
```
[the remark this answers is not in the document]
jalencato: Assume edge will be much larger, so I increase the number here a little bit.

PR #1195, opened 2025-02-27
[Feature] Adding logging for early stopping

RonaldBXu:
*Issue #, if available:*

*Description of changes:*
Adding logging for early stopping.

### DIFF
--- a/python/graphstorm/eval/evaluator.py
+++ b/python/graphstorm/eval/evaluator.py
@@ -30,6 +30,7 @@ from ..config.config import (EARLY_STOP_AVERAGE_INCREASE_STRATEGY,
                              EARLY_STOP_CONSECUTIVE_INCREASE_STRATEGY,
                              LINK_PREDICTION_MAJOR_EVAL_ETYPE_ALL)
 from ..utils import get_rank, get_world_size, barrier
+import logging
 
 
 def early_stop_avg_increase_judge(val_score, val_perf_list, comparator):
@@ -381,6 +382,9 @@ class GSgnnBaseEvaluator():
 
         self._val_perf_list.pop(0)
         self._val_perf_list.append(val_score)
+
+        if early_stop and get_rank() == 0:
+            logging.info("Early stop criterion %s satisfied. Training job stopping.", self._early_stop_strategy)
 
         return early_stop
 


python/graphstorm/eval/evaluator.py
```diff
@@ -382,6 +383,9 @@ def do_early_stop(self, val_score):
         self._val_perf_list.pop(0)
         self._val_perf_list.append(val_score)
 
+        if early_stop and get_rank() == 0:
+            logging.info("Early stop criterion %s satisfied. Training job stopping.", self._early_stop_strategy)
```
thvasilo, line 387: Should we log the val_score  and val_perf_list @zhjwy9343 ?

zhjwy9343: Yes, need to create different early stop messages for different criterion, and output the `val_perf_list` for reference.

thvasilo: @RonaldBXu see above

thvasilo (approved):
LGTM

python/graphstorm/eval/evaluator.py
```diff
@@ -379,9 +380,32 @@ def do_early_stop(self, val_score):
...
+            elif self._early_stop_strategy == EARLY_STOP_CONSECUTIVE_INCREASE_STRATEGY:
+                val_perf_list = self._val_perf_list.copy()
+
         self._val_perf_list.pop(0)
         self._val_perf_list.append(val_score)
 
+        if early_stop and get_rank() == 0:
+            if self._early_stop_strategy == EARLY_STOP_AVERAGE_INCREASE_STRATEGY:
+                logging.info(
+                    "Early stopping: Current validation score %.3f is lower than "
```
zhjwy9343, line 395: Because the evaluation of val values depends on specific metrics, some time we hope to stop earlier if the current val value does not increase, e.g., accuracy and auc, while some time if the val values does not decrease, e.g., mae and rmse. So here we might want to use `validation score %.3f is worse than ...`, rather than saying `lower`, `higher`, `increasing` or `decreasing`.

python/graphstorm/eval/evaluator.py
```diff
@@ -379,9 +380,32 @@ def do_early_stop(self, val_score):
...
+                    "Early stopping: Current validation score %.3f is worse than "
+                    "or the same as the average of the last %d validation scores %.3f.",
+                    val_score,
+                    self._early_stop_rounds,
+                    avg_val_perf
+                )
+            elif self._early_stop_strategy == EARLY_STOP_CONSECUTIVE_INCREASE_STRATEGY:
+                logging.info(
+                    "Early stopping: For the last %d consecutive steps, "
+                    "the validation scores %s are worse.",
```
zhjwy9343, line 404:
```suggestion
                    "the validation scores %s are not improving. Stop training.",
```

python/graphstorm/eval/evaluator.py
```diff
@@ -379,9 +380,32 @@ def do_early_stop(self, val_score):
...
+                val_perf_list = self._val_perf_list.copy()
+
         self._val_perf_list.pop(0)
         self._val_perf_list.append(val_score)
 
+        if early_stop and get_rank() == 0:
+            if self._early_stop_strategy == EARLY_STOP_AVERAGE_INCREASE_STRATEGY:
+                logging.info(
+                    "Early stopping: Current validation score %.3f is worse than "
+                    "or the same as the average of the last %d validation scores %.3f.",
```
zhjwy9343, line 396:
```suggestion
                    "or the same as the average of the last %d validation scores %.3f. Stop training.",
```

zhjwy9343 (approved):
LGTM

PR #1196, opened 2025-02-28
[Fix] Fix an HGT Encoder Corner Case: 0 edge in a block

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
This PR fixes an HGT Encoder corner case, where the edge type with feature has 0 edge in a block. In such case the input of edge feature become empty dict, and will trigger an assertion error in the previous implementation. This PR handles this corner case.

### DIFF
--- a/python/graphstorm/model/hgt_encoder.py
+++ b/python/graphstorm/model/hgt_encoder.py
@@ -716,7 +716,6 @@ class HGTLayerwithEdgeFeat(HGTLayer):
             total_num_edge = 0
             for can_etype in self.edge_feat_name.keys():
                 total_num_edge += g.num_edges(etype=can_etype)
-            # print(f'Total number of feature associated edges are {total_num_edge}.')
             assert total_num_edge == 0, f"No edge features provided for {total_num_edge} edges " + \
                 "in HGTLayerwithEdgeFeat, please provide edge feature " + \
                 "dictionary specified in the \"edge_feat_name\" argument."
@@ -734,10 +733,6 @@ class HGTLayerwithEdgeFeat(HGTLayer):
                 c_etype_str = '_'.join((srctype, etype, dsttype))
                 # extract each relation as a sub graph
                 sub_graph = g[srctype, etype, dsttype]
-
-                # no edge, no message passing computation
-                if sub_graph.num_edges() == 0:
-                    continue
 
                 # extract source, destination, and edge embeds
                 src_nh = h[srctype]
--- a/python/graphstorm/model/hgt_encoder.py
+++ b/python/graphstorm/model/hgt_encoder.py
@@ -710,9 +710,22 @@ class HGTLayerwithEdgeFeat(HGTLayer):
         -------
         dict of Tensor: New node embeddings for each node type in the format of {ntype: tensor}.
         """
-        assert e_h is not None and len(e_h) != 0,  "No edge features provided for message " + \
-            "passing computation in HGTLayerwithEdgeFeat, please provide edge feature " + \
-            "dictionary specified in the \"edge_feat_name\" argument."
+        # A corner case, there is 0 edges of edge type with edge features. So no input e_h will
+        # be given.
+        if e_h is None:
+            total_num_edge = 0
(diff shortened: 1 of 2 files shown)


python/graphstorm/model/hgt_encoder.py
```diff
@@ -710,9 +710,22 @@ def forward(self, g, h, e_h=None):
...
-        assert e_h is not None and len(e_h) != 0,  "No edge features provided for message " + \
-            "passing computation in HGTLayerwithEdgeFeat, please provide edge feature " + \
-            "dictionary specified in the \"edge_feat_name\" argument."
+        # A corner case, there is 0 edges of edge type with edge features. So no input e_h will
+        # be given.
+        if e_h is None:
+            total_num_edge = 0
+            for can_etype in self.edge_feat_name.keys():
+                total_num_edge += g.num_edges(etype=can_etype)
+            # print(f'Total number of feature associated edges are {total_num_edge}.')
```
RonaldBXu, line 719: should we remove this print statement or un-comment?

python/graphstorm/model/hgt_encoder.py
```diff
@@ -710,9 +710,22 @@ def forward(self, g, h, e_h=None):
         -------
         dict of Tensor: New node embeddings for each node type in the format of {ntype: tensor}.
         """
-        assert e_h is not None and len(e_h) != 0,  "No edge features provided for message " + \
-            "passing computation in HGTLayerwithEdgeFeat, please provide edge feature " + \
-            "dictionary specified in the \"edge_feat_name\" argument."
+        # A corner case, there is 0 edges of edge type with edge features. So no input e_h will
+        # be given.
+        if e_h is None:
```
classicsong, line 715: Should it be an empty dict instead of None?

classicsong (commented):
You need to add unit test for this PR.

zhjwy9343:
> You need to add unit test for this PR.

yes, will add new unit test for this corner case.

tests/unit-tests/test_nn_model.py
```diff
@@ -1184,18 +1184,67 @@ def test_hgt_with_edge_features(input_dim, output_dim, dev):
...
+        edge_feat_name={("n0", "r0", "n1"): ['feat']},
+        edge_feat_mp_op='sub',
+        activation=th.nn.ReLU(),
+        dropout=0.0,
+        norm='')
+    init_hgtlayer(layerwithef)
+    layerwithef = layerwithef.to(dev)
+    layerwithef.eval()
+
+    # not provide edge feature as there is no edge featuer in ('n0', 'r0', 'n1')
```
RonaldBXu, line 1224:
```suggestion
    # not provide edge feature as there is no edge feature in ('n0', 'r0', 'n1')
```

RonaldBXu (approved):
LGTM

tests/unit-tests/test_nn_model.py
```diff
@@ -1184,18 +1184,67 @@ def test_hgt_with_edge_features(input_dim, output_dim, dev):
...
-    test_rgcn_with_edge_features(32, 64, 'cuda:0')
+    assert_almost_equal(baseline_emb['n1'].detach().cpu().numpy(),
+                        emb['n1'].detach().cpu().numpy())
 
-    test_hgt_with_edge_features(32, 64, 'cpu')
\ No newline at end of file
+    #       8.2 abnormal case, set layer with  ('n0', 'r0', 'n1') edge feature name, but has >0
+    #           input edges. This will trigger an asserttion error.
+    #       This has been tested in Test Case 5.
+
```
classicsong, line 1250: Where are the codes for 8.2?

RonaldBXu: I think this code already exists in test case 5. Could we rewrite this comment to "... edge feature name, has >0 input edges, but don't pass edge features to forward..."? I think that makes more sense

classicsong (commented):
Can you also check whether RGCN handles the same corner case?

classicsong (approved):
LGTM

Issue #1198, opened 2025-03-01
[Bug] Prediction and label shape mismatch error will be raised when using rmse when the label is a 1D tensor.


classicsong:
When the regression task label is a 1D tensor (Predicting 1 float value for each node), we may got the following error:
```
AssertionError: prediction and labels have different shapes. torch.Size([100, 1]) vs. torch.Size([100])
```

This is annoying as we have to rebuild the graph and cast the label tensor into a 2D tensor with the shape (len, 1).

PR #1199, opened 2025-03-01
Fix the error case when the label is a 1D tensor for regression tasks

classicsong:
*Issue #, if available:*
#1198 

*Description of changes:*
The prediction results of regression tasks are always 2D tensors, even when the prediction target is a single float value for each node. If the label is a 1D tensor, it will trigger a tensor shape mismatch error. This PR fixes this error.


RonaldBXu (approved):
LGTM

zhjwy9343 (approved):
LGTM

PR #1201, opened 2025-03-03
[GSProcessing] Have 'mean' default imputation when converting GConstr…

thvasilo:
…uct configs

*Issue #, if available:*

*Description of changes:*

* GConstruct numerical configs do not have an imputer entry so we add an imputation for GSProcessing when converting GConstruct configs. This ensure we can parse data with missing values by default, even when the config is GConstruct.


jalencato (approved):
LGTM

PR #1203, opened 2025-03-05
[Docker] Ensure hostname fix is at right place for SageMaker image. Use recommended paths for code

thvasilo:
…

*Issue #, if available:*

*Description of changes:*

* Previously our SageMaker jobs would log errors like `ERROR: ld.so: object '/libchangehostname.so' from LD_PRELOAD cannot be preloaded (cannot open shared object file): ignored.` at the beginning of the job because the entry point script is trying to use a library (changehostname.c) that gets compiled at launch time, but we were using the wrong path there, which meant there was no library compiled.
  * This PR fixes the path for the `changehostname.c`
* Our containers use a custom entry point, and code under `/opt/ml/code/` is usually reserved to be executed straight from SageMaker i.e. we do [this](https://docs.aws.amazon.com/sagemaker/latest/dg/your-algorithms-training-algo-dockerfile.html#your-algorithms-training-algo-dockerfile-api-ep-in) and not [this](https://docs.aws.amazon.com/sagemaker/latest/dg/adapt-training-container.html#byoc-training-step2). The presence of the code under `/opt/ml/code/` would occasionally cause problems in cases like HPO.
   * This PR uses `/usr/local/bin/` as our destination for our code to ensure we don't modify the files that SageMaker itself creates.

### DIFF
--- a/docker/sagemaker/Dockerfile.sm
+++ b/docker/sagemaker/Dockerfile.sm
@@ -52,7 +52,7 @@ COPY build_artifacts/start_with_right_hostname.sh /usr/local/bin/start_with_righ
 COPY code/graphstorm/ /usr/local/lib/graphstorm/
 ENV PYTHONPATH="/usr/local/lib/graphstorm/python/:${PYTHONPATH}"
 
-RUN cp /usr/local/lib/graphstorm/sagemaker/run/* /usr/local/lib/
+COPY code/graphstorm/sagemaker/run/* /opt/ml/code/
 
 # Download DGL source code
 RUN cd /root; git clone --branch v${DGL_VERSION} --single-branch https://github.com/dmlc/dgl.git
@@ -60,7 +60,7 @@ RUN cd /root; git clone --branch v${DGL_VERSION} --single-branch https://github.
 # COPY dgl /root/dgl
 ENV PYTHONPATH="/root/dgl/tools/:${PYTHONPATH}"
 
-WORKDIR /usr/local/lib/
+WORKDIR  /opt/ml/code/
 
 ENTRYPOINT ["bash", "-m", "/usr/local/bin/start_with_right_hostname.sh"]
 CMD ["/bin/bash"]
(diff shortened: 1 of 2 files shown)


docker/sagemaker/Dockerfile.sm
```diff
@@ -44,23 +44,23 @@ RUN pip3 install \
...
+COPY build_artifacts/changehostname.c /usr/local/bin/changehostname.c
 COPY build_artifacts/start_with_right_hostname.sh /usr/local/bin/start_with_right_hostname.sh
 
-# /opt/ml and all subdirectories are utilized by SageMaker, we use the /code subdirectory to store our user code.
-# TODO(xiangsx): change to pip install when PIP package is available
-COPY code/graphstorm/ /opt/ml/code/graphstorm/
-ENV PYTHONPATH="/opt/ml/code/graphstorm/python/:${PYTHONPATH}"
+# /opt/ml and all subdirectories are utilized by SageMaker,
+# we use the /usr/local/bin subdirectory to store our user code.
+COPY code/graphstorm/ /usr/local/bin/graphstorm/
```
classicsong, line 52: I think we should move graphstorm to /user/local/lib/

docker/sagemaker/Dockerfile.sm
```diff
@@ -44,23 +44,23 @@ RUN pip3 install \
...
-# TODO(xiangsx): change to pip install when PIP package is available
-COPY code/graphstorm/ /opt/ml/code/graphstorm/
-ENV PYTHONPATH="/opt/ml/code/graphstorm/python/:${PYTHONPATH}"
+# /opt/ml and all subdirectories are utilized by SageMaker,
+# we use the /usr/local/lib subdirectory to store our user code.
+COPY code/graphstorm/ /usr/local/lib/graphstorm/
+ENV PYTHONPATH="/usr/local/lib/graphstorm/python/:${PYTHONPATH}"
 
-RUN cp /opt/ml/code/graphstorm/sagemaker/run/* /opt/ml/code/
+RUN cp /usr/local/lib/graphstorm/sagemaker/run/* /usr/local/lib/
```
classicsong, line 55: For the WORKDIR, we can still use `/opt/ml/code/`.

In the future, we may use `pip install graphstorm` to replace L51 and L53.

classicsong (approved):
LGTM

Issue #1205, opened 2025-03-07
[Bug] GConstruct will fail when the input graph is a homogeneous graph without any mask.


classicsong:
When the input graph is a homogeneous graph and the user does not define any mask, the gconstruct.partition_graph will fail.

thvasilo:
Fixed in https://github.com/awslabs/graphstorm/pull/1206

PR #1206, opened 2025-03-07
[Bug Fix] Fix the case when input graph is a homogeneous graph without any mask, gconstruct will fail.

classicsong:
*Issue #, if available:*
#1205 

*Description of changes:*
Add unit test for Issue 1205 and fix the bug.


thvasilo (approved):
LGTM and the test confirms graph generation should work now.

PR #1207, opened 2025-03-07
[Docs] Add SageMaker pipeline docs

thvasilo:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docs/source/advanced/sagemaker-pipelines.rst
+++ b/docs/source/advanced/sagemaker-pipelines.rst
@@ -126,7 +126,7 @@ We list those here, with the step name that you can provide in ``--jobs-to-run``
    Multi-instance graph partitioning for distributed training.
 
 4. **GraphBolt Conversion** (``gb_convert``):
-   Converts partitioned data to GraphBolt format for improved training/inference efficiency..
+   Converts partitioned data to GraphBolt format for improved training/inference efficiency.
 
 5. **Training** (``train``):
    Trains your graph neural network model.
--- a/docs/source/advanced/sagemaker-pipelines.rst
+++ b/docs/source/advanced/sagemaker-pipelines.rst
@@ -1,0 +1,303 @@
+.. _graphstorm-sagemaker-pipeline-ref:
+
+Using SageMaker Pipelines with GraphStorm
+=========================================
+
+GraphStorm provides integration with Amazon SageMaker Pipelines to automate and orchestrate graph machine learning workflows at scale.
+This guide shows you how to use the provided tools to create, configure, and execute SageMaker pipelines for graph construction, training, and inference tasks.
+
+Introduction
+------------
+
+SageMaker Pipelines enable you to create automated MLOps workflows for your GraphStorm applications. Using these workflows you can:
+
+* Automate the end-to-end process of preparing graph data, training models, and running inference
+* Ensure reproducibility of your graph machine learning experiments
+* Scale your workflows efficiently using SageMaker's managed infrastructure
+* Track and version your pipeline executions
+
+Pre-requisites
+--------------
+
+Before starting with GraphStorm Pipelines on SageMaker, you'll need:
+
+* Python 3.8 or later
+* An AWS account with appropriate permissions. See the official
+  `SageMaker documentation <https://docs.aws.amazon.com/sagemaker/latest/dg/build-and-manage-access.html>`_
+  for detailed information about required permissions.
+* Basic familiarity with Amazon SageMaker and
+  `SageMaker Pipelines <https://docs.aws.amazon.com/sagemaker/latest/dg/pipelines.html>`_.
+* Understanding of graph neural networks and the `GraphStorm framework <https://graphstorm.readthedocs.io/en/latest/index.html>`_.
+
+Setting Up Your Environment
+---------------------------
+
+To work with GraphStorm SageMaker pipelines, you'll need the GraphStorm source code
+and a Python environment with the SageMaker SDK and AWS SDK (boto3) installed.
+
+1. First, clone the GraphStorm repository and navigate to the pipeline directory:
+
+.. code-block:: bash
+
+   git clone https://github.com/awslabs/graphstorm.git
+   cd graphstorm/sagemaker/pipeline
+
+2. Install the required Python packages:
+
+.. code-block:: bash
+
+   pip install sagemaker boto3
+
+Creating Your First Pipeline
+--------------------------
+
+GraphStorm provides Python scripts to help you create and manage SageMaker pipelines. The main tools are:
+
+* ``create_sm_pipeline.py``: Creates or updates pipeline definitions
+* ``pipeline_parameters.py``: Manages pipeline configuration
+* ``execute_sm_pipeline.py``: Runs pipeline executions
+
+Here's an example of how to create a basic pipeline that includes graph construction, training, and inference:
+
+.. code-block:: bash
+
+    python create_sm_pipeline.py \
+        --graph-construction-config-filename my_gconstruct_config.json \
+        --graph-name my-graph \
+        --graphstorm-pytorch-cpu-image-url 123456789012.dkr.ecr.us-west-2.amazonaws.com/graphstorm:sagemaker-cpu \
+        --input-data-s3 s3://input-bucket/data \
+        --instance-count 2 \
+        --jobs-to-run gconstruct train inference \
+        --output-prefix s3://output-bucket/results \
+        --pipeline-name my-graphstorm-pipeline \
+        --region us-west-2 \
+        --role arn:aws:iam::123456789012:role/SageMakerExecutionRole \
+        --train-inference-task node_classification \
+        --train-yaml-s3 s3://config-bucket/train.yaml
+
+This command sets up a pipeline with three main stages:
+
+1. Graph construction using the configuration in ``my_gconstruct_config.json``
+2. Model training using the settings in ``train.yaml``
+3. Inference using the trained model
+
+The pipeline will use 2 instances (specified by ``--instance-count``) for distributed training and inference.
+
+You'll need to provide:
+
+* A SageMaker execution role (``--role``) with appropriate permissions
+* A GraphStorm Docker image (``--graphstorm-pytorch-cpu-image-url``) for running the tasks
+* S3 locations for your input data and where to store results
+
+Running Pipeline Executions
+-------------------------
+
+Once you've created a pipeline, you can execute it using the ``execute_sm_pipeline.py`` script:
+
+.. code-block:: bash
+
+    python execute_sm_pipeline.py \
+        --pipeline-name my-graphstorm-pipeline \
+        --region us-west-2
+
+You can override default parameters during execution to customize the run:
+
+.. code-block:: bash
+
+    python execute_sm_pipeline.py \
+        --pipeline-name my-graphstorm-pipeline \
+        --region us-west-2 \
+        --instance-count 4 \
+        --gpu-instance-type ml.g4dn.12xlarge
+
+Pipeline Components
+-----------------
+
+A GraphStorm SageMaker pipeline can include several components that you can combine based on your needs.
+We list those here, with the step name that you can provide in ``--jobs-to-run`` in parentheses.
+
+1. **Single-instance Graph Construction** (``gconstruct``):
+   Single-instance graph construction for smaller graphs.
+
+2. **Distributed Graph pre-processing** (``gsprocessing``):
+   PySpark-based distributed data preparation for large graphs.
+
+3. **Distributed Graph Partitioning** (``dist_part``):
+   Multi-instance graph partitioning for distributed training.
+
+4. **GraphBolt Conversion** (``gb_convert``):
+   Converts partitioned data to GraphBolt format for improved performance.
+
+5. **Training** (``train``):
+   Trains your graph neural network model.
+
+6. **Inference** (``inference``):
+   Runs predictions using your trained model.
+
+Configuration Options
+---------------------
+
+This section provides a comprehensive list of all available configuration options for creating and executing GraphStorm SageMaker pipelines.
+
+AWS Configuration
+^^^^^^^^^^^^^^^^^
+
+* ``--execution-role``: SageMaker execution IAM role ARN. (Required)
+* ``--region``: AWS region. (Required)
+* ``--graphstorm-pytorch-cpu-image-uri``: GraphStorm GConstruct/dist_part/train/inference CPU ECR image URI. (Required)
+* ``--graphstorm-pytorch-gpu-image-uri``: GraphStorm GConstruct/dist_part/train/inference GPU ECR image URI.
+* ``--gsprocessing-pyspark-image-uri``: GSProcessing SageMaker PySpark ECR image URI. (Required if running a ``gsprocessing`` job.)
+
+Instance Configuration
+^^^^^^^^^^^^^^^^^^^^^^
+
+* ``--instance-count`` / ``--num-parts``: Number of worker instances/partitions for partition, training, inference. (Required)
+* ``--cpu-instance-type``: CPU instance type. (Default: ml.m5.4xlarge)
+* ``--gpu-instance-type``: GPU instance type. (Default: ml.g5.4xlarge)
+* ``--train-on-cpu``: Run training and inference on CPU instances instead of GPU. (Flag)
+* ``--graph-construction-instance-type``: Instance type for graph construction.
+* ``--gsprocessing-instance-count``: Number of GSProcessing instances (PySpark cluster size, default is equal to ``--instance-count``).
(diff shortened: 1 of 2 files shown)


docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+* Ensure reproducibility of your graph machine learning experiments
+* Scale your workflows efficiently using SageMaker's managed infrastructure
+* Track and version your pipeline executions
+
+Pre-requisites
+--------------
+
+Before starting with GraphStorm Pipelines on SageMaker, you'll need:
+
+* Python 3.8 or later
```
classicsong, line 24:
```suggestion
* An execution environment with Python 3.8 or later
```

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+   pip install sagemaker boto3
+
+Creating Your First Pipeline
+--------------------------
+
+GraphStorm provides Python scripts to help you create and manage SageMaker pipelines. The main tools are:
+
+* ``create_sm_pipeline.py``: Creates or updates pipeline definitions
+* ``pipeline_parameters.py``: Manages pipeline configuration
+* ``execute_sm_pipeline.py``: Runs pipeline executions
```
classicsong, line 58:
```suggestion
* ``execute_sm_pipeline.py``: Runs pipelines
```

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+1. Graph construction using the configuration in ``my_gconstruct_config.json``
+2. Model training using the settings in ``train.yaml``
+3. Inference using the trained model
+
+The pipeline will use 2 instances (specified by ``--instance-count``) for distributed training and inference.
+
+You'll need to provide:
+
+* A SageMaker execution role (``--role``) with appropriate permissions
+* A GraphStorm Docker image (``--graphstorm-pytorch-cpu-image-url``) for running the tasks
```
classicsong, line 89: Can we set different image for graph construction and training/inference?

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+        --pipeline-name my-graphstorm-pipeline \
+        --region us-west-2
+
+You can override default parameters during execution to customize the run:
+
+.. code-block:: bash
+
+    python execute_sm_pipeline.py \
+        --pipeline-name my-graphstorm-pipeline \
+        --region us-west-2 \
```
classicsong, line 109: Can we change the execution region?

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+        --gpu-instance-type ml.g4dn.12xlarge
+
+Pipeline Components
+-----------------
+
+A GraphStorm SageMaker pipeline can include several components that you can combine based on your needs.
+We list those here, with the step name that you can provide in ``--jobs-to-run`` in parentheses.
+
+1. **Single-instance Graph Construction** (``gconstruct``):
+   Single-instance graph construction for smaller graphs.
```
classicsong, line 120:
```suggestion
   Single-instance graph construction for small graphs.
```

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+
+2. **Distributed Graph pre-processing** (``gsprocessing``):
+   PySpark-based distributed data preparation for large graphs.
+
+3. **Distributed Graph Partitioning** (``dist_part``):
+   Multi-instance graph partitioning for distributed training.
+
+4. **GraphBolt Conversion** (``gb_convert``):
+   Converts partitioned data to GraphBolt format for improved performance.
```
classicsong, line 129:
```suggestion
   Converts partitioned data to GraphBolt format for improved training/inference efficiency..
```
People may misunderstand performance as model performance.

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+
+5. **Training** (``train``):
+   Trains your graph neural network model.
+
+6. **Inference** (``inference``):
+   Runs predictions using your trained model.
+
```
classicsong, line 136: It is better to have a section talking about recommended job combinations.
For example, `gsprocessing` and `dist_part` should be used together and `gconstruct` does not need `dist_part`, etc.

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+* ``--execution-role``: SageMaker execution IAM role ARN. (Required)
+* ``--region``: AWS region. (Required)
+* ``--graphstorm-pytorch-cpu-image-uri``: GraphStorm GConstruct/dist_part/train/inference CPU ECR image URI. (Required)
+* ``--graphstorm-pytorch-gpu-image-uri``: GraphStorm GConstruct/dist_part/train/inference GPU ECR image URI.
+* ``--gsprocessing-pyspark-image-uri``: GSProcessing SageMaker PySpark ECR image URI. (Required if running a ``gsprocessing`` job.)
+
+Instance Configuration
+^^^^^^^^^^^^^^^^^^^^^^
+
+* ``--instance-count`` / ``--num-parts``: Number of worker instances/partitions for partition, training, inference. (Required)
```
classicsong, line 154: We should emphasize `--instance-count` and `--num-parts` must be same.

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+* ``--cpu-instance-type``: CPU instance type. (Default: ml.m5.4xlarge)
+* ``--gpu-instance-type``: GPU instance type. (Default: ml.g5.4xlarge)
+* ``--train-on-cpu``: Run training and inference on CPU instances instead of GPU. (Flag)
+* ``--graph-construction-instance-type``: Instance type for graph construction.
+* ``--gsprocessing-instance-count``: Number of GSProcessing instances (PySpark cluster size, default is equal to ``--instance-count``).
```
classicsong, line 159: What about gsprocessing-instance-type?

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+* ``--volume-size-gb``: Additional volume size for SageMaker instances in GB. (Default: 100)
+
+Task Configuration
+^^^^^^^^^^^^^^^^^^
+
+* ``--graph-name``: Name of the graph. (Required)
+* ``--input-data-s3``: S3 path to the input graph data. (Required)
+* ``--output-prefix-s3``: S3 prefix for the output data. (Required)
+* ``--pipeline-name``: Name for the pipeline.
+* ``--base-job-name``: Base job name for SageMaker jobs. (Default: 'gs')
```
classicsong, line 169: is this a job name prefix?

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+* ``--jobs-to-run``: Space-separated string of jobs to run in the pipeline.
```
classicsong, line 170:
```suggestion
* ``--jobs-to-run``: Space-separated strings of jobs to run in the pipeline.
```

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+  Possible values are: ``gconstruct``, ``gsprocessing``, ``dist_part``, ``gb_convert``, ``train``, ``inference`` (Required).
+* ``--log-level``: Logging level for the jobs. (Default: INFO)
+* ``--step-cache-expiration``: Expiration time for the step cache. (Default: 30d)
+* ``--update-pipeline``: Update an existing pipeline instead of creating a new one. (Flag)
+
+Graph Construction Configuration
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+* ``--graph-construction-config-filename``: Filename for the graph construction config.
+* ``--graph-construction-args``: Additional parameters to be passed directly to the GConstruct/GSProcessing job.
```
classicsong, line 180: It is better to provide an example here.

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+
+Partition Configuration
+^^^^^^^^^^^^^^^^^^^^^^^
+
+* ``--partition-algorithm``: Partitioning algorithm to use. (Default: random)
+* ``--partition-input-json``: Name for the JSON file that describes the input data for distributed partitioning. (Default: updated_row_counts_metadata.json)
```
classicsong, line 186: Are `--partition-input-json` and `--partition-output-json` optional?

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+* ``--partition-output-json``: Name for the output JSON file that describes the partitioned data generated by GConstruct or GSPartition.
+  (Default: metadata.json for GSPartition,  use <graph_name>.json for ``gconstruct``.)
+
+Training Configuration
+^^^^^^^^^^^^^^^^^^^^^^
+
+* ``--model-output-path``: S3 path for model output.
+* ``--num-trainers``: Number of trainers (per-instance training processes) to use during training/inference. Set this equal to number of GPUs (Default: 4)
```
classicsong, line 194: The default GPU instance type only has 1 GPUs, so if a user sets up his pipeline with all default arguments, the training task will fail.

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+* ``--train-inference-task-type``: Task type for training and inference. (Required)
```
classicsong, line 195: Can you list the possible options?

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+
+* ``--inference-yaml-s3``: S3 path to inference YAML configuration file.
+* ``--inference-model-snapshot``: Which model snapshot to choose to run inference with, e.g. ``epoch-9`` to use the model generated by the 10th (zero-indexed) epoch.
+* ``--save-predictions``: Whether to save predictions to S3 during inference. (Flag)
+* ``--save-embeddings``: Whether to save embeddings to S3 during inference. (Flag)
+
+Script Paths
+^^^^^^^^^^^^
+
+* ``--dist-part-script``: Path to DistPartition SageMaker entry point script.
```
classicsong, line 210: Can you also point the readers where to find these scripts in GraphStorm repo?

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+* ``--gb-convert-script``: Path to GraphBolt partition conversion script.
+* ``--train-script``: Path to training SageMaker entry point script.
+* ``--inference-script``: Path to inference SageMaker entry point script.
+* ``--gconstruct-script``: Path to GConstruct SageMaker entry point script.
+* ``--gsprocessing-script``: Path to GSProcessing SageMaker entry point script.
+
+Using Configuration Options
```
classicsong, line 217:
```suggestion
Using Configuration Options (Example)
```

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -0,0 +1,303 @@
...
+
+GraphBolt enabled faster training, see :ref:`using-graphbolt-ref`. To enable GraphBolt for your pipeline:
+
+.. code-block:: bash
+
+    python create_sm_pipeline.py \
+        ... \
+        --use-graphbolt true
+
+For distributed processing with GraphBolt, you will need to include a ``gb_convert`` step after ``dist_part``:
```
classicsong, line 255: What about gconstruct?

docs/source/advanced/sagemaker-pipelines.rst
thvasilo: For GSProcessing yes. GConstruct always uses the Pytorch CPU image. Would we like the GConstruct image to be separate?

docs/source/advanced/sagemaker-pipelines.rst
thvasilo: We can't change the execution region, but when I was writing the execution script I wasn't able to find a consistent way to derive it from the user environment so I made it a required arg instead.

We could look further into this and try to guess the execution region (every pipeline is unique per account/region combination)

docs/source/advanced/sagemaker-pipelines.rst
thvasilo: There's only one argument, we're just using an alias to use both options (this makes it consistent with our sagemaker launch scripts and other documentation).

So the user can only set this once.

docs/source/advanced/sagemaker-pipelines.rst
thvasilo: Yes, SageMaker uses "base" as the argument name so I used the same here.

docs/source/advanced/sagemaker-pipelines.rst
thvasilo: Yes, I think we derive them from the job sequence provided.

docs/source/advanced/sagemaker-pipelines.rst
thvasilo: I'll link to the section in the docs so we don't have to duplicate this info (tasks can change, one less place to keep updated)

docs/source/advanced/sagemaker-pipelines.rst
thvasilo: We can set a different image for GSProcessing, GConstruct will use the PyTorch CPU image.

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -126,14 +126,22 @@ We list those here, with the step name that you can provide in ``--jobs-to-run``
    Multi-instance graph partitioning for distributed training.
 
 4. **GraphBolt Conversion** (``gb_convert``):
-   Converts partitioned data to GraphBolt format for improved performance.
+   Converts partitioned data to GraphBolt format for improved training/inference efficiency..
```
classicsong, line 129:
```suggestion
   Converts partitioned data to GraphBolt format for improved training/inference efficiency.
```

docs/source/advanced/sagemaker-pipelines.rst
```diff
@@ -126,14 +126,22 @@ We list those here, with the step name that you can provide in ``--jobs-to-run``
...
 6. **Inference** (``inference``):
    Runs predictions using your trained model.
 
+The choice of jobs to run will mostly stem from the size of your graph.
+For graphs that can fit into the memory of one machine, a typical
+job sequence would be ``gconstruct train inference``.
+
+For graphs that are too large to fit into one machine, you will need to
+pre-process them using GSProcessing and use distributed GSPartition.
+Such a sequence of jobs would be ``gsprocessing dist_part train inference``.
```
classicsong, line 143: Do we support `gsprocessing dist_part inference` ?

docs/source/advanced/sagemaker-pipelines.rst
classicsong: OK

docs/source/advanced/sagemaker-pipelines.rst
classicsong: Can we add a hint in the doc highlighting this restriction, i.e., once the region is set for a pipeline, user can only run tasks on that region?

docs/source/advanced/sagemaker-pipelines.rst
thvasilo: I haven't tried it but we don't have any restrictions against such a sequence.

thvasilo: Actually took a look at the implementation. Right now we do not support this because the inference step always gets its model S3 path from a train step which we assume was created previously. 

To support inference-only pipeline creation we would need to make a couple of changes to the pipeline creation code.
In code a user can hack this by assigning a value to `self.model_input_path` after they create a `GraphStormPipelineGenerator`, but there's no CLI support yet.

classicsong: Let's create a corresponding issue for 0.5 release.

classicsong (approved):
LGTM

docs/source/advanced/sagemaker-pipelines.rst
thvasilo: Created https://github.com/awslabs/graphstorm/issues/1211

Issue #1220, opened 2025-03-21
Cannot use early stopping when `report_eval_per_type` is set to true


thvasilo:
Early stopping assumes a scalar metric, while per-type metrics are returned as dicts.

Need to come up with a mechanism that allows early stop to be used with per-type eval (e.g. average improvement over all metrics)

![Image](https://github.com/user-attachments/assets/b2a8aaee-0265-4eb3-96a0-0226eaae1360)

PR #1223, opened 2025-03-25
[BugFix] Fix the case when early stopping does not work with report_eval_per_type

classicsong:
*Issue #, if available:*
#1220 

*Description of changes:*
Fix the bug.


python/graphstorm/eval/evaluator.py
```diff
@@ -341,6 +341,14 @@ def do_eval(self, total_iters, epoch_end=False):
             return True
         return False
 
+    def _get_early_stop_score(self, val_score):
+        """ By default, always get the first validation
+        score as the indicator for early stop.
+        """
+        # First metric is the key eval metric
+        metric = self.metric_list[0]
```
thvasilo, line 349: Is there a guarantee metric_list will have at least one element?

python/graphstorm/eval/evaluator.py
```diff
@@ -341,6 +341,14 @@ def do_eval(self, total_iters, epoch_end=False):
             return True
         return False
 
+    def _get_early_stop_score(self, val_score):
```
thvasilo, line 344: Since it looks like this is becoming part of the expected inner interface of the class (`do_early_stop` is public, and calls this inner function) can we add some type annotations so we know what to expect in the input arguments and return values? Then new implementers of this function will not what is the expected input/output.

python/graphstorm/eval/evaluator.py
```diff
@@ -1655,6 +1663,13 @@ def _get_major_score(self, score):
             major_score = score[self.major_etype]
         return major_score
 
+    def _get_early_stop_score(self, val_score):
+        """ By default, call _get_major_score
```
thvasilo, line 1667: This is an implementation detail, we can either rely on the docstring of the parent class, or include a functional description here (ie. what this function does, not how it does it)

tests/unit-tests/test_evaluator.py
```diff
@@ -1412,15 +1412,87 @@ def test_early_stop_lp_evaluator():
...
+    evaluator = GSgnnPerEtypeMrrLPEvaluator(config.eval_frequency,
+                                    use_early_stop=config.use_early_stop,
+                                    early_stop_burnin_rounds=config.early_stop_burnin_rounds,
+                                    early_stop_rounds=config.early_stop_rounds,
+                                    early_stop_strategy=config.early_stop_strategy)
+    for _ in range(5):
+        # always return false
+        assert evaluator.do_early_stop(
+            {"mrr": {("src", "r0", "dst"):0.4,
+                     ("src", "r1", "dst"):0.6}}) is False
```
thvasilo, line 1443: Since the input to `do_early_stop` doesn't change is there any benefit to calling it 5 times?

tests/unit-tests/test_evaluator.py
```diff
@@ -1412,15 +1412,87 @@ def test_early_stop_lp_evaluator():
...
+
+    for _ in range(3):
```
thvasilo, line 1445: Same here, why do we call this multiple times?

tests/unit-tests/test_evaluator.py
```diff
@@ -1412,15 +1412,87 @@ def test_early_stop_lp_evaluator():
...
+
+    assert evaluator.do_early_stop(
+        {"mrr": {("src", "r0", "dst"):0.7,
+                 ("src", "r1", "dst"):0.7}}) is False # better than average
+    assert evaluator.do_early_stop(
+        {"mrr": {("src", "r0", "dst"):0.68,
+                 ("src", "r1", "dst"):0.6}}) is False # still better
+    assert evaluator.do_early_stop(
+        {"mrr": {("src", "r0", "dst"):0.66,
+                 ("src", "r1", "dst"):0.5}})
```
thvasilo, line 1495: Could we move this to a separate test since they are testing different implementations? We can share setup with a fixture.

```
import pytest

@pytest.fixture()
def early_stop_config() -> Dummy:
    """Create a dummy early stop configuration object"""
    return Dummy({
            "eval_frequency": 100,
            "use_early_stop": True,
            "early_stop_burnin_rounds": 5,
            "early_stop_rounds": 3,
            "early_stop_strategy": EARLY_STOP_AVERAGE_INCREASE_STRATEGY,
        })

def test_early_stop_per_etype(early_stop_config: Dummy):
     evaluator = GSgnnPerEtypeMrrLPEvaluator(early_stop_config.eval_frequency,
                                    use_early_stop=early_stop_config.use_early_stop,
                                    early_stop_burnin_rounds=early_stop_config.early_stop_burnin_rounds,
                                    early_stop_rounds=early_stop_config.early_stop_rounds,
                                    early_stop_strategy=early_stop_config.early_stop_strategy)
```

tests/unit-tests/test_evaluator.py
```diff
@@ -1654,24 +1726,3 @@ def check_multi_task_eval(mock_reg_compute_score, mock_class_compute_score, mock
         assert best_iter_num["r_eval"]["rmse"] == 200
 
     check_multi_task_eval()
-
```
thvasilo, line 1657: Let's leave the final newline this changed removed it i think

python/graphstorm/eval/evaluator.py
zhjwy9343: +1. And can we check if the `metric` is in the `val_score`?

classicsong: There is at least one element in the `metric_list`.
But we can add an assert to check ` if the metric is in the val_score`

tests/unit-tests/test_evaluator.py
classicsong: early_stop_burnin_rounds is set to 5, so the first five rounds should not trigger early stop.

tests/unit-tests/test_evaluator.py
classicsong: As, we set `early_stop_rounds=3`, so at least, we need to run 3 more rounds before triggering early stop.

python/graphstorm/eval/evaluator.py
```diff
@@ -341,12 +341,16 @@ def do_eval(self, total_iters, epoch_end=False):
...
-    def _get_early_stop_score(self, val_score):
+    def _get_early_stop_score(self, val_score: dict) -> float:
         """ By default, always get the first validation
         score as the indicator for early stop.
         """
         # First metric is the key eval metric
         metric = self.metric_list[0]
+        assert metric in val_score, \
+            f"{metric} is used to guide the early stop. " \
+            f"But the evaluation scores do not report {metric}." \
```
thvasilo, line 352:
```suggestion
            f"But the evaluation scores did not include {metric}. " \
```

python/graphstorm/eval/evaluator.py
```diff
@@ -1236,6 +1240,16 @@ def _get_major_score(self, score):
             major_score = score[self.major_etype]
         return major_score
 
+    def _get_early_stop_score(self, val_score: dict[dict]) -> float:
+        # First metric is the key eval metric
+        metric = self.metric_list[0]
+        assert metric in val_score, \
+            f"{metric} is used to guide the early stop. " \
+            f"But the evaluation scores do not report {metric}." \
```
thvasilo, line 1248:
```suggestion
            f"But the evaluation scores did not include {metric}. " \
```

python/graphstorm/eval/evaluator.py
```diff
@@ -1236,6 +1240,16 @@ def _get_major_score(self, score):
...
+            f"Instead we get {val_score}"
+        # call _get_major_score to get the score for early stop.
+        return self._get_major_score(val_score[metric])
```
thvasilo, line 1251: Should be `return val_score[metric]` ?

python/graphstorm/eval/evaluator.py
```diff
@@ -1663,12 +1675,17 @@ def _get_major_score(self, score):
...
+        score as the indicator for early stop.
         """
         # First metric is the key eval metric
         metric = self.metric_list[0]
-        return self._get_major_score(val_score[metric])
+        assert metric in val_score, \
+            f"{metric} is used to guide the early stop. " \
+            f"But the evaluation scores do not report {metric}." \
+            f"Instead we get {val_score}"
+        return val_score[metric]
```
thvasilo, line 1688: Should be `return self._get_major_score(val_score[metric])` ?

python/graphstorm/eval/evaluator.py
```diff
@@ -1981,6 +1996,18 @@ def _get_major_score(self, score):
             major_score = score[self.major_etype]
         return major_score
 
+    def _get_early_stop_score(self, val_score: dict) -> float:
```
thvasilo, line 1999: Are there other classes that implement `GSgnnBaseEvaluator`? If yes they need to implement `_get_early_stop_score` as well.

python/graphstorm/eval/evaluator.py
classicsong: For per edge type evaluator, we call `_get_major_score`. By default `_get_major_score` will return the average score of all the edge types. But if a user provides the `major_etype`, then the score of `major_etype` will be returned.

python/graphstorm/eval/evaluator.py
classicsong: Yes. But most of they can use the GSgnnBaseEvaluator. _get_early_stop_score.

tests/unit-tests/test_evaluator.py
thvasilo: Can we add a comment about what's happening here? 

```
# early_stop_burnin_rounds is set to 5, so even though the metrics don't change for 5 rounds, early stop shouldn't trigger
```

tests/unit-tests/test_evaluator.py
thvasilo: Same here, let's add a more verbose explanation, as this test is sequentially dependent on the last one.

```
# After the 5 burn-in rounds, because we set early_stop_rounds=3, we shouldn't trigger early stop for the first 3 rounds of non-improving metrics
```

thvasilo (approved):
LGTM, just a couple of clarifications

PR #1225, opened 2025-03-28
[GConstruct] [GSProcessing] Allow no-op operation in gconstruct to read strings of delimited numbers as vectors

thvasilo:
*Issue #, if available:*

*Description of changes:*

* We add an explicit `no-op` transformation to GConstruct that can read delimited strings as vectors. 
* The previous no-op behavior when "transform" entry is not provided remains unchanged.
* Add docs and handling for conversion to GSProcessing.

### DIFF
--- a/docs/source/cli/graph-construction/single-machine-gconstruct.rst
+++ b/docs/source/cli/graph-construction/single-machine-gconstruct.rst
@@ -187,14 +187,22 @@ GraphStorm provides a set of transformation operations for different types of fe
                   "bucket_cnt": 2,
                   "slide_window_size": 10},
 
-* **No-op vector truncation (experimental)** truncates feature vectors to the length requested. The ``name`` field can be empty (e.g., ``{name: }``), and an integer ``truncate_dim`` value will determine the length of the output vector. This can be useful when experimenting with input features that were trained using `Matryoshka Representation Learning <https://arxiv.org/abs/2205.13147>`_.
+* **No-op vector parsing and truncation (experimental)** This is an no-op transformation that passes data along as-is. The ``name`` field needs to be ``no-op``.
+  Optionally it can parse vectors formatted as strings, splitting each string in a row with a user-defined separator string. For example,
+  if your file contains entries like ``0.4;0.3;0.9`` you can parse this as a numerical vector by setting the ``separator`` field to ``;``.
+  This operation can also truncate feature vectors to a specific length, by setting
+  the field ``truncate_dim`` to an integer value that will determine the length of the output vector.
+  This can be useful when experimenting with input features that were trained using `Matryoshka Representation Learning <https://arxiv.org/abs/2205.13147>`_.
 
   Example:
 
   .. code:: json
 
-    "transform": {"name": ,
-                  "truncate_dim": 24},
+    "transform": {
+        "name": "no-op",
+        "separator": ";",
+        "truncate_dim": 64
+    }
 
 .. _gcon-output-format:
 
--- a/docs/source/cli/graph-construction/single-machine-gconstruct.rst
+++ b/docs/source/cli/graph-construction/single-machine-gconstruct.rst
@@ -187,7 +187,7 @@ GraphStorm provides a set of transformation operations for different types of fe
                   "bucket_cnt": 2,
                   "slide_window_size": 10},
 
-* **No-op vector parsing and truncation (experimental)** This is an no-op transformation that passes data along as-is. The ``name`` field needs to be ``no-op``.
+* **No-op vector parsing and truncation (experimental)** This is a no-op transformation that passes data along as-is. The ``name`` field needs to be ``no-op``.
   Optionally it can parse vectors formatted as strings, splitting each string in a row with a user-defined separator string. For example,
   if your file contains entries like ``0.4;0.3;0.9`` you can parse this as a numerical vector by setting the ``separator`` field to ``;``.
   This operation can also truncate feature vectors to a specific length, by setting
--- a/graphstorm-processing/graphstorm_processing/config/config_conversion/gconstruct_converter.py
+++ b/graphstorm-processing/graphstorm_processing/config/config_conversion/gconstruct_converter.py
@@ -225,6 +225,16 @@ class GConstructConfigConverter(ConfigConverter):
                         gsp_transformation_dict["kwargs"] = {
                             "separator": gconstruct_transform_dict["separator"],
                         }
+                elif gconstruct_transform_dict["name"] == "no-op":
+                    gsp_transformation_dict["name"] = "no-op"
+                    if "separator" in gconstruct_transform_dict:
+                        gsp_transformation_dict["kwargs"] = {
+                            "separator": gconstruct_transform_dict["separator"],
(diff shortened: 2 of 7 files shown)


docs/source/cli/graph-construction/single-machine-gconstruct.rst
```diff
@@ -187,14 +187,22 @@ GraphStorm provides a set of transformation operations for different types of fe
                   "bucket_cnt": 2,
                   "slide_window_size": 10},
 
-* **No-op vector truncation (experimental)** truncates feature vectors to the length requested. The ``name`` field can be empty (e.g., ``{name: }``), and an integer ``truncate_dim`` value will determine the length of the output vector. This can be useful when experimenting with input features that were trained using `Matryoshka Representation Learning <https://arxiv.org/abs/2205.13147>`_.
+* **No-op vector parsing and truncation (experimental)** This is an no-op transformation that passes data along as-is. The ``name`` field needs to be ``no-op``.
```
jalencato, line 190:
```suggestion
* **No-op vector parsing and truncation (experimental)** This is a no-op transformation that passes data along as-is. The ``name`` field needs to be ``no-op``.
```

python/graphstorm/gconstruct/transform.py
```diff
@@ -1056,15 +1059,24 @@ class Noop(FeatTransform):
     out_dtype : str
         The dtype of the transformed feature.
         Default: None, we will not do data type casting.
-    truncate_dim : int
+    truncate_dim : int, optional
         When provided, will truncate the output float-vector feature to the specified dimension.
         This is useful when the feature is a multi-dimensional vector and we only need
         a subset of the dimensions, e.g. for Matryoshka Representation Learning embeddings.
+    separator: str , optional
```
jalencato, line 1066:
```suggestion
    separator: str, optional
```

python/graphstorm/gconstruct/transform.py
```diff
@@ -1078,6 +1090,27 @@ def call(self, feats):
         -------
         dict : The key is the feature name, the value is the feature.
         """
+
+        if self.separator is not None:
+            if isinstance(feats, ExtMemArrayWrapper):
+                feats = feats.to_numpy()
+
+            # Ensure the array is string type
+            feats = np.array(feats, dtype=str)
```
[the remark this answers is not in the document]
thvasilo: If the user provides a separator we have to make the assumption that the data will contain the separator string, therefore we implicitly assume the data type is string. If it was just a single number the user should not be providing a separator.

We can look into corner cases like the separator being the empty string and make sure we cover that case, but for now I'm assuming valid input from the user. We only check for type of the separator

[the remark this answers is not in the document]
jalencato: Make sense.

graphstorm-processing/graphstorm_processing/config/config_conversion/gconstruct_converter.py
```diff
@@ -225,6 +225,16 @@ def _convert_feature(feats: list[Mapping[str, Any]]) -> list[dict]:
...
                             "separator": gconstruct_transform_dict["separator"],
                         }
+                elif gconstruct_transform_dict["name"] == "no-op":
+                    gsp_transformation_dict["name"] = "no-op"
+                    if "separator" in gconstruct_transform_dict:
+                        gsp_transformation_dict["kwargs"] = {
+                            "separator": gconstruct_transform_dict["separator"],
+                        }
+                    if "truncate_dim" in gconstruct_transform_dict:
+                        gsp_transformation_dict["kwargs"]["truncate_dim"] = (
```
classicsong, line 235: Why L235 and L231 are different?

python/graphstorm/gconstruct/transform.py
```diff
@@ -1056,15 +1059,24 @@ class Noop(FeatTransform):
...
         a subset of the dimensions, e.g. for Matryoshka Representation Learning embeddings.
+    separator: str, optional
+        When provided will split every string in the input array along this separator
+        to create an array of vectors.
+
+        .. versionadded:: 0.5.0
     """
-    def __init__(self, col_name, feat_name, out_dtype=None, truncate_dim=None):
+    def __init__(self, col_name, feat_name, out_dtype=None,
+                 truncate_dim=None, separator: Optional[str]=None):
```
classicsong, line 1073: truncate_dim=None -> truncate_dim: Optional[int]=None

python/graphstorm/gconstruct/transform.py
```diff
@@ -1078,6 +1090,27 @@ def call(self, feats):
...
+
+            # Ensure the array is string type
+            feats = np.array(feats, dtype=str)
+
+            # Split every string in the input array along the specified delimiter,
+            # using vectorized numpy op
+            feats = np.char.split(feats, sep=self.separator)
+
+            # Convert to regular Python list
+            feats_list = feats.tolist()
```
classicsong, line 1106: Can we directly loop over feats?

python/graphstorm/gconstruct/transform.py
```diff
@@ -1088,13 +1121,17 @@ def call(self, feats):
...
+
+
         if self.truncate_dim is not None:
             if isinstance(feats, np.ndarray):
                 feats = feats[:, :self.truncate_dim]
             else:
                 assert isinstance(feats, ExtMemArrayWrapper)
                 # Need to convert to in-memory array to make truncation possible
                 feats = feats.to_numpy()[:, :self.truncate_dim]
+
```
classicsong, line 1133: Too many new lines

python/graphstorm/gconstruct/transform.py
```diff
@@ -1404,6 +1442,18 @@ def parse_feat_ops(confs, input_data_format=None):
                 transform = HardEdgeDstNegativeTransform(feat['feature_col'],
                                                          feat_name,
                                                          separator=separator)
+            elif conf['name'] == 'no-op':
+                if 'separator' in conf:
+                    assert isinstance(conf['separator'], str)
+                if 'truncate_dim' in conf:
+                    assert isinstance(conf['truncate_dim'], Number)
```
classicsong, line 1449: truncate_dim can not be a float.

classicsong: Add an error message.

python/graphstorm/gconstruct/utils.py
```diff
@@ -25,6 +25,7 @@
 import traceback
 import shutil
 import uuid
+from abc import ABC, abstractmethod
```
classicsong, line 28: Why we want to update this file?

tests/unit-tests/gconstruct/test_construct_graph.py
```diff
@@ -229,6 +229,28 @@ def check_feat_ops_noop():
     assert res[0].feat_name == feat_op1[0]["feature_col"]
     assert isinstance(res[0], Noop)
 
+def test_noop_string():
+    text_vector_data = {
```
classicsong, line 233: Can you also test the case when there is no separator?
Also the case when there is a separator and also truncate_dim.

PR #1227, opened 2025-03-28
Don't print torchdata deprecation warnings

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Currently every time DGL is imported the torchdata deprecation warnings are printed. For logs with multiple workers and trainers we can have the same message repeated tens of times (once per DGL process), polluting the logs. 
* This PR filters out the warning since we are aware of the deprecation notice and it come from DGL and not graphstorm

Old behavior:

```
root@3ef2b8d262b3:/opt/ml/code# python
>>> import graphstorm
/opt/conda/lib/python3.11/site-packages/torchdata/datapipes/__init__.py:18: UserWarning:
################################################################################
WARNING!
The 'datapipes', 'dataloader2' modules are deprecated and will be removed in a
future torchdata release! Please see https://github.com/pytorch/data/issues/1196
to learn more and leave feedback.
################################################################################

  deprecation_warning()
>>>
```

new behavior:

```
root@88e89a8ea077:/opt/ml/code# python
>>> import graphstorm
>>>
```

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1228, opened 2025-03-30
Apply suggestions for Noop transformation

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Apply final comments from #1225 
* Also change test names for other transformations. Previously all transformations were tested under a single unit test. This meant that if one of the tests failed the following ones would not run which is not the intended behavior when testing, we don't want tests to depend on each other. This also allows the individual tests to be discoverable and testable within an IDE.
* Also fixes a pylint error that was introduced in #1227 

### DIFF
--- a/python/graphstorm/gconstruct/transform.py
+++ b/python/graphstorm/gconstruct/transform.py
@@ -1118,8 +1118,6 @@ class Noop(FeatTransform):
             assert validate_numerical_feats(feats), \
                 f"There are NaN, Inf or missing value in the {self.feat_name} feature."
 
-
-
         if self.truncate_dim is not None:
             if isinstance(feats, np.ndarray):
                 feats = feats[:, :self.truncate_dim]
@@ -1127,7 +1125,6 @@ class Noop(FeatTransform):
                 assert isinstance(feats, ExtMemArrayWrapper)
                 # Need to convert to in-memory array to make truncation possible
                 feats = feats.to_numpy()[:, :self.truncate_dim]
-
 
         return {self.feat_name: feats}
 
--- a/python/graphstorm/gconstruct/transform.py
+++ b/python/graphstorm/gconstruct/transform.py
@@ -24,7 +24,7 @@ import abc
 import json
 import warnings
 from abc import ABC, abstractmethod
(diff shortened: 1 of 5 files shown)


python/graphstorm/gconstruct/transform.py
```diff
@@ -1102,11 +1102,8 @@ def call(self, feats):
             # using vectorized numpy op
             feats = np.char.split(feats, sep=self.separator)
 
-            # Convert to regular Python list
-            feats_list = feats.tolist()
-
-            # Convert lists of str to lists of float, this is faster than using numpy
-            feats = [[float(x) for x in sublist] for sublist in feats_list]
+            # Convert array of str to lists of float, this is faster than using numpy
+            feats = [[float(x) for x in sublist] for sublist in feats]
```
jalencato, line 1106: What is the expected type on the ``feats`` here? Why do we remove the ``feats.tolist()``?

thvasilo: This is to address this comment: https://github.com/awslabs/graphstorm/pull/1225/files#r2019697591

At input feats is an ndarray with str dtype. The returned value is a list of float lists. We then use that to construct a new float ndarray. 

The reason we do this back and forth conversion is that it's faster to apply the string casting in pure python than using numpy. We could also use pandas for this that seems to be as/more efficient. 

The benchmark below has the fastest method as 1.00 relative speed, and anything above that is slower by the factor mentioned, i.e. using `np.array([np.array([float(x) for x in sublist]) for sublist in arr]` which we list as "apply along axis" is 1.34x slower than a list comprehension.

```
Benchmark results for converting 10000 lists of 256 strings to floats:
Method                    Avg Time (s)    Relative Speed 
-------------------------------------------------------
List comprehension (current)        0.192263        1.00           x
Pandas apply              0.193357        1.01           x
Map with lambda           0.209482        1.09           x
List comp + NumPy array   0.224480        1.17           x
Apply along axis          0.258348        1.34           x
NumPy vectorize           1.765081        9.18           x
```

jalencato (approved):
LGTM, can merge after passing all the tests.

PR #1230, opened 2025-03-31
[GConstruct/GSProcessing] Allow GConstruct accept GSProcessing Config

jalencato:
*Issue #, if available:*

*Description of changes:*

* Allow GConstruct to accept GSProcessing Config, necessary for our real-time inference workflow.
* Imitate our previous GSProcessing-to-GConstruct Config Conversion part. 
* Add pytest and e2e test for GConstruct with GSProcessing Config input.

Done testing locally and on sagemaker. We can launch gconstruct job with gsprocessing config on sagemaker. Similar locally:
```
python3 launch/launch_gconstruct.py \
          --graph-name <graph-name> \
          --graph-data-s3 <input-s3-url> \
          --graph-config-file <gsp-config-name> \
          --output-graph-s3 <output-s3-url> \
          --role <role-name> \
          --image-url <image-url> \
          --region <region-name>
```


tests/end2end-tests/data_gen/movielens_gsp.json
```diff
@@ -0,0 +1,94 @@
+{
+    "version": "gsprocessing-v1.0",
+    "graph": {
+        "nodes": [
+            {
+                "data": {
+                    "format": "parquet",
+                    "files": [
+                        "/data/ml-100k/users.parquet"
```
thvasilo, line 9: GSProcessing does not allow paths that start with `/`

tests/end2end-tests/data_process/movielens_test.sh
```diff
@@ -32,6 +32,11 @@ python3 -m graphstorm.gconstruct.construct_graph --conf-file $GS_HOME/tests/end2
 
 error_and_exit $?
 
+echo "********* Test the GSProcessing Config Input ********"
+python3 -m graphstorm.gconstruct.construct_graph --conf-file $GS_HOME/tests/end2end-tests/data_gen/movielens_gsp.json --num-processes 1 --output-dir /tmp/movielens_gsp --graph-name ml-gsp
```
thvasilo, line 36: How long does this test take to run? I'd like us to only add end2end tests when they are testing something that's missing from the unit tests, because our test time keeps increasing.
Is there something here that we could be testing with a unit test instead?

tests/unit-tests/gconstruct/test_converter.py
```diff
@@ -0,0 +1,521 @@
...
+    """We should test about giving unknown feature transformation type."""
+    node_dict["nodes"][0]["features"] = [
+        {
+            "column": ["paper_title"],
+            "transformation": {"name": "unknown"},
+        }
+    ]
+
+    with pytest.raises(ValueError):
+        _ = converter.convert_nodes(node_dict["nodes"])
```
thvasilo, line 57:
```suggestion
        converter.convert_nodes(node_dict["nodes"])
```

tests/unit-tests/gconstruct/test_converter.py
```diff
@@ -0,0 +1,521 @@
...
+                "train": ["customized_label/node_train_idx.parquet"],
+                "valid": ["customized_label/node_val_idx.parquet"],
+                "test": ["customized_label/node_test_idx.parquet"],
+                "column": ["ID"],
+            },
+        }
+    ]
+
+
+@pytest.mark.parametrize("col_name", ["author"])
```
thvasilo, line 185: If we only test one variant we don't need to parametrize right?

tests/unit-tests/gconstruct/test_converter.py
```diff
@@ -0,0 +1,521 @@
...
+def test_convert_gsprocessing_config(converter: GSProcessingConfigConverter):
+    """Multiple test cases for end2end GSProcessing-to-GConstruct conversion"""
+    # test empty
+    assert converter.convert_to_gconstruct({}) == {
+        "version": "gconstruct-v0.1",
+        "nodes": [],
+        "edges": [],
+    }
+
+    gsp_conf = {}
```
thvasilo, line 301: Can we put all the GSP config in a file instead?

tests/unit-tests/gconstruct/test_converter.py
```diff
@@ -0,0 +1,521 @@
...
+                        "train": 0.9,
+                        "val": 0.1,
+                        "test": 0.0
+                    },
+                },
+            ],
+        }
+    ]
+
+    assert len(converter.convert_to_gconstruct(gsp_conf["graph"])["nodes"]) == 1
```
thvasilo, line 434: Let's assign `converter.convert_to_gconstruct(gsp_conf["graph"])` to a value we re-use, instead of calling the conversion each time.

python/graphstorm/gconstruct/config_conversion/gsprocessing_converter.py
```diff
@@ -0,0 +1,329 @@
...
+            gconstruct_feat_dict["feature_col"] = gsp_feat_dict["column"]
+            if "name" in gsp_feat_dict:
+                gconstruct_feat_dict["feature_name"] = gsp_feat_dict["name"]
+
+            gconstruct_transformation_dict: dict[str, Any] = {}
+            if "transformation" in gsp_feat_dict:
+                gsp_transformation_dict = gsp_feat_dict["transformation"]
+
+                ### START TRANSFORMATION ALTERNATIVES ###
+                if gsp_transformation_dict["name"] == "numerical":
```
thvasilo, line 149: We have support for string-vectors+truncating in no-op Gconstruct now. Can we ensure to include that?

python/graphstorm/gconstruct/config_conversion/gsprocessing_converter.py
```diff
@@ -0,0 +1,329 @@
...
+                            f"{gsp_transformation_dict['kwargs']['normalizer']}"
+                        )
+                    if "out_dtype" in gsp_transformation_dict["kwargs"]:
+                        gconstruct_transformation_dict["out_dtype"] = (
+                            gsp_transformation_dict["kwargs"]["out_dtype"]
+                        )
+                elif gsp_transformation_dict["name"] == "bucket_numerical":
+                    assert (
+                        "bucket_cnt" in gsp_transformation_dict["kwargs"]
+                    ), "bucket_cnt should be in the gconstruct bucket feature transform field"
```
thvasilo, line 174: Do we mean gsprocessing bucket feature transform field? We're creating the gconstruct transform. Similarly for other similar errors here.

python/graphstorm/gconstruct/config_conversion/gsprocessing_converter.py
```diff
@@ -0,0 +1,329 @@
...
+            # format
+            # Assume the format is already valid in GSProcessing
+            node_format = n["data"]["format"]
+            if "separator" not in n["data"]:
+                node_separator = None
+            else:
+                node_separator = n["data"]["separator"]
+
+            # files
+            # Do we support S3 in GConstruct?
```
thvasilo, line 243: We do not, the assumption here should be that all paths are relative. 

We need practical examples of how the converter would be used locally and on SageMaker.

thvasilo (changes requested):
I would like to see a practical example of how to use the converter, my main concern being that GSprocessing will always have relative paths, so I'd like to see those applied in GConstruct. For example  how would we run this on SageMaker and locally.

tests/end2end-tests/data_process/movielens_test.sh
jalencato: The test here should be very quick. This tests can be very necessary as our actual entry is in the ``construct_graph``.  When doing unit testing, we only do testing for the new converter python class. We should also do end-to-end test to verify if the whole test works.

jalencato:
> I would like to see a practical example of how to use the converter, my main concern being that GSprocessing will always have relative paths, so I'd like to see those applied in GConstruct. For example how would we run this on SageMaker and locally.

Ideally we can modify the tests to relative path to test on both locally and SageMaker. I would say let us add one more test internally for the SageMaker regression test for it. GConstruct definitely supports relative path.

python/graphstorm/gconstruct/config_conversion/gsprocessing_converter.py
jalencato: Right. We have similar assertion as well.

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -792,6 +794,17 @@ def process_graph(args):
...
+        process_confs["version"] = "gconstruct-v0.1"
+    if process_confs["version"].startswith("gsprocessing"):
+        logging.warning("Need to convert to GConstruct Config")
+        gsp_converter = GSProcessingConfigConverter()
+        process_confs = gsp_converter.convert_to_gconstruct(process_confs["graph"])
+    elif process_confs["version"].startswith("gconstruct"):
+        logging.info("Parsing config file as GConstruct config")
+    else:
+        logging.warning("Unrecognized configuration file version name: %s",
+                        process_confs["version"])
```
thvasilo, line 807: If that happens we proceed to parse as a GConstruct config correct? Let's mention that.

thvasilo (approved):
LGTM now, one small comment.

Let's wait for tests to pass before merging.

PR #1231, opened 2025-04-01
Change decoder output definition so that focal loss does not require setting num_classes to 1.

thvasilo:
*Issue #, if available:*

*Description of changes:*

* We use a condition to only set the decoder output dimension to 1 for focal loss, using num_classes for other cases.

### DIFF
--- a/docs/source/advanced/imbalanced-labels.rst
+++ b/docs/source/advanced/imbalanced-labels.rst
@@ -34,7 +34,7 @@ in a binary classification. This function has two hyperparameters, :math:`\alpha
 corresponding to the ``alpha`` and ``gamma`` configuration in GraphStorm. Larger values of ``gamma`` will help
 update models on harder cases so as to detect more positive samples if the positive to negative ratio is small.
 There is no clear guideline for values of ``alpha``. You can use its default value(``0.25``) first, and then
-search for optimal values. Below is an example about how to set the `focal loss funciton` in a YAML configuration file.
+search for optimal values. Below is an example about how to set the `focal loss function` in a YAML configuration file.
 
   .. code-block:: yaml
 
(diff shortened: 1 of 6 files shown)


python/graphstorm/gsf.py
```diff
@@ -476,28 +476,34 @@ def create_builtin_node_decoder(g, decoder_input_dim, config, train_task):
...
-                                       use_bias=config.decoder_bias)
+
+            decoder_output_dim = config.num_classes
             if config.class_loss_func == BUILTIN_CLASS_LOSS_CROSS_ENTROPY:
                 loss_func = ClassifyLossFunc(config.multilabel,
                                              config.multilabel_weights,
                                              config.imbalance_class_weights)
             elif config.class_loss_func == BUILTIN_CLASS_LOSS_FOCAL:
-                assert config.num_classes == 1, \
+                assert config.num_classes in [1, 2], \
```
classicsong, line 486: Add a comment here:
`For backward compatibility, we allow the num_classes to be 1. We suggest users to set it to 2. `

python/graphstorm/gsf.py
```diff
@@ -617,8 +623,21 @@ def create_builtin_edge_decoder(g, decoder_input_dim, config, train_task):
...
     """
     dropout = config.dropout if train_task else 0
+
+
     if config.task_type == BUILTIN_TASK_EDGE_CLASSIFICATION:
         num_classes = config.num_classes
+
+        # Focal loss expects 1-dimensional output
+        if config.class_loss_func == BUILTIN_CLASS_LOSS_FOCAL:
+            assert num_classes in [1, 2], (
```
classicsong, line 633: Add a comment here:
`For backward compatibility, we allow the num_classes to be 1. We suggest users to set it to 2. `

docs/source/advanced/imbalanced-labels.rst
```diff
@@ -34,7 +34,7 @@ in a binary classification. This function has two hyperparameters, :math:`\alpha
 corresponding to the ``alpha`` and ``gamma`` configuration in GraphStorm. Larger values of ``gamma`` will help
 update models on harder cases so as to detect more positive samples if the positive to negative ratio is small.
 There is no clear guideline for values of ``alpha``. You can use its default value(``0.25``) first, and then
-search for optimal values. Below is an example about how to set the `focal loss funciton` in a YAML configuration file.
+search for optimal values. Below is an example about how to set the `focal loss function` in a YAML configuration file.
```
classicsong, line 37: Please also mention that the  num_class should be set to 2 when using focal loss.

Also add some hints in the docstring of `class_loss_func`

classicsong (approved):
LGTM

PR #1234, opened 2025-04-03
[GSProcessing] Support `:` in label column names

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Previously, label column names that included colons in their name like `label:String`, which is how Neptune Analytics exports CSV columns, would trigger an SQL syntax error because of the way we defined the label DataFrame schema using an SQL string. 
 
For example with a column like `isFraud:Int`

```
[PARSE_SYNTAX_ERROR] Syntax error at or near ':'.(line 1, pos 3)
== SQL ==
idx: int, isFraud:Int: string
---^^^
```

* This PR programmatically defines the schema to avoid this issue.
* Modify a test to match the corner case.

### DIFF
--- a/graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_label_transformation.py
+++ b/graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_label_transformation.py
@@ -18,6 +18,7 @@ import json
 from typing import Dict, Sequence
 
 from pyspark.sql import DataFrame, functions as F, SparkSession
+from pyspark.sql.types import StructType, StructField, StringType, IntegerType
 from pyspark.ml.feature import StringIndexer
 
 from .base_dist_transformation import DistributedTransformation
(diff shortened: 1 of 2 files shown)


graphstorm-processing/graphstorm_processing/data_transformations/dist_transformations/dist_label_transformation.py
```diff
@@ -75,10 +76,18 @@ def apply(self, input_df: DataFrame) -> DataFrame:
             *original_cols,
         )
 
+        # Create schema for label_df
```
jalencato, line 79: What if the label column name is not like the format "idx:Int"?

thvasilo: The `idx` column is one we create through the call to `list(enumerate(str_indexer_model.labelsArray[0]))`

jalencato (approved):
LGTM

PR #1239, opened 2025-04-04
[GSPartition] Ensure random partition assigns at least one node per partition

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Previously it was possible for random partition to create assignments that led to some node types having 0 nodes in one partition. This would eventually trigger errors in DistPart.
* Similarly, we allowed the creation of partitions for node types with fewer nodes that the requested partitions.
* This PR ensures every partition will have at least one node, and raises an error if we try to create a partition for a node type with fewer nodes than the requested partitions.
* Expand tests to cover all cases.

### DIFF
--- a/python/graphstorm/gpartition/random_partition.py
+++ b/python/graphstorm/gpartition/random_partition.py
@@ -62,11 +62,12 @@ class RandomPartitionAlgorithm(LocalPartitionAlgorithm):
                 raise RuntimeError(
                     f"Number of nodes for node type {ntype} ({num_nodes_for_type}) "
                     f"is less than the number of partitions ({num_partitions})"
+                    f"To fix the error, reduce the number of partitions or drop the '{ntype}' nodes"
                 )
 
             # Ensure at least one node per partition
             partition_assignment = np.empty(num_nodes_for_type, dtype=partition_dtype)
-            # Assign first num_partitions nodes to each one of the partitions and shuffle
+            # Assign first num_partitions nodes to each one of the partitions
             all_parts = np.arange(num_partitions)
             np.random.shuffle(all_parts)
             partition_assignment[:num_partitions] = all_parts
--- a/python/graphstorm/gpartition/random_partition.py
+++ b/python/graphstorm/gpartition/random_partition.py
@@ -45,8 +45,7 @@ class RandomPartitionAlgorithm(LocalPartitionAlgorithm):
     def _assign_partitions(self, num_partitions: int, partition_dir: str):
         num_nodes_per_type = self.metadata_dict["num_nodes_per_type"]
         ntypes = self.metadata_dict["node_type"]
-        assert len(num_nodes_per_type) == len(ntypes), \
-            "Number of nodes per type does not match number of node types"
+        # Note: This assumes that the order of node_type is the same as the order num_nodes_per_type
         assert num_partitions > 0, "Number of partitions must be greater than 0"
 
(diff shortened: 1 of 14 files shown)


python/graphstorm/gpartition/random_partition.py
```diff
@@ -45,18 +45,39 @@ class RandomPartitionAlgorithm(LocalPartitionAlgorithm):
...
-            partition_dtype = np.uint8 if num_partitions <= 256 else np.uint16
+            assert num_partitions < 2**16, \
+                f"Number of partitions must be smaller than 65536, got {num_partitions}"
+            partition_dtype = np.uint8 if num_partitions < 256 else np.uint16
+
+            if num_nodes_for_type < num_partitions:
+                # If we have fewer nodes than partitions, raise an error
+                raise RuntimeError(
+                    f"Number of nodes for node type {ntype} ({num_nodes_for_type}) "
+                    f"is less than the number of partitions ({num_partitions})"
```
classicsong, line 64:
```suggestion
                    f"is less than the number of partitions ({num_partitions})"
                    f"To fix the error, reduce the number of partitions or drop the {ntype} nodes"
```

python/graphstorm/gpartition/random_partition.py
```diff
@@ -45,18 +45,39 @@ class RandomPartitionAlgorithm(LocalPartitionAlgorithm):
...
+                )
+
+            # Ensure at least one node per partition
+            partition_assignment = np.empty(num_nodes_for_type, dtype=partition_dtype)
+            # Assign first num_partitions nodes to each one of the partitions and shuffle
```
classicsong, line 69:
```suggestion
            # Assign first num_partitions nodes to each one of the partitions
```

tests/unit-tests/gpartition/test_random_partition.py
```diff
@@ -13,11 +13,59 @@
...
+    """Test random partition when num_nodes==num_partitions"""
+    chunked_meta_dict = {
+        "num_nodes_per_type": [num_nodes_parts],
+        "node_type": ["a"],
+    }
+
+    rand_partitioner = RandomPartitionAlgorithm(chunked_meta_dict)
+    # Repeat assignment ten times to get random results
+    for _ in range(10):
+        simple_test_partition(
```
classicsong, line 48: why we what to assign ten times?

thvasilo: Explained in the comment above, we just want to be more certain that the specific random assignment just happened to have one node in each partition, so we repeat the process ten times to increase confidence.

classicsong (approved):
LGTM

python/graphstorm/gpartition/random_partition.py
```diff
@@ -45,18 +45,40 @@ class RandomPartitionAlgorithm(LocalPartitionAlgorithm):
     def _assign_partitions(self, num_partitions: int, partition_dir: str):
         num_nodes_per_type = self.metadata_dict["num_nodes_per_type"]
         ntypes = self.metadata_dict["node_type"]
-        # Note: This assumes that the order of node_type is the same as the order num_nodes_per_type
+        assert len(num_nodes_per_type) == len(ntypes), \
```
jalencato, line 48: In what case will it be non-equal? Even if there is no such node in one node type, there will be a zero here, right?

jalencato (approved):
LGTM

PR #1251, opened 2025-04-16
[Feature] Pack model artifact func for realtime inference

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
This PR add a packing function and a script to pack trained model artifacts into a `model.tar.gz` file. This function will be a part of launch script for realtime inference deployment.


python/graphstorm/inference/utils.py
```diff
@@ -0,0 +1,89 @@
+"""
+    Copyright 2023 Contributors
```
classicsong, line 2:
```suggestion
    Copyright 2025 Contributors
```

python/graphstorm/inference/utils.py
```diff
@@ -0,0 +1,89 @@
...
+import tarfile
+
+
+def wrap_model_artifacts(path_to_model, path_to_yaml, path_to_json, path_to_entry,
+                         wrap_name='model', output_path=None):
+    """ A utility function to zip model artifacts into a tar package
+
+    This function will put the entry point file into a sub-folder,named `code`, and then zip the
+    `model.bin`, the `**.yaml`, the `**.json` file, and the `code` sub-folder into a tar package
+    with the name specified by the wrap_name argument, and save it to the `output_path`.
```
classicsong, line 30: Do you want to point to a SageMaker doc about model artifacts for inference endpoints?

python/graphstorm/inference/utils.py
```diff
@@ -0,0 +1,89 @@
...
+
+    This function will check if given files exist. Non-existance will trigger an error raising.
```
classicsong, line 32:
```suggestion
    This function will check if given files exist. Non-existance will raise an error.
```

python/graphstorm/inference/utils.py
```diff
@@ -0,0 +1,89 @@
...
+    
+    Parameters
+    ----------
+    path_to_model: str
+        The path of GraphStorm's `model.bin` file.
```
classicsong, line 37: What about learnable embeddings?

python/graphstorm/inference/utils.py
```diff
@@ -0,0 +1,89 @@
...
+    path_to_yaml: str
+        The path of the YAML file generated by GraphStorm during model training.
+    path_to_json: str
+        The path of JSON file generated by GraphStorm gconstruct or GSProcessing.
+    path_to_entry: str
+        The path of the entry point file for a specific task. The file will be put into a
+        sub-folder, named 'code'.
+    wrap_name: str
```
classicsong, line 45: `wrap_name` is weird. Why not call it file_name or out_file_name.

python/graphstorm/inference/utils.py
```diff
@@ -0,0 +1,89 @@
...
+    output_path: str
+        The folder where the output tar package will be saved. If not provided, will
+        raise an error. Default is None, meaning not provided.
+
+    """
+    # check if files exist
+    assert os.path.exists(path_to_model), f'The model file, {path_to_model}, does not exist.'
+    assert os.path.isfile(path_to_model), f'The {path_to_model} should be a file path, \
+                                            but got a folder.'
+    assert os.path.exists(path_to_yaml), f'The model configuration YAML file, {path_to_yaml}, \
```
classicsong, line 56: I don't think we need to check the path_to_yaml. We already have os.path.isfile(path_to_yaml). If the file does not exist, isfile will also return False.

python/graphstorm/inference/utils.py
```diff
@@ -0,0 +1,89 @@
...
+                                           does not exist.'
+    assert os.path.isfile(path_to_yaml), f'The {path_to_yaml} should be a file path, \
+                                           but got a folder.'
+    assert os.path.exists(path_to_json), f'The graph metadata JSON file, {path_to_json}, does \
+                                           not exist.'
+    assert os.path.isfile(path_to_json), f'The {path_to_json} should be a file path, \
```
classicsong, line 62: ditto

python/graphstorm/inference/utils.py
```diff
@@ -0,0 +1,89 @@
...
+                                           but got a folder.'
+    assert os.path.exists(path_to_entry), f'The SageMaker entry point file, {path_to_entry}, does \
+                                           not exist.'
+    assert os.path.isfile(path_to_entry), f'The {path_to_entry} should be a file path, \
```
classicsong, line 66: ditto

python/graphstorm/inference/utils.py
```diff
@@ -0,0 +1,89 @@
...
+    # create the output folder if not exist.
+    assert output_path is not None, 'Output path is not provided.'
+    assert os.path.isdir(output_path), 'Output path should be a folder name, but got a ' + \
+                                       f'file {output_path}.'
+    os.makedirs(output_path, exist_ok=True)
+    os.makedirs(os.path.join(output_path, 'code'), exist_ok=True)
+
+    # copy the artifacts files to output_path
+    shutil.copy(path_to_entry, os.path.join(output_path, 'code', os.path.basename(path_to_entry)))
+    if output_path != os.path.dirname(path_to_model):
```
classicsong, line 78: output_path is a directory and path_to_model is a file path. The if statement will always be True.

Are you check whether path_to_model has already been under output_path? 
You may want to use `os.path.dirname`

python/graphstorm/inference/utils.py
```diff
@@ -0,0 +1,89 @@
...
+        shutil.copy(path_to_model, os.path.join(output_path, os.path.basename(path_to_model)))
+    if output_path != os.path.dirname(path_to_yaml):
```
classicsong, line 80: ditto

python/graphstorm/inference/utils.py
```diff
@@ -0,0 +1,89 @@
...
+        shutil.copy(path_to_yaml, os.path.join(output_path, os.path.basename(path_to_yaml)))
+    if output_path != os.path.dirname(path_to_json):
```
classicsong, line 82: ditto

tests/unit-tests/test_utils.py
```diff
@@ -1237,6 +1239,41 @@ def test_get_graph_name():
...
+    create_dummy_file(entry_path)
+    create_dummy_file(model_path)
+    create_dummy_file(yaml_path)
+    create_dummy_file(json_path)
+
+    output_path = '/tmp/output_folder/'
+    os.makedirs(output_path, exist_ok=True)
+    output_file = wrap_model_artifacts(model_path, yaml_path, json_path, entry_path,
+                                       wrap_name='model', output_path=output_path)
+    assert os.path.exists(os.path.join(output_file))
```
classicsong, line 1271: You should also test all the failure cases.

tools/pack_model_artifacts.py
```diff
@@ -0,0 +1,83 @@
+"""
+    Copyright 2023 Contributors
```
classicsong, line 2:
```suggestion
    Copyright 2025 Contributors
```

tools/pack_model_artifacts.py
```diff
@@ -0,0 +1,83 @@
...
+                        help='The file path of model configuration YAML file. By default, ' + \
+                             'GraphStorm generates a new YAML file that stores all model ' + \
+                             'configurations during a training.')
+    parser.add_argument('--graph-metadata-json-path', type=str, required=True,
+                        help='The file path of graph metadata JSON file. By default, ' + \
+                             'GraphStorm generates a new JSON file that stores all graph ' + \
+                             'metadata during a graph construction.')
+    parser.add_argument('--graphstorm-home-folder', type=str, required=True,
+                        help='The home folder path of GraphStorm source code.')
+    parser.add_argument('--gml-task', type=str, choices=SUPPORTED_TASKS, required=True,
```
classicsong, line 76:
```suggestion
    parser.add_argument('--gml-task-type', type=str, choices=SUPPORTED_TASKS, required=True,
```

tools/pack_model_artifacts.py
```diff
@@ -0,0 +1,83 @@
...
+                             '\'model.bin\' file.')
+    parser.add_argument('--model-config-yaml-path', type=str, required=True,
+                        help='The file path of model configuration YAML file. By default, ' + \
+                             'GraphStorm generates a new YAML file that stores all model ' + \
+                             'configurations during a training.')
+    parser.add_argument('--graph-metadata-json-path', type=str, required=True,
+                        help='The file path of graph metadata JSON file. By default, ' + \
+                             'GraphStorm generates a new JSON file that stores all graph ' + \
+                             'metadata during a graph construction.')
+    parser.add_argument('--graphstorm-home-folder', type=str, required=True,
```
classicsong, line 74: You assume the user clones the graphstorm repo on its local disk.
But users do not necessarily clone the repo. They only need to download the entry_point file or they may customize their own entry_point.

tools/pack_model_artifacts.py
```diff
@@ -0,0 +1,83 @@
+"""
```
classicsong, line 1: Can we move this file under python/graphstorm/sagemaker? So it can be installed with pip.
And also rename the file to pack_sm_model_artifacts.py or pack_sagemaker_model_artifacts.py.

python/graphstorm/inference/utils.py
thvasilo: The recommendation is to avoid years in copyright statements to avoid having to maintain them. Applicability is the same.

tools/pack_model_artifacts.py
thvasilo: +1 this is related to the above comment on where this file should live as well.

tools/pack_model_artifacts.py
```diff
@@ -0,0 +1,83 @@
...
+                                        output_path=args.output_folder)
+    print(f'Successfully pack GraphStorm model artifacts into {output_file}.')
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser('Package GraphStorm model artifacts.')
+    parser.add_argument('--model-file-path', type=str, required=True,
+                        help='The file path of trained GraphStorm model. By default, ' + \
+                             'GraphStorm saves model parameters into a binary file, named ' + \
+                             '\'model.bin\' file.')
```
thvasilo, line 65: The recommendation is to a pair of parentheses for line continuation, see https://google.github.io/styleguide/pyguide.html#32-line-length and https://peps.python.org/pep-0008/#maximum-line-length

```suggestion
                        help=('The file path of trained GraphStorm model. By default, '
                             'GraphStorm saves model parameters into a binary file, named '
                             '\'model.bin\' file.'))
```

thvasilo (changes requested):

python/graphstorm/inference/utils.py
zhjwy9343: Yes, will update the doc string with this link: https://sagemaker.readthedocs.io/en/stable/frameworks/pytorch/using_pytorch.html#model-directory-structure

python/graphstorm/inference/utils.py
zhjwy9343: For realtime inference, we don't support using the learnable embeddings. So, in the current implementation, not wrap the learnable embeddings.

python/graphstorm/inference/utils.py
zhjwy9343: Changed to `output_package_name`.

python/graphstorm/inference/utils.py
zhjwy9343: The error message will be different. Using `isfile` only could give a misleading error message to users.

tools/pack_model_artifacts.py
zhjwy9343: See the reply of the next comment.

tools/pack_model_artifacts.py
zhjwy9343: This script will not be used in the realtime inference pipeline. Instead, it provides users some flexibility to pack model artifacts on their own.
The same wrapping function will be used in launch endpoint script that is part of the sagemaker core code.

tests/unit-tests/test_utils.py
```diff
@@ -19,7 +19,7 @@
 import pytest
 import multiprocessing as mp
 import h5py
-
+import shutil
```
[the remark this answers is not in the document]
zhjwy9343: yes. moved.

tests/unit-tests/test_utils.py
```diff
@@ -1237,6 +1239,41 @@ def test_get_graph_name():
...
 
 
+def create_dummy_file(file_path):
+    """ Create an empty dummy file for testing
+    """
+    if os.path.exists(file_path):
+        return
+
+    with open(file_path, 'w') as f:
+        f.close()
```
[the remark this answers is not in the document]
zhjwy9343: Yes. clean the code to use tempfile.

tests/unit-tests/test_utils.py
```diff
@@ -1237,6 +1239,41 @@ def test_get_graph_name():
...
+
+def test_wrap_model_artifacts():
+    """
+    """
+    # with tempfile.TemporaryDirectory() as tmpdirname:
+    tmpdirname = '/tmp/wrap_model_artifacts/'
```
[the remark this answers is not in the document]
zhjwy9343: ditto

tests/unit-tests/test_utils.py
```diff
@@ -1237,6 +1239,41 @@ def test_get_graph_name():
...
+    create_dummy_file(json_path)
+
+    output_path = '/tmp/output_folder/'
+    os.makedirs(output_path, exist_ok=True)
+    output_file = wrap_model_artifacts(model_path, yaml_path, json_path, entry_path,
+                                       wrap_name='model', output_path=output_path)
+    assert os.path.exists(os.path.join(output_file))
+
+    shutil.rmtree(tmpdirname)
+    shutil.rmtree(output_path)
```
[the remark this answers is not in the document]
zhjwy9343: ditto

tests/unit-tests/test_utils.py
```diff
@@ -1237,6 +1239,41 @@ def test_get_graph_name():
...
+    model_path = os.path.join(tmpdirname, 'model.bin')
+    yaml_path = os.path.join(tmpdirname, 'test.yaml')
+    json_path = os.path.join(tmpdirname, 'test.json')
+
+    create_dummy_file(entry_path)
+    create_dummy_file(model_path)
+    create_dummy_file(yaml_path)
+    create_dummy_file(json_path)
+
+    output_path = '/tmp/output_folder/'
```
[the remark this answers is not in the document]
zhjwy9343: ditto

tests/unit-tests/test_utils.py
zhjwy9343: added all failure cases.

tests/unit-tests/test_utils.py
```diff
@@ -1251,27 +1252,151 @@ def create_dummy_file(file_path):
...
+        contents = tar_object.getmembers()
+        results = {content.name: content.isfile() for content in contents}
+
+        assert 'code' in results        
+        assert 'code/nc_infer_entry.py' in results and \
+               results['code/nc_infer_entry.py'] == True
+        assert 'model.bin' in results and results['model.bin'] == True
+        assert 'test.yaml' in results and results['test.yaml'] == True
+        assert 'test.json' in results and results['test.json'] == True
+        tar_object.close()
```
thvasilo, line 1284: Can use a context manager here?

```suggestion
         with  tarfile.open(output_file) as tar_object:
             contents = tar_object.getmembers()
             results = {content.name: content.isfile() for content in contents}

             assert 'code' in results        
             assert 'code/nc_infer_entry.py' in results and \
                 results['code/nc_infer_entry.py'] == True
             assert 'model.bin' in results and results['model.bin'] == True
             assert 'test.yaml' in results and results['test.yaml'] == True
             assert 'test.json' in results and results['test.json'] == True
```

tests/unit-tests/test_utils.py
```diff
@@ -1251,27 +1252,151 @@ def create_dummy_file(file_path):
...
+        json_path = os.path.join(tmpdirname, 'test.json')
+
+        output_path = os.path.join(tmpdirname, 'output_folder')
+        os.makedirs(output_path, exist_ok=True)
+
+        # not create the entry point file
+        create_dummy_file(model_path)
+        create_dummy_file(yaml_path)
+        create_dummy_file(json_path)
+        with assert_raises(AssertionError):
```
thvasilo, line 1302: Can we use pytest.raises and use `match` to ensure the assertion error is the one we expect? 

See https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

Same for the other raises checks.

tests/unit-tests/test_utils.py
```diff
@@ -1251,27 +1252,151 @@ def create_dummy_file(file_path):
...
+        create_dummy_file(model_path)
+        create_dummy_file(yaml_path)
+        # not create the test.json file.
+        with assert_raises(AssertionError):
+            wrap_model_artifacts(model_path, yaml_path, json_path, entry_path,
+                                        output_package_name='model', output_path=output_path)
+
+    #     2.2: missing output folder
+    with tempfile.TemporaryDirectory() as tmpdirname:
+        os.makedirs(tmpdirname, exist_ok=True)
```
thvasilo, line 1335: Is this needed? tempfile will create the temp directory right? Same for other occurrences of this

python/graphstorm/inference/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    Utility functions for inference
+"""
+
+import os
+import shutil
+import tarfile
+
+
+def wrap_model_artifacts(path_to_model, path_to_yaml, path_to_json, path_to_entry,
+                         output_package_name='model', output_path=None):
```
thvasilo, line 25: Let's try to avoid `package` because usually refers to an abstract collection of classes and interfaces, similar to a python module, e.g. we have [Java packages](https://docs.oracle.com/javase/tutorial/java/concepts/package.html).

Tar files are [usually referred to](https://en.wikipedia.org/wiki/Tar_(computing)) as `archives` or `tarballs`, let's use that naming.

tools/pack_model_artifacts.py
thvasilo: Is this targeted only at SageMaker then? If that's the case we can put it under the `sagemaker` directory, under an `endpoint` or `real-time-inference` path? Also see other comment about asking for entry points explicitly.

tools/pack_model_artifacts.py
```diff
@@ -0,0 +1,83 @@
...
+        path_to_entry = os.path.join(args.graphstorm_home_folder, entry_base_path,
+                                     'node_regression_entry_point.py')
+    elif args.gml_task == BUILTIN_TASK_EDGE_CLASSIFICATION:
+        path_to_entry = os.path.join(args.graphstorm_home_folder, entry_base_path,
+                                     'edge_classification_entry_point.py')
+    elif args.gml_task == BUILTIN_TASK_EDGE_REGRESSION:
+        path_to_entry = os.path.join(args.graphstorm_home_folder, entry_base_path,
+                                     'edge_regression_entry_point.py')
+    else:
+        raise NotImplementedError(f'Not supported GML task {args.gml_task}.')
```
thvasilo, line 50: Usually better if we don't have strong assumptions about file structure, e.g. `entry_base_path = 'python/graphstorm/sagemaker/deploy/'`.

We can make `--entry-point` a required argument and just ask for the full path to it, as we do with the other files. This also allows users to define their own entry points.

thvasilo (changes requested):

tools/pack_model_artifacts.py
zhjwy9343: reorg folder locations.

tests/unit-tests/test_utils.py
zhjwy9343: changed.

python/graphstorm/inference/utils.py
zhjwy9343: renamed to `output_tarfile_name`

tools/pack_model_artifacts.py
zhjwy9343: just remove this script as we don't it as a tool.

tests/unit-tests/test_utils.py
zhjwy9343: Good suggest. Changed all.

tests/unit-tests/test_utils.py
zhjwy9343: Good catch. Removed all.

tests/sagemaker-tests/test_sagemaker_utils.py
```diff
@@ -0,0 +1,193 @@
...
+        create_dummy_file(yaml_path)
+        # create a json file as a folder
+        os.remove(json_path)
+        os.makedirs(json_path)
+        with pytest.raises(AssertionError, match='graph metadata JSON file, .* got a folder'):
+            wrap_model_artifacts(model_path, yaml_path, json_path, entry_path,
+                                        output_tarfile_name='model')
+
+if __name__ == '__main__':
+    test_wrap_model_artifacts()
```
thvasilo, line 193: We use `pytest` to run tests so this is not needed

tests/unit-tests/test_utils.py
thvasilo, line 1: Are changes to this file needed any more?

sagemaker/launch/launch_utils.py
```diff
@@ -0,0 +1,91 @@
...
+    path_to_model: str
+        The path of GraphStorm's `model.bin` file.
+    path_to_yaml: str
+        The path of the YAML file generated by GraphStorm during model training.
+    path_to_json: str
+        The path of JSON file generated by GraphStorm gconstruct or GSProcessing.
+    path_to_entry: str
+        The path of the entry point file for a specific task. The file will be put into a
+        sub-folder, named 'code'.
+    output_package_name: str
```
thvasilo, line 47:
```suggestion
    output_tarfile_name: str
```

sagemaker/launch/launch_utils.py
```diff
@@ -0,0 +1,91 @@
...
+        The name of the tar package. Default is `model`.
+    output_path: str
+        The folder where the output tar package will be saved. If not provided, will
+        raise an error. Default is None, meaning not provided.
```
thvasilo, line 51: If we want to raise an error when the output is not provided, why not make the value not have a default, so that it's required to call the function?

tests/sagemaker-tests/test_sagemaker_utils.py
zhjwy9343: removed.

sagemaker/launch/launch_utils.py
zhjwy9343: remove the default and revise test cases.

tests/unit-tests/test_utils.py
zhjwy9343: No. removed the main function.

sagemaker/launch/launch_utils.py
```diff
@@ -0,0 +1,90 @@
...
+                                           not exist.'
+    assert os.path.isfile(path_to_json), f'The graph metadata JSON file, {path_to_json}, should \
+                                           be a file path, but got a folder.'
+    assert os.path.exists(path_to_entry), f'The SageMaker entry point file, {path_to_entry}, does \
+                                           not exist.'
+    assert os.path.isfile(path_to_entry), f'The SageMaker entry point file, {path_to_entry}, \
+                                           should be a file path, but got a folder.'
+
+    # create the output folder if not exist.
+    assert os.path.isdir(output_path), 'Output path should be a folder name, but got a ' + \
```
classicsong, line 72: What will happen if output_path does not exist?

sagemaker/launch/launch_utils.py
```diff
@@ -0,0 +1,90 @@
...
+                                       f'file {output_path}.'
+    os.makedirs(output_path, exist_ok=True)
+    os.makedirs(os.path.join(output_path, 'code'), exist_ok=True)
+
+    # copy the artifacts files to output_path
+    shutil.copy(path_to_entry, os.path.join(output_path, 'code', os.path.basename(path_to_entry)))
```
classicsong, line 78: We are assuming the entry point is implemented in a single file.
How to handle the case when the entry_point imports modules from other local files?

sagemaker/launch/launch_utils.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def wrap_model_artifacts(path_to_model, path_to_yaml, path_to_json, path_to_entry,
+                         output_path, output_tarfile_name='model'):
+    """ A utility function to zip model artifacts into a tar package
+
+    According to SageMaker's specification of the `Model Directory Structure
+    https://sagemaker.readthedocs.io/en/stable/frameworks/pytorch/using_pytorch.html#model-directory-structure`_,
+    this function will put the entry point file into a sub-folder,named `code`, and then zip the
+    `model.bin`, the `**.yaml`, the `**.json` file, and the `code` sub-folder into a tar package
```
thvasilo, line 31:
```suggestion
    `model.bin`, the `**.yaml`, the `**.json` file, and the `code` sub-folder into a tar archive
```

sagemaker/launch/launch_utils.py
```diff
@@ -0,0 +1,90 @@
...
+    with the name specified by the wrap_name argument, and save it to the `output_path`.
+
+    This function will check if given files exist. Non-existance will raise an error.
+    
+    Parameters
+    ----------
+    path_to_model: str
+        The path of GraphStorm's `model.bin` file.
```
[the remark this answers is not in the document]
thvasilo: Good callout here, even if we don't support sparse embeddings in v1, this function needs to eventually support them, so good to have the API in place so we don't have to make a breaking change.

tests/sagemaker-tests/test_sagemaker_utils.py
```diff
@@ -0,0 +1,193 @@
...
+        yaml_path = os.path.join(tmpdirname, 'test.yaml')
+        json_path = os.path.join(tmpdirname, 'test.json')
+
+        create_dummy_file(entry_path)
+        create_dummy_file(model_path)
+        create_dummy_file(yaml_path)
+        create_dummy_file(json_path)
+
+        output_path = os.path.join(tmpdirname, 'output_folder')
+        os.makedirs(output_path, exist_ok=True)
```
classicsong, line 48: You also need to test the case when output_path does not exist.

tests/sagemaker-tests/test_sagemaker_utils.py
```diff
@@ -0,0 +1,193 @@
...
+        entry_path = os.path.join(tmpdirname, 'nc_infer_entry.py')
+        model_path = os.path.join(tmpdirname, 'model.bin')
+        yaml_path = os.path.join(tmpdirname, 'test.yaml')
+        json_path = os.path.join(tmpdirname, 'test.json')
+
+        create_dummy_file(entry_path)
+        create_dummy_file(model_path)
+        create_dummy_file(yaml_path)
+        create_dummy_file(json_path)
+        with pytest.raises(TypeError):
```
classicsong, line 121: Why this is a TypeError?

sagemaker/launch/launch_utils.py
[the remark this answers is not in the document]
zhjwy9343: No. Sparse embedding will be saved separately in sub-folders named after node type string.

We have discussed if we want to eventually support sparse embedding. So far, the conclusion is that we will NOT support the sparse embedding for reasons like:

1. The sparse embedding could be extremely larger than model parameters. It will put a lot of burden for us to handle them in an endpoint, which normally just hosts several ML models with limited learnable parameters.
2. Support of sparse embedding means we need to have some in-memory index-searching system, e.g., OpenSearch or FeatureStore. This will significantly increase the complexity of the realtime inference solution.
3. If we incorporate an in-memory index-searching system into our solution, this adds additional dependencies to our open source solution. This also limit the flexibility for users to implement the in-memory index-searching tools.

sagemaker/launch/launch_utils.py
zhjwy9343: Add one test case.

sagemaker/launch/launch_utils.py
zhjwy9343: changed.

tests/sagemaker-tests/test_sagemaker_utils.py
zhjwy9343: add a new test case.

sagemaker/launch/launch_utils.py
zhjwy9343: The entry point file will be executed in the endpoint docker image. As designed, all functions used in entry point file MUST be included in GraphStorm core code because entry point file are part of GraphStorm.

If later we need to include more files, we can modify this function with backward compatibility, i.e., check if given a string, a list of strings, or a folder, and proceed accordingly.
1. if a string, meaning just one file, copy it to the `code/` folder.
2. if a list of strings, meaning several files, copy them to the `code/` folder.
3. if a folder, meaning all files in that folder, copy all files in that folder to the `code/` folder.

So far, we can keep this simple version, and keep used functions in GraphStorm core code.

tests/sagemaker-tests/test_sagemaker_utils.py
zhjwy9343: Because missing of required argument error is TypeError.

classicsong (approved):
LGTM

thvasilo (approved):

PR #1260, opened 2025-05-01
Add support for Knowledge graph embedding training.

classicsong:
*Issue #, if available:*

*Description of changes:*

We introduce a GSPureLearnableInputLayer that it create learnable embeddings for each node in the graph. With GSPureLearnableInputLayer, we can train a KGE model (i.e., learnable embeddings + relation embeddings only)


zhjwy9343 (approved):
LGTM, only a few string fixes.

PR #1278, opened 2025-05-28
[Docker] Real time inference Docker Build

jalencato:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/docker/sagemaker/Dockerfile.rtinference
+++ b/docker/sagemaker/Dockerfile.rtinference
@@ -1,0 +1,68 @@
+# Docker file for building a docker image for running GraphStorm code on Amazon SageMaker
+# Note: Distributed graph partition will use another docker image which will come soon.
+
+ARG DEVICE=gpu
+ARG TORCH_VERSION=2.3.0
+ARG SOURCE
+
+FROM ${SOURCE:-763104351884.dkr.ecr.us-east-1.amazonaws.com/pytorch-inference:2.3.0-gpu-py311-cu121-ubuntu20.04-sagemaker} as branch-gpu
+ENV dev_type=GPU
+ARG TORCH_VERSION
+
+# Unistall preset Pytorch packages
+RUN pip3 uninstall -y torchdata torch torchvision torchaudio
+
+# Pip install GraphStorm to install dependencies
+RUN pip3 install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple graphstorm==0.4.1
+
+# Install Pytorch to fit DGL GraphBolt requirements
+RUN pip3 install torchdata==0.9.0 pydantic
+RUN pip3 install torch==${TORCH_VERSION} torchvision torchaudio --index-url https://download.pytorch.org/whl/cu121
+
+# Install DGL GPU version
+RUN pip3 install dgl==2.5.0+cu121 -f https://data.dgl.ai/wheels-internal/torch-2.3/cu121/repo.html && rm -rf /root/.cache
+
+FROM ${SOURCE:-763104351884.dkr.ecr.us-east-1.amazonaws.com/pytorch-inference:2.3.0-cpu-py311-ubuntu20.04-sagemaker} as branch-cpu
+ENV dev_type=CPU
+ARG TORCH_VERSION
+
+# Unistall preset Pytorch packages
+RUN pip3 uninstall -y torchdata torch torchvision torchaudio
+
+# Pip install GraphStorm to install dependencies
+RUN pip3 install --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple graphstorm==0.4.1
+
+# Install Pytorch to fit DGL GraphBolt requirements
+RUN pip3 install torchdata==0.9.0 pydantic
+RUN pip3 install torch==${TORCH_VERSION} torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu
+
+# Install DGL CPU version
+RUN pip3 install dgl==2.5.0 -f https://data.dgl.ai/wheels-internal/torch-2.3/repo.html && rm -rf /root/.cache
+
+FROM branch-${DEVICE} AS final
+
+LABEL maintainer="Amazon AI Graph ML team"
+
+# Install MPI etc needed by DistDGL
+RUN apt-get update; apt-get install -y --no-install-recommends libopenmpi-dev \
+    build-essential software-properties-common; add-apt-repository ppa:ubuntu-toolchain-r/test; \
+    apt-get update; apt-get upgrade libstdc++6 -y && rm -rf /var/lib/apt/lists/*
+
+# /opt/ml and all subdirectories are utilized by SageMaker, we use the /code subdirectory to store our user code.
+# TODO(xiangsx): change to pip install when PIP package is available
+COPY code/graphstorm/ /opt/ml/code/graphstorm/
+ENV PYTHONPATH="/opt/ml/code/graphstorm/python/:${PYTHONPATH}"
+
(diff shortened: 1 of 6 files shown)


docker/sagemaker/Dockerfile.endpoint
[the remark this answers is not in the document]
jalencato: Not exactly the same the main difference is here: 
https://github.com/awslabs/graphstorm/pull/1278/files#diff-f06db24f9c598e1ac42d1372bd3b4da098a0030ddff0dcdf5c271ed93cb13fcaR12

For sagemaker-endpoint, in addition to the different base image, we also need to delete the prebuilt torch version and install it again.

docker/sagemaker/Dockerfile.endpoint
```diff
@@ -0,0 +1,75 @@
+# Docker file for building a docker image for running GraphStorm code on Amazon SageMaker
+# Note: Distributed graph partition will use another docker image which will come soon.
```
thvasilo, line 2: Remove

docker/sagemaker/Dockerfile.endpoint
```diff
@@ -0,0 +1,75 @@
...
+# Install Dependencies to fit DGL GraphBolt requirements
+RUN pip3 install \
+        torchdata==0.9.0 \
+        pydantic==2.7.1 \
+        transformers==4.47.1 \
+        ogb==1.3.6 \
+    && rm -rf /root/.cache
+
+RUN pip3 install \
+    torch==2.3 torchvision torchaudio \
```
thvasilo, line 23: Is torchvision and torchaudio needed?

docker/sagemaker/Dockerfile.endpoint
```diff
@@ -0,0 +1,75 @@
...
+    --index-url https://download.pytorch.org/whl/cpu \
+    && rm -rf /root/.cache
+
+# Install DGL CPU version
+RUN pip3 install dgl==2.5.0 -f https://data.dgl.ai/wheels-internal/torch-2.3/repo.html \
```
thvasilo, line 52: Do we want DGL version to be customizable?

docker/sagemaker/Dockerfile.endpoint
```diff
@@ -0,0 +1,75 @@
...
+# Install Dependencies to fit DGL GraphBolt requirements
+RUN pip3 install \
+        torchdata==0.9.0 \
+        pydantic==2.7.1 \
+        transformers==4.47.1 \
+        ogb==1.3.6 \
+    && rm -rf /root/.cache
+
+RUN pip3 install \
+    torch==2.3 torchvision torchaudio \
```
thvasilo, line 47: Same here, remove vision/audio if not needed

docker/sagemaker/Dockerfile.endpoint
```diff
@@ -0,0 +1,75 @@
...
+    && rm -rf /root/.cache
+
+FROM branch-${DEVICE} AS final
+
+LABEL maintainer="Amazon AI Graph ML team"
+
+# Install MPI etc needed by DistDGL
+RUN apt-get update; apt-get install -y --no-install-recommends libopenmpi-dev \
+    build-essential software-properties-common; add-apt-repository ppa:ubuntu-toolchain-r/test; \
+    apt-get update; apt-get upgrade libstdc++6 -y && rm -rf /var/lib/apt/lists/*
```
thvasilo, line 62: Are these necessary when not training?

docker/sagemaker/Dockerfile.endpoint
```diff
@@ -0,0 +1,75 @@
...
+
+# /opt/ml and all subdirectories are utilized by SageMaker, we use the /code subdirectory to store our user code.
+# TODO(xiangsx): change to pip install when PIP package is available
+COPY code/graphstorm/ /opt/ml/code/graphstorm/
+ENV PYTHONPATH="/opt/ml/code/graphstorm/python/:${PYTHONPATH}"
+RUN cp /opt/ml/code/graphstorm/sagemaker/run/* /opt/ml/code/
+
+# Download DGL source code
+RUN cd /root; git clone --branch 2.5.x --single-branch https://github.com/dmlc/dgl.git
```
thvasilo, line 71: Add ARG for dgl version?

docker/sagemaker/Dockerfile.endpoint
```diff
@@ -0,0 +1,75 @@
...
+# Install MPI etc needed by DistDGL
+RUN apt-get update; apt-get install -y --no-install-recommends libopenmpi-dev \
+    build-essential software-properties-common; add-apt-repository ppa:ubuntu-toolchain-r/test; \
+    apt-get update; apt-get upgrade libstdc++6 -y && rm -rf /var/lib/apt/lists/*
+
+# /opt/ml and all subdirectories are utilized by SageMaker, we use the /code subdirectory to store our user code.
+# TODO(xiangsx): change to pip install when PIP package is available
+COPY code/graphstorm/ /opt/ml/code/graphstorm/
+ENV PYTHONPATH="/opt/ml/code/graphstorm/python/:${PYTHONPATH}"
+RUN cp /opt/ml/code/graphstorm/sagemaker/run/* /opt/ml/code/
```
thvasilo, line 68: We may want to avoid using /opt/ml/code here, see https://github.com/awslabs/graphstorm/blob/d6a928b012c9076acbbe8867eefe7ab0468060c8/docker/sagemaker/Dockerfile.sm#L50-L53

docker/sagemaker/Dockerfile.endpoint
```diff
@@ -0,0 +1,75 @@
...
+# Uninstall preset Pytorch packages
+RUN pip3 uninstall -y torchdata torch torchvision torchaudio
+
+# Install Dependencies to fit DGL GraphBolt requirements
+RUN pip3 install \
+        torchdata==0.9.0 \
+        pydantic==2.7.1 \
+        transformers==4.47.1 \
+        ogb==1.3.6 \
+    && rm -rf /root/.cache
```
thvasilo, line 44: In the train image we pin more dependencies, should we do the same here? 

https://github.com/awslabs/graphstorm/blob/d6a928b012c9076acbbe8867eefe7ab0468060c8/docker/sagemaker/Dockerfile.sm#L31-L43

docker/sagemaker/Dockerfile.endpoint
zhjwy9343: @jalencato could test if these installations are necessary. Even though the endpoint does not use DistDGL directly, it will still use `gs.initialize()` and import DistDGL classes. So need to test before remove them

docker/sagemaker/Dockerfile.endpoint
jalencato: I try to only install use least dependencies in our endpoint image. I think we can just keep what we have now.

docker/sagemaker/Dockerfile.endpoint
jalencato: I can make it an argument, but we should expect most of users will use dgl 2.5.0.

docker/sagemaker/Dockerfile.endpoint
zhjwy9343: For inference, we don't need copy the  sagemaker/run code.

thvasilo: But we will need `sagemaker/launch/realtime_entry_points/`?

zhjwy9343: Cannot do that as the entry point should be in a specific position, i.e., `/opt/ml/model/code/`. But during deployment SageMaker will attach an external folder to it, hence cleaning any contents backed in the Docker image.

zhjwy9343 (approved):
LGTM

PR #1284, opened 2025-06-02
Save combined training config file with saved model artifacts

thvasilo:
*Issue #, if available:*

*Description of changes:*

* We now save the training config file modified with runtime arguments in the model output

### DIFF
--- a/python/graphstorm/config/argument.py
+++ b/python/graphstorm/config/argument.py
@@ -21,49 +21,48 @@ import sys
 import argparse
 import math
 import logging
+import warnings
+from typing import Any, Dict
 
 import yaml
 import torch as th
 import torch.nn.functional as F
 from dgl.distributed.constants import DEFAULT_NTYPE, DEFAULT_ETYPE
 
-from .config import BUILTIN_GNN_ENCODER
-from .config import BUILTIN_ENCODER
-from .config import SUPPORTED_BACKEND
-from .config import BUILTIN_EDGE_FEAT_MP_OPS
-from .config import (BUILTIN_LP_LOSS_FUNCTION,
-                     BUILTIN_LP_LOSS_CROSS_ENTROPY,
-                     BUILTIN_LP_LOSS_CONTRASTIVELOSS,
-                     BUILTIN_CLASS_LOSS_CROSS_ENTROPY,
-                     BUILTIN_CLASS_LOSS_FUNCTION,
-                     BUILTIN_REGRESSION_LOSS_MSE,
-                     BUILTIN_REGRESSION_LOSS_FUNCTION)
-
-from .config import BUILTIN_TASK_NODE_CLASSIFICATION
-from .config import BUILTIN_TASK_NODE_REGRESSION
-from .config import BUILTIN_TASK_EDGE_CLASSIFICATION
-from .config import BUILTIN_TASK_EDGE_REGRESSION
-from .config import (BUILTIN_TASK_LINK_PREDICTION,
-                     LINK_PREDICTION_MAJOR_EVAL_ETYPE_ALL)
-from .config import (BUILTIN_TASK_RECONSTRUCT_NODE_FEAT,
-                     BUILTIN_TASK_RECONSTRUCT_EDGE_FEAT)
-from .config import BUILTIN_GNN_NORM
-from .config import EARLY_STOP_CONSECUTIVE_INCREASE_STRATEGY
-from .config import EARLY_STOP_AVERAGE_INCREASE_STRATEGY
-from .config import GRAPHSTORM_SAGEMAKER_TASK_TRACKER
-from .config import SUPPORTED_TASK_TRACKER
-
-from .config import SUPPORTED_TASKS
-
-from .config import BUILTIN_LP_DISTMULT_DECODER
-from .config import SUPPORTED_LP_DECODER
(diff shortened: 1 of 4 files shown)


zhjwy9343 (approved):
LGTM

python/graphstorm/config/argument.py
```diff
@@ -305,6 +310,52 @@ def set_task_attributes(self, configuration):
...
+                key = attr_name[1:]
+
+                # Find the appropriate section to update
+                if 'gsf' in yaml_config:
+                    for section in yaml_config['gsf'].values():
+                        if isinstance(section, dict) and key in section:
+                            section[key] = attr_value
+                            break
+                    else:
+                        # If not found in any section, add to the first section
```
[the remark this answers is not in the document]
thvasilo: This is a bigger conversation about how to handle yaml sections and it ties to input verification as well.

As it stands we don't know which section a runtime argument belongs to, unless we introduce some structure into GSConfig.

AFAIK sections are currently ignored during yaml parsing. Since this YAML file is only meant for machine consumption (users will never see it) it won't make a big difference which section runtime args end up under.

Putting under an existing section is a backwards compatible way to update the args that does not directly affect future input verification. Creating a new section to throw every runtime arg in would mean we'd need to handle that section during input verification.

python/graphstorm/config/config.py
```diff
@@ -102,6 +102,9 @@
                         BUILTIN_LP_TRANSE_L1_DECODER,
                         BUILTIN_LP_TRANSE_L2_DECODER]
 
+# Filename constants
+COMBINED_CONFIG_FILENAME = "GRAPHSTORM_COMBINED_TRAINING_CONFIG.yaml"
```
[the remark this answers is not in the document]
jalencato: What about append the "_combined_config" after the original yaml file name?

tests/unit-tests/test_config.py
```diff
@@ -2263,19 +2263,34 @@ def test_multi_task_config():
...
-    test_node_class_info()
-    test_node_regress_info()
-    test_edge_class_info()
-    test_lp_info()
-
-    test_lm()
-    test_check_node_lm_config()
\ No newline at end of file
+
+def test_save_combined_config():
```
jalencato, line 2267: One question on the test, as our yaml config is nested in ways like:

```
gnn:
   lr: 0.05
```

Do we keep the same nested structure as the original one? Or we can have some tests on it as well.

python/graphstorm/config/argument.py
```diff
@@ -305,6 +310,52 @@ def set_task_attributes(self, configuration):
...
+
+        # Update with runtime args (all attributes starting with '_')
+        for attr_name, attr_value in vars(self).items():
+            if attr_name.startswith('_') and not attr_name.startswith('__'):
+                # Extract the key without the underscore
+                key = attr_name[1:]
+
+                # Find the appropriate section to update
+                if 'gsf' in yaml_config:
+                    for section in yaml_config['gsf'].values():
```
jalencato, line 332: What if we originally have a parameter not under gsf but override in the command? Like

```
gnn:
   lr: 0.05
```

What will the final result be here? Just want to avoid the `lr` exist in two sections, like:
```
gsf:
  lr: 0.01
gnn:
  lr: 0.05
```

python/graphstorm/config/config.py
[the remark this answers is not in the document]
thvasilo: > What about append the "_combined_config" after the original yaml file name?

That would make the naming unpredictable, downstream we need to be able to rely on a specific name.

After discussion offline I think we can go with `GRAPHSTORM_RUNTIME_UPDATE_TRAINING_CONFIG.yaml`

tests/unit-tests/test_config.py
thvasilo: If the runtime arg already existed in one of the YAML sections we just update its value. 

If the runtime argument did not already exist in the YAML, we decided to add it to its own section called `runtime` after discussion with @classicsong

python/graphstorm/config/argument.py
```diff
@@ -334,8 +334,12 @@ def _save_combined_config(self, output_path: str):
                             section[key] = attr_value
                             break
                     else:
-                        # If not found in any section, add to the first section
-                        next(iter(yaml_config['gsf'].values()))[key] = attr_value
+                        # If not found in any section, add to its own `runtime` section
```
classicsong, line 337:
```suggestion
                        # If not found in any section, add to the `runtime` section
```

tests/unit-tests/test_config.py
```diff
@@ -2268,29 +2269,62 @@ def test_save_combined_config():
...
         # Load the saved config and verify it contains the overridden value
-        with open(output_path, 'r') as f:
-            saved_config = yaml.safe_load(f)
+        with open(updated_yaml, 'r') as f:
+            updated_config = yaml.safe_load(f)
 
-        # Check that the lr value was updated
-        assert saved_config['gsf']['hyperparam']['lr'] == 0.02
+        # Check that the wd_l2norm value was added to the 'runtime' key
+        assert updated_config['gsf']['runtime']['wd_l2norm'] == 0.0001
```
classicsong, line 2330: Can you also check that when we create a GSConfig object, the gsconfig.wd_l2norm is also 0.0001?

classicsong (approved):
LGTM, minor comments.

PR #1285, opened 2025-06-16
[Feature] Add Metadata and Metadata graph classes

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
This PR added `GSGraphMetadata` and `GSMetadataGraph` classes, and simulated `GSMetadataDglGraph` and `GSMetadataDistDglGraph` class for rebuilding models in SageMaker endpoints.

This PR also added comprehensive test cases for these new APIs.


python/graphstorm/dataloading/metadata.py
```diff
@@ -0,0 +1,954 @@
...
+
+    def to_string(self):
+        """ Convert the graph metadata to a simple dictionary string.
+
+        Return
+        -------
+        str: string of the graph metadata dictionary.
+        """
+        metadata_dict = self.to_dict()
+        return str(metadata_dict)
```
thvasilo, line 344: What's the intended use of this function? The result will not be valid JSON and is not readable either.

python/graphstorm/dataloading/metadata.py
```diff
@@ -0,0 +1,954 @@
...
+        """
+        if self._efeat_dims is not None:
+            efeat_all_dims = self._efeat_dims.get(etype, None)
+        else:
+            efeat_all_dims = None
+
+        return efeat_all_dims
+
+    # output functions
+    def to_dict(self):
```
thvasilo, line 245: It seems like this serializes the class which means we need a well-defined, versioned schema to ensure we maintain backwards compatibility as new entries are added to the representation.

python/graphstorm/dataloading/metadata.py
```diff
@@ -0,0 +1,954 @@
...
+        Return
+        ------
+        dict : the dictionary of nodes and their DataView classes.
+        """
+
+        logging.warning(('This %s is a metadata graph that simulates a DGL graph without '
+            'actual graph structure, i.e., nodes and edges, but meta information, e.g., node '
+            'types, edge types, and node or edge feature names. You can use the '
+            '".edges[etype].data" or ".edges[etype].data[efeat_name]" interface '
+            'to retrieve edges\' data information.'),
```
classicsong, line 511:
```suggestion
            'to retrieve node\' data information.'),
```

python/graphstorm/dataloading/metadata.py
```diff
@@ -0,0 +1,954 @@
...
+
+        Return
+        ------
+        dict : the dictionary of nodes and their DataView classes.
+        """
+
+        logging.warning(('This %s is a metadata graph that simulates a DGL graph without '
+            'actual graph structure, i.e., nodes and edges, but meta information, e.g., node '
+            'types, edge types, and node or edge feature names. You can use the '
+            '".edges[etype].data" or ".edges[etype].data[efeat_name]" interface '
```
classicsong, line 510:
```suggestion
            '".nodes[ntype].data" or ".nodes[ntype].data[efeat_name]" interface '
```

python/graphstorm/dataloading/metadata.py
```diff
@@ -0,0 +1,954 @@
...
+
+        Return
+        ------
+        dict : the dictionary of edges and their DataView classes.
+        """
+        logging.warning(('This %s is a metadata graph that simulates a DGL graph without '
+            'actual graph structure, i.e., nodes and edges, but meta information, e.g., node '
+            'types, edge types, and node or edge feature names. You can use the '
+            '".edges[etype].data" or ".edges[etype].data[efeat_name]" interface '
+            'to retrieve edges\' data information.'),
```
classicsong, line 531:
```suggestion
            'to retrieve edge\' data information.'),
```

python/graphstorm/dataloading/metadata.py
```diff
@@ -0,0 +1,954 @@
...
+        return self._edges
+
+    def is_homogeneous(self):
+        """ Check if the metadata graph is a homogeneous one.
+        """
+        return self._graph_metadata.is_homo()
+
+    @property
+    def ndata(self):
+        """ Provide an interface for retrieving node data in a homogeneous metadata graph.
```
classicsong, line 542: What will happen if a user access ndata with a heterogeneous graph?

tests/unit-tests/test_dataloading.py
```diff
@@ -77,10 +83,474 @@
...
 from graphstorm.dataloading.utils import trim_data
+from graphstorm.dataloading.metadata import (GSGraphMetadata,
+                                             GSMetadataDglDistGraph,
+                                             config_json_sanity_check,
+                                             load_metadata_from_json)
 
-from numpy.testing import assert_equal, assert_almost_equal
-from transformers import AutoTokenizer
-from torch.utils.data import DataLoader
+def build_gcons_json_example():
```
classicsong, line 91: Can we have a separate test file, e.g., test_metagraph.py ?

tests/unit-tests/test_dataloading.py
```diff
@@ -77,10 +83,474 @@
...
+                        "type": "paper"
+                    },
+                    "relation": {
+                        "type": "has"
+                    }
+                }
+            ],
+            "is_homogeneous": "false"
+        },
+        "version": "gsprocessing-v1.0"
```
classicsong, line 550: You may also need a homogeneous graph config.

python/graphstorm/dataloading/metadata.py
```diff
@@ -0,0 +1,940 @@
...
+    Graph metadata contains general information of a graph, such as graph types, node types,
+    edge types, and features. Such information could be used to generate lightweight graph
+    instances without read in actual data, or to work as a source for retrieving.
+
+    In addtion, the graph metadata is agnoistic to graph libraries, e.g., DGL and PyG.
+
+    Parameters
+    ----------
+    gtype: str
+        The type of a graph. Options include "homogeneous" and "heterogeneous".
```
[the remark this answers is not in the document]
zhjwy9343: changed.

python/graphstorm/dataloading/metadata.py
```diff
@@ -0,0 +1,940 @@
...
+                    'edge feature name strings, and values are lists of feature dimenions, ' + \
+                    'but got {efeat_dims}.'
+            assert all(etype in self._etypes for etype in efeat_dims.keys()), 'Some edge ' + \
+                f'types in edge feature dimensions: {efeat_dims} are not in the edge type ' + \
+                f'list: {etypes}.'
+        self._efeat_dims = efeat_dims
+
+    # getters
+    def is_homo(self):
+        """ Check if the grahp metadata is for a homogeneous graph.
```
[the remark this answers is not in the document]
zhjwy9343: changed.

python/graphstorm/dataloading/metadata.py
zhjwy9343: This is for print this class only. If users print an instance of this class, the output will be `<GSGraphMetadata.*****>`. So this `to_string()` function overwrite the default one.

python/graphstorm/dataloading/metadata.py
zhjwy9343: Added a version  number of this JSON to keep track its evolution.

python/graphstorm/dataloading/metadata.py
```diff
@@ -0,0 +1,940 @@
...
+HOMO_GRAPH_TYPE = 'homogeneous'
+HETE_GRAPH_TYPE = 'heterogeneous'
+SUPPORT_GPAPH_TYPES = [HOMO_GRAPH_TYPE, HETE_GRAPH_TYPE]
+
+# ============ Graph Metadata Classes ============ #
+class GSGraphMetadata():
+    """ Metadata class for saving and retrieving graph information.
+
+    .. versionadded:: 0.5.0
+        Add ``GSGraphMetadata`` in 0.5.0 to support real-time inference on Amazon SageMaker
```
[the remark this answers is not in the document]
zhjwy9343: Yes, it can be used later on for other use cases, e.g., build models.

python/graphstorm/dataloading/metadata.py
```diff
@@ -0,0 +1,954 @@
...
+                 ntypes,
+                 etypes,
+                 nfeat_dims=None,
+                 efeat_dims=None):
+        assert gtype in SUPPORT_GPAPH_TYPES, (
+            f'Graph types can only be in {SUPPORT_GPAPH_TYPES}, but got {gtype}.')
+        self._gtype = gtype
+
+        assert (isinstance(ntypes, list) and all(isinstance(ntype, str) for ntype in ntypes)) \
+            or isinstance(ntypes, str), (
```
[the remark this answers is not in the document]
zhjwy9343: Here follows the CLIs convention for better compatibility and constant UX.

python/graphstorm/dataloading/metadata.py
```diff
@@ -0,0 +1,954 @@
...
+
+        # sanity check if node types in edge types exist in node type list
+        for src_ntype, _, dst_ntype in self._etypes:
+            assert (src_ntype in self._ntypes) and ((dst_ntype in self._ntypes)), (
+                f'Some node types {src_ntype} and {dst_ntype} do not exist in the given node type '
+                f'list: {self._ntypes}.')
+
+        if nfeat_dims is not None:
+            assert isinstance(nfeat_dims, dict) and all(isinstance(key, str) for key in \
+                nfeat_dims.keys()) and all(isinstance(val, dict) for val in \
```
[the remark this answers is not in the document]
zhjwy9343: Did. Split to two assertions.

python/graphstorm/dataloading/metadata.py
```diff
@@ -0,0 +1,954 @@
...
+    """
+    def __init__(self,
+                 graph_metadata,
+                 device='cpu'):
+        super().__init__(graph_metadata)
+        self._device = device
+
+        # check homogeneous and convert node and edge node type string to DGL's default name
+        if self._graph_metadata.is_homo():
+            assert len(self._graph_metadata.get_ntypes()) == 1 and \
```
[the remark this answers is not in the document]
zhjwy9343: yes, removed.

tests/unit-tests/test_dataloading.py
zhjwy9343: Separate new tests into a `test_metadata.py`.

tests/unit-tests/test_dataloading.py
zhjwy9343: added.

python/graphstorm/dataloading/metadata.py
```diff
@@ -113,39 +113,43 @@ def __init__(self,
...
+            assert all(ntype in self._ntypes for ntype in nfeat_dims.keys()), ('Some node ' \
+                f'types in node feature dimensions: {nfeat_dims} are not in the node type ' \
+                f'list: {ntypes}.')
         self._nfeat_dims = nfeat_dims
 
         if efeat_dims is not None:
             assert isinstance(efeat_dims, dict) and all(isinstance(key, tuple) for key in \
-                efeat_dims.keys()) and all(len(key)==3 for key in efeat_dims.keys()) and \
-                all(isinstance(val, dict) for val in efeat_dims.values()) and all( \
+                efeat_dims.keys()) and all(len(key)==3 for key in efeat_dims.keys()), ( \
```
classicsong, line 133: Can you further break the assertion?
If this check fails, users have to figure out whether it is because 1/ The edge feature dimensions should be a dictionary or 2/ keys are canonical edge types.

tests/unit-tests/test_metadata.py
```diff
@@ -0,0 +1,1452 @@
...
+    with pytest.raises(AssertionError, match='A \"feature_dim\" field must be'):
+        config_json_sanity_check(gcont_config_json)
+
+    gcont_config_json = build_gcons_json_example()
+    gcont_config_json['edges'][2]['features'][0]['feature_dim'] = 16
+    with pytest.raises(AssertionError, match='Values of \"feature_dim\" field must be a list'):
+        config_json_sanity_check(gcont_config_json)
+
+    gcont_config_json = build_gcons_json_example()
+    gcont_config_json['edges'][-1]['relation'] = gcont_config_json['edges'][0]['relation']
```
classicsong, line 992: Please add a comment that "The src and dst type are same for [-1] and [0]"

tests/unit-tests/test_metadata.py
```diff
@@ -0,0 +1,1452 @@
...
+        assert md_dist_g.canonical_etypes == expected_etypes
+        assert md_dist_g.device() == 'cpu'
+        assert md_dist_g.is_homogeneous()
+        assert all(list(md_dist_g.nodes[ntype].data[feat_name].shape) == [0, dim[0]] for \
+            ntype, feat_dims in expected_nfeat_dims.items() for feat_name, dim in feat_dims.items())
+        assert all(list(md_dist_g.edges[etype].data[feat_name].shape) == [0, dim[0]] for \
+            etype, feat_dims in expected_efeat_dims.items() for feat_name, dim in feat_dims.items())
+
+
+if __name__ == '__main__':
```
classicsong, line 1447: remove __main__

python/graphstorm/dataloading/metadata.py
```diff
@@ -0,0 +1,958 @@
...
+class GSGraphFromMetadata(ABC):
+    """ Abstract class as the base of graphs built from metadata.
+
+    .. versionadded:: 0.5.0
+        Add ``GSGraphFromMetadata`` in 0.5.0 to support lightweight graph construction using
+        graph metadata only.
+
+    """
+    def __init__(self, graph_metadata):
+        self._graph_metadata = graph_metadata
```
thvasilo, line 367: Can we annotate the expected type of graph_metadata

python/graphstorm/dataloading/metadata.py
thvasilo: Without knowledge of the consumers it's harder to evaluate this PR. If we control all instantiations of this object we can restrict its API to the minimum possible.

Is there a downstream situation where a consumer instantiates this object with a heterog graph? If yes, how does that consumer handle the case where it gets None as a return value?

For this kind of problem it's much easier to use type annotations (`def ndata(self) -> Optional[return_type]:`) . Consuming code will then be flagged if the caller tries to use the returned value without checking if it's None or not.

thvasilo: > Although GS does not call this function, as a general API simulation, it is better to provide this API.

In this context who is instantiating this object? Its intended use is GraphStorm-internal right? Why not keep the API to the absolute minimum that GraphStorm needs to function?

python/graphstorm/dataloading/metadata.py
```diff
@@ -0,0 +1,958 @@
...
+        construction using graph metadata only.
+
+    This metadata graph only provides two DGL distributed graph APIs, i.e.,
+
+    - get_node_partition_policy(ntype)
+    - get_partition_book()
+
+    However, as DGL distributed graph requires actual graph structure data, which the metadata
+    cannot provide. Therefore, this metadata graph will raise NotImplementedError to let
+    callers know this metadata graph cannot be used for DGL distributed graph.
```
thvasilo, line 611: Can we clarify the need for this class? It will provide the APIs that `GSDglGraphFromMetadata` provides, but why do we add two functions just to raise NotImplemented errors? Isn't it better to just not provide the functions?

tests/unit-tests/test_metadata.py
```diff
@@ -0,0 +1,1452 @@
...
+
+# from data_utils import (
+#     generate_dummy_dist_graph,
+#     generate_special_dummy_dist_graph_for_efeat_gnn,
+#     generate_dummy_dist_graph_reconstruct,
+#     generate_dummy_dist_graph_homogeneous_failure_graph,
+#     generate_dummy_dist_graph_multi_task,
+#     create_distill_data,
+#     SIZE_DICT
+# )
```
thvasilo, line 34: Remove?

tests/unit-tests/test_metadata.py
thvasilo, line 1: We have tens of assertions in this file but how many are checking logic that's more complex that Python language features? e.g. if my function is 

```python
def is_string(input):
    if isinstance(input, str)
        return true
    else:
        raise ValueError("My error message")
````

I'm not sure how much value there is in testing every possible variation. If the type checking is the same everywhere we can test a couple of values and rely on the language for the rest, I'm concerned about having to maintain all these assertions.

**Review by thvasilo — CHANGES_REQUESTED** [review]

File: python/graphstorm/dataloading/metadata.py (hunk 19 above)
Comment (zhjwy9343) [reply python/graphstorm/dataloading/metadata.py:611]: We dedicatedly added these two APIs to provide users hints that the `GSDglDistGraphFromMetedata` cannot rebuild models with learnable embeddings in the current version.

File: tests/unit-tests/test_metadata.py (hunk 20 above)
Comment (zhjwy9343) [reply tests/unit-tests/test_metadata.py:34]: removed.

File: python/graphstorm/dataloading/metadata.py [hunk 21]
```diff
@@ -113,39 +113,43 @@ def __init__(self,
...
-                val.values()), ('The node feature dimensions should be a dictionary, whose '
-                    'keys are node type strings, and values are dictionaries whose keys are '
-                    'node feature name strings, and values are lists of feature dimensions, '
-                    f'but got {nfeat_dims}.')
-            assert all(ntype in self._ntypes for ntype in nfeat_dims.keys()), 'Some node ' + \
-                f'types in node feature dimensions: {nfeat_dims} are not in the node type ' + \
-                f'list: {ntypes}.'
+                nfeat_dims.keys()), ('The node feature dimensions should be a dictionary, ' \
+                    f'whose keys are node type strings, but got {nfeat_dims}.')
+            assert all(isinstance(val, dict) for val in nfeat_dims.values()) and \
```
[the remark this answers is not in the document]
zhjwy9343: revised, split all checks into smaller ones

tests/unit-tests/test_metadata.py
zhjwy9343: removed

tests/unit-tests/test_metadata.py
zhjwy9343: Each assertion is for a specific JSON field. In order to sanity check of the formats of the gconstruction and gsprocessing JSON to avoid errors, we need to check all required fields.

python/graphstorm/dataloading/metadata.py
```diff
@@ -112,15 +112,20 @@ def __init__(self,
...
-                in val.keys()) and all(isinstance(val_val, list) for val in nfeat_dims.values() \
-                for val_val in val.values()), ('The node feature dimension dictionary\'s ' \
-                    'values must have a string as the dictionary key, and a list of feature ' \
-                    f'dimensions as the value, but got {nfeat_dims}.')
+            assert isinstance(nfeat_dims, dict), ('The node feature dimensions should be a ' \
+                    f'dictionary, but got {nfeat_dims}.')
+            for ntype, feat_dims in nfeat_dims.items():
+                assert isinstance(ntype, str), ('The key of node feature dimensions should be ' \
+                    f'a string, but got {ntype}.')
+                assert isinstance(feat_dims, dict), ('The value of node feature dimensions ' \
```
classicsong, line 120:
```suggestion
                assert isinstance(feat_dims, dict), (f'The value of node feature dimension of  {ntype}' \
```

python/graphstorm/dataloading/metadata.py
```diff
@@ -112,15 +112,20 @@ def __init__(self,
...
+                    f'should be a dictionary, but got {ntype}.')
```
classicsong, line 121:
```suggestion
                    f'should be a dictionary, but got {feat_dims}.')
```

python/graphstorm/dataloading/metadata.py
```diff
@@ -129,18 +134,22 @@ def __init__(self,
...
-                val.keys()) and all(isinstance(val_val, list) for val in efeat_dims.values() \
-                for val_val in val.values()), ('The edge feature dimension dictionary\'s ' \
-                    'values must have a string as the key of dictionary, and a list of ' \
-                    f'feature dimensions as the value, but got {efeat_dims}.')
+            assert isinstance(efeat_dims, dict), ('The edge feature dimensions should be a ' \
+                f'dictionary, but got {efeat_dims}.')
+            for etype, feat_dims in efeat_dims.items():
+                assert isinstance(etype, tuple) and len(etype)==3, ('The edge feature ' \
+                    'dimension dictionary\'s key should be a tuple, which includes three ' \
+                    'strings to indicate source node type, edge type, destination node type), ' \
```
classicsong, line 142:
```suggestion
                    'strings to indicate source node type, relation type, destination node type), ' \
```

python/graphstorm/dataloading/metadata.py
```diff
@@ -129,18 +134,22 @@ def __init__(self,
...
+                    f'but got {etype}.')
+                assert isinstance(feat_dims, dict), ('The edge feature dimension ' \
```
classicsong, line 144: ditto

python/graphstorm/dataloading/metadata.py
```diff
@@ -129,18 +134,22 @@ def __init__(self,
...
+                    f'dictionary\'s value should be a dictionary, but got {feat_dims}.')
+                for feat_name, dim_list in feat_dims.items():
+                    assert isinstance(feat_name, str), ('The edge feature dimension ' \
+                        'dictionary\'s values should be dictionaries, whose keys are ' \
```
classicsong, line 148:
```suggestion
                        object is a dictionary, whose keys are ' \
```

python/graphstorm/dataloading/metadata.py
```diff
@@ -129,18 +134,22 @@ def __init__(self,
...
+                        f'strings, but got the key {feat_name} for the {etype} node.')
+                    assert isinstance(dim_list, list), ('The edge feature dimension ' \
+                        'dictionary\'s values should be dictionaries, whose values are lists, ' \
```
classicsong, line 151: ditto

classicsong (approved):
Please fix my comments

PR #1298, opened 2025-07-07
[GSProcessing] Solve local model cache issue when running on EMR

jalencato:
*Issue #, if available:*

*Description of changes:*


thvasilo (approved):
LGTM, reasonable way to be explicit about first trying to load models from cache, and trying a fallback if that fails.

jalencato:
Will test on large dataset again on emr-serverless and then merge it.

PR #1163, opened 2025-02-13
Save relation_embeddings for models trained with link prediction in gs_gen_node_embedding

classicsong:
*Issue #, if available:*
#1160 

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1299, opened 2025-07-08
[BugFix] Fix a bug in remapping node embeddings, when shared fs is not avaliable

classicsong:
*Issue #, if available:*
The bug was introduced in https://github.com/awslabs/graphstorm/pull/1163, where it saves relation2id_map.json into the emb output path. The remap script will mistakenly load the .json file.

*Description of changes:*
Fix the bug

### DIFF
--- a/python/graphstorm/gconstruct/remap_result.py
+++ b/python/graphstorm/gconstruct/remap_result.py
@@ -807,11 +807,11 @@ def main(args, gs_config_args):
             #     task_id1/
             #     ...
             # We need to exclude both emb_info.json, relation2id_map.json
-            # and task_id directories, when we are collecting node types
-            # with node embeddings.
+            # rel_emb.pt and task_id directories, when we are collecting 
+            # node types with node embeddings.
             emb_names = [e_name for e_name in emb_names \
                 if e_name not in task_emb_dirs + \
-                    ["emb_info.json", "relation2id_map.json"]]
+                    ["emb_info.json", "relation2id_map.json", "rel_emb.pt"]]
 
             emb_ntypes = emb_names
     else:
--- a/python/graphstorm/gconstruct/remap_result.py
+++ b/python/graphstorm/gconstruct/remap_result.py
@@ -806,12 +806,14 @@ def main(args, gs_config_args):
             #     task_id0/
             #     task_id1/
             #     ...
-            # We need to exclude both emb_info.json, relation2id_map.json
-            # rel_emb.pt and task_id directories, when we are collecting
+            # We need to exclude task_id directories, when we are collecting
             # node types with node embeddings.
             emb_names = [e_name for e_name in emb_names \
                 if e_name not in task_emb_dirs + \
                     ["emb_info.json", "relation2id_map.json", "rel_emb.pt"]]
+            # embeddings are under <node-type> directories, exclude other files
+            emb_names = [e_name for e_name in emb_names \
+                if os.path.isdir(os.path.join(node_emb_dir, e_name))]
 
             emb_ntypes = emb_names
     else:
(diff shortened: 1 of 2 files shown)


python/graphstorm/gconstruct/remap_result.py
```diff
@@ -806,10 +806,12 @@ def main(args, gs_config_args):
...
             #     ...
-            # We need to exclude both emb_info.json and task_id directories,
-            # when we are collecting node types with node embeddings.
+            # We need to exclude both emb_info.json, relation2id_map.json
+            # rel_emb.pt and task_id directories, when we are collecting
+            # node types with node embeddings.
             emb_names = [e_name for e_name in emb_names \
-                if e_name not in task_emb_dirs + ["emb_info.json"]]
+                if e_name not in task_emb_dirs + \
+                    ["emb_info.json", "relation2id_map.json", "rel_emb.pt"]]
```
thvasilo, line 814: Do we have a standardized file name  regex for embeddings we could use ?  

We need a method that includes the correct files and only those, without us needing to update the code.

tests/end2end-tests/data_process/test.sh
```diff
@@ -353,6 +353,11 @@ rm -fr /tmp/np_remap/
 echo "********* Test the remap node emb/partial emb *********"
 python3 $GS_HOME/tests/end2end-tests/data_process/gen_emb_predict_remap_test.py --output /tmp/em_remap/
 
+# Add relation2id_map.json and rel_emb.pt into /tmp/em_remap/partial-emb/
+# remap_result should ignore this file
+echo 'Dummy' > /tmp/em_remap/partial-emb/relation2id_map.json
+echo 'Dummy' > /tmp/em_remap/partial-emb/rel_emb.pt
```
thvasilo, line 359: If we just want files to be created we can use
```suggestion
touch /tmp/em_remap/partial-emb/relation2id_map.json
touch /tmp/em_remap/partial-emb/rel_emb.pt
```

tests/end2end-tests/data_process/test.sh
```diff
@@ -447,6 +452,13 @@ cp -r /tmp/em_remap/partial-emb/n0/*1.pt /tmp/em_remap/partial-emb/1/n0/
 cp -r /tmp/em_remap/partial-emb/n1/*0.pt /tmp/em_remap/partial-emb/0/n1/
 cp -r /tmp/em_remap/partial-emb/n1/*1.pt /tmp/em_remap/partial-emb/1/n1/
 
+# Add relation2id_map.json and into /tmp/em_remap/partial-emb/
+# remap_result should ignore this file
+echo 'Dummy' > /tmp/em_remap/partial-emb/0/rel_emb.pt
+echo 'Dummy' > /tmp/em_remap/partial-emb/1/rel_emb.pt
+echo 'Dummy' > /tmp/em_remap/partial-emb/0/rel_emb.pt
+echo 'Dummy' > /tmp/em_remap/partial-emb/1/rel_emb.pt
```
thvasilo, line 460: Lines 1-3 and 2-4 are identical and simply replace the contents of the existing file. We can use
```suggestion
for i in 0 1; do
  touch /tmp/em_remap/partial-emb/$i/rel_emb.pt
done
```

python/graphstorm/gconstruct/remap_result.py
thvasilo: After offline discussion let's ensure the retrieved path is a directory. We can strengthen the guarantee that every directory is a node type name later

thvasilo (approved):
LGTM, this should ensure we only take embedding dirs into consideration when collecting node types. We can further strengthen with checking directory contents.

PR #1306, opened 2025-07-15
[GSProcessing] Fix docker script&doc

jalencato:
*Issue #, if available:*

*Description of changes:*

Fix GSProcessing docker push script:
* Add version validation for both build&push script.
* Fix the region issues, now users can either set up the aws config region or pass the region as an argument with `-r`.

### DIFF
--- a/graphstorm-processing/docker/push_gsprocessing_image.sh
+++ b/graphstorm-processing/docker/push_gsprocessing_image.sh
@@ -103,7 +103,7 @@ parse_params "${@}"
 
 # Add version validation
 if [[ ! $VERSION =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
-    echo "Error: Invalid determine valid version number, got '$VERSION'"
+    echo "Error: Invalid version number, got '$VERSION'. Set desired version with --version argument"
     exit 1
 fi
 
--- a/graphstorm-processing/docker/push_gsprocessing_image.sh
+++ b/graphstorm-processing/docker/push_gsprocessing_image.sh
@@ -48,6 +48,7 @@ parse_params() {
   ARCH='x86_64'
   SUFFIX=""
 
+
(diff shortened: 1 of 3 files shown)


thvasilo:
> Current push script should be run under graphstorm/graphstorm-processing/docker folder.

Is there a reason for this change ? The script is designed to run from anywhere regardless of working directory.

graphstorm-processing/docker/push_gsprocessing_image.sh
```diff
@@ -43,12 +43,11 @@ parse_params() {
   VERSION=$(find "$SCRIPT_DIR" -maxdepth 1 -type d | sort --version-sort | tail -1 | xargs basename)
   LATEST_VERSION=${VERSION}
   # If the user did not provide a region, try to get from config
-  REGION=${REGION:-$(aws configure get region)}
+  REGION=$(aws configure get region) || REGION=""
```
thvasilo, line 46: What happens downstream when the REGION is set to the empty string ?


I guess the point is that we want to make the empty var assertion to trigger?

jalencato:
> > Current push script should be run under graphstorm/graphstorm-processing/docker folder.
> 
> Is there a reason for this change ? The script is designed to run from anywhere regardless of working directory.

If not in the right directory, the `VERSION=$(find "$SCRIPT_DIR" -maxdepth 1 -type d | sort --version-sort | tail -1 | xargs basename)` will load unexpected value.

graphstorm-processing/docker/push_gsprocessing_image.sh
jalencato: If not passing `-r` argument, and the not set the aws configure get region, then it will throw an error later to ask for passing the `-r` value.

thvasilo (approved):
LGTM now small comment

PR #1305, opened 2025-07-12
[Feature] Real-time inference for Node Prediction Tasks

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
This PR provided real-time inference support for node classification/regression tasks where trained models are work on no-learnable-embedding setting. Specifically, this PR includes:

1. A new Docker file for real-time inference based on SageMaker PyTorch inference base images, and corresponding changes on image building and pushing.
2. A set of new functions to update configurations to be used for SageMaker endpoint building, including:
> - Save runtime configurations used in training into a new YAML file at the same folder defined by `save_model_path` configuration .
> - Save the feature transformation results in graph construction and gsprocessing into a new JSON file at the same location of output of graph construction and gsprocessing.
> - update GSConfig to support the two new configuration files.
3. For real-time inference code, create a set of new classes and functions in GraphStorm core code, including:
> - `GSgnnRealtimeInferNodeDataLoader` that processes real-time payload for one-batch inference.
> - `GSGraphMetadata`, `GSGraphFromMetadata`, `GSDglGraphFromMetadata`, and `GSDglDistGraphFromMetadata` that  function as a set of lightweight methods to restore DGL distributed graph based on graph metadata only.
> - `GSGnnNodePredictionRealtimeInferrer` that conducts the prediction tasks on endpoints.
> - `GSRealTimeInferenceResponseMessage` that defines the response status code and message in an explicit way for users.
> - `restore_builtin_model_from_artifacts` function that restore a GraphStorm built-in model for endpoint model loading.
4. A helper script  `launch/launch_realtime_endpoint.py` under `graphstorm/sagemaker` package that can automatically pack model artifacts, upload to S3 bucket, and deploy SageMake endpoint based on given model artifacts. For custom model package, provides a helper script `launch_custom_model_realtime_endpoint.py` under `graphstorm/tools/` folder.
5. A built-in node prediction entry point file, `sagemaker/run/realtime_entry_points/node_prediction_entry.py`, which will be used as one model artifacts for endpoint deployment.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1288, opened 2025-06-25
Copy over GConstruct config from part_config dir to model output if one exists

thvasilo:
*Issue #, if available:*

*Description of changes:*

* We add copying of the GConstruct config if one exists during GSConfig creation

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1307, opened 2025-07-23
Forward graph construction config when input is in chunked format, GSProcessing output

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Holding for merging of https://github.com/awslabs/graphstorm/pull/1305 will change base branch after
* Adds forwarding of graph construction file during SageMaker DistPartition job and local GSPartition runs
* This should ensure the required graph construction file is present when the data have been processed with GSProcessing, as a follow-up to https://github.com/awslabs/graphstorm/pull/1288


sagemaker/launch/launch_partition.py
```diff
@@ -119,14 +120,18 @@ def get_partition_parser():
 
 if __name__ == "__main__":
     args = get_partition_parser()
-    print(args)
```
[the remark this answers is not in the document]
thvasilo: All GS args are printed. The launch args are given by the user so I don't think we need to print them? We could add them as debug logs perhaps

docs/source/cli/graph-construction/distributed/gspartition/sagemaker.rst
```diff
@@ -9,77 +9,8 @@ Build the Docker Image for GSPartition Jobs on Amazon SageMaker
...
-
-And then use the following command to push the built GraphStorm Docker image to users' own ECR repository.
-
-.. code-block:: bash
-
-    docker push <IMAGE_NAME>:<IMAGE_TAG>
-
-Please replace the `<IMAGE_NAME>` and `<IMAGE_TAG>` with the actual Docker image name and tag, e.g., ``888888888888.dkr.ecr.us-east-1.amazonaws.com/graphstorm:sm-gpu``.
+To build and push the GraphStorm SageMaker image follow the instructions
+in :ref:`Setup GraphStorm SageMaker Docker Image<build_sagemaker_docker>`.
```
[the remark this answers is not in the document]
thvasilo: As you will see the original instructions were out of date. I'm trying to keep a single source of information to avoid maintenance burden of documentation.

python/graphstorm/sagemaker/s3_utils.py
```diff
@@ -3,53 +3,101 @@
...
-        sagemaker_session = sagemaker.Session(boto_session)
+        sagemaker_session = Session(boto_session)
     try:
         ret = S3Uploader.upload(local_data_path, s3_data_path,
             sagemaker_session=sagemaker_session)
     except Exception as e: # pylint: disable=broad-except
         raise RuntimeError(f"Error uploading data to {s3_data_path}: {e}")
     return ret
+
+def list_s3_files(s3_file_path, sagemaker_session=None) -> List[str]:
```
thvasilo, line 59: We are no longer using this function but it's a good to have. I would suggest leaving it here for future use.

python/graphstorm/sagemaker/sagemaker_partition.py
```diff
@@ -386,6 +392,75 @@ def data_dispatch_step(partition_dir):
...
+                local_gsp_config,
+                sagemaker_session
+            )
+            logging.info(
+                "Uploaded runtime GSProcessing config %s to %s",
+                local_gsp_config,
+                s3_dglgraph_output
+            )
+        else:
+            warnings.warn("No runtime GSProcessing configuration file found under "
```
zhjwy9343, line 459: This if condition is for launch argument. Should this warning is about "No runtime launch configuration file found ..."?

sagemaker/run/partition_entry.py
```diff
@@ -34,7 +34,8 @@ def partition_arg_parser():
         help="S3 location to store the partitioned graph",
         required=True)
     parser.add_argument("--metadata-filename", type=str,
-        default="metadata.json", help="file name of metadata config file")
+        default="updated_row_counts_metadata.json",
+        help="file name of input 'chunked' DGL metadata config file")
```
zhjwy9343, line 38:
```suggestion
        help="The file name of input 'chunked' DGL metadata config file")
```

python/graphstorm/gpartition/dist_partition_graph.py
```diff
@@ -195,20 +197,32 @@ def main():
...
+        )
+        gsp_config_fpath = os.path.join(args.input_path, gsprocessing_config_fname)
+
+        # Copy GSProcessing runtime-updated config to output path with new name so it can be used
+        # during inference
+        dst_config = os.path.join(output_path, 'dist_graph', GS_RUNTIME_GCONSTRUCT_FILENAME)
+        logging.info("Copying GSProcessing config from %s to %s", gsp_config_fpath, dst_config)
+        shutil.copy2(gsp_config_fpath, dst_config)
+
+        # Optional hard negative mapping
```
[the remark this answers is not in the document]
thvasilo: The condition is determined inside the function. @jalencato owns this code, we could pull out the condition here as well to avoid calling the function entirely (which I'd be fine with but likely belongs in a separate PR)

python/graphstorm/sagemaker/sagemaker_partition.py
```diff
@@ -386,6 +392,75 @@ def data_dispatch_step(partition_dir):
...
 
+        # Leader instance copies raw_id_mappings from input to dist_graph output on S3
+        # using aws cli, usually much faster than using boto3
+        # Will also upload the updated GConstruct/GSProcessing config, if one exists
+        raw_id_mappings_s3_path = os.path.join(graph_data_s3, "raw_id_mappings")
+        logging.info("Copying raw_id_mappings from %s to %s",
+            raw_id_mappings_s3_path, s3_dglgraph_output)
+        # Copy raw_id_mappings from input to dist_graph output on S3
+        subprocess.call([
+            "aws", "configure", "set", "default.s3.max_concurrent_requests", "150"])
```
classicsong, line 403: How do we get 150 here? Is this value instance dependent?

python/graphstorm/sagemaker/sagemaker_partition.py
```diff
@@ -386,6 +392,75 @@ def data_dispatch_step(partition_dir):
...
+                f"{s3_dglgraph_output}/raw_id_mappings"])
+
+        # Try to download the GSProcessing launch arguments JSON
+        launch_arguments_s3_path = os.path.join(graph_data_s3_no_trailing, "launch_arguments.json")
+        download_data_from_s3(
+            launch_arguments_s3_path,
+            graph_data_path,
+            sagemaker_session
+        )
+        launch_arguments_local = os.path.join(graph_data_path, "launch_arguments.json")
```
classicsong, line 422: launch_arguments of GSProcessing?

docs/source/cli/graph-construction/distributed/example.rst
```diff
@@ -243,12 +243,13 @@ the graph structure, features, and labels. In more detail:
 
 * ``gsprocessing-config_with_transformations.json``: This is the input configuration
   we used, modified to include representations of any supported transformations
-  we applied. This file can be used to re-apply the transformations on new data.
+  we applied. This file can be used to re-apply the transformations on new data,
+  for example during online inference.
```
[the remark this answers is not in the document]
jalencato: I previously thought it was about rerunning the GSProcessing on the data again. Yeah, we can resolve the issue here.

python/graphstorm/sagemaker/sagemaker_partition.py
thvasilo: Around 100 is a good value when looking for parallel download capacity. We can select a safer value but for all instances I've used this is an OK value. The max value will depend on a number of factors but 50-150 should be safe for most instance types

python/graphstorm/sagemaker/sagemaker_partition.py
thvasilo: Yes, this is how GSprocessing itself will name the file

python/graphstorm/gpartition/post_hard_negative.py
```diff
@@ -80,6 +80,10 @@ def shuffle_hard_negative_nids(gsprocessing_config: str,
     """
     shuffled_edge_config = load_hard_negative_config(gsprocessing_config)
 
+    # Do not try to shuffle if no edge types with hard negatives exist
```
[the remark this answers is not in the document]
thvasilo: It's fixing a shaky assumption. Previously the function looked like this: https://github.com/awslabs/graphstorm/blob/02901a129d6d0c01c73e80d4331da618f3c2bd20/python/graphstorm/gpartition/post_hard_negative.py#L67

In that version when `shuffled_edge_config = load_hard_negative_config(gsprocessing_config)` returned an empty `shuffled_edge_config` the function would still iterate over num_parts in 

```
# iterate all the partitions to convert hard negative node ids.
    for i in range(num_parts):
        part_path = os.path.join(f"{graph_path}/dist_graph", f"part{i}")
        edge_feat_path = os.path.join(part_path, "edge_feat.dgl")

        # load edge features first
        edge_feats = load_tensors(edge_feat_path)
        for single_shuffled_edge_config in shuffled_edge_config:
             # We never got here because shuffled_edge_config was empty
```

but 1. we assumed `edge_feat_path = os.path.join(part_path, "edge_feat.dgl")` must exist, which is fine for real data, but in testing it's causing issues, and more importantly 2. when `shuffled_edge_config` is empty, there's no reason to load edge features in `edge_feats = load_tensors(edge_feat_path)` just to do nothing to them.

So if there's no work to be done we simply return.

Note @jalencato for visibility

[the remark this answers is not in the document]
classicsong: Maybe you can make the comment here more clear.

PR #1315, opened 2025-08-05
[docker] [sagemaker] Fix order of installation to ensure torhc is not pulled twice

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Previously we'd try to install torchdata and trasformers _before_ installing torch. These both depend on torch, leading to the GPU version + all cuda dependencies being pulled into the image.
* We move the torch installation first, avoiding the double-pulling and significantly reducing image size for CPU.


jalencato (approved):
Pass the test. LGTM

PR #1319, opened 2025-08-06
Add exception handling when trying to save runtime-updated configuration files

thvasilo:
*Issue #, if available:*

*Description of changes:*

* Sometime we might try to instantiate a GSConfig object that includes a save_model_path that is not writable (e.g. deploying an endpoint with a config from a previous train job).
* Previously that would cause the process to fail, now we add exception handling to allow graceful handling of non-writable save_model_path

### DIFF
--- a/python/graphstorm/config/argument.py
+++ b/python/graphstorm/config/argument.py
@@ -22,6 +22,7 @@ import math
 import os
 import shutil
 import sys
+import tempfile
 import warnings
 from typing import Any, Dict
 
@@ -181,7 +182,7 @@ class GSConfig:
         """ Construct a GSConfig object.
 
         Parameters:
-        ------------
+        -----------
         cmd_args: Arguments
             Commend line arguments.
         """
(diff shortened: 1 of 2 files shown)


python/graphstorm/config/argument.py
```diff
@@ -230,8 +231,18 @@ def __init__(self, cmd_args):
         # If model output is configured, save the runtime train config as a yaml file there,
         # and the graph construction config, if one exists in the input
         if hasattr(self, "_save_model_path") and self._save_model_path:
-            # Ensure model output directory exists
-            os.makedirs(self._save_model_path, exist_ok=True)
+            try:
```
zhjwy9343, line 234: Do we need to handle the writable error and save configuration errors separately? And do we need to check the `OSError` and `PermissionError` in such fine-grained level?
How about following Python Exception handle guideline: https://docs.python.org/3/tutorial/errors.html#handling-exceptions? 
Use one `try->execept` to enclose both `os.makdir` and `self._save_graph_construct_config`, and then handle each known exception one by one.

thvasilo: Right, what we can do is enclose all permissions/writable checks in one function that checks for both errors, and if that one passes, we can proceed with writing the files if they exist. 

That should simplify the work flow.

python/graphstorm/config/argument.py
```diff
@@ -230,43 +231,90 @@ def __init__(self, cmd_args):
...
+                            f"Failed to copy {GS_RUNTIME_GCONSTRUCT_FILENAME} to model output: "
+                            f"{str(e)}. You will need to manually copy over the graph "
+                            "construction config for model deployment.")
                 else:
                     warnings.warn(
                         f"Graph construction config {GS_RUNTIME_GCONSTRUCT_FILENAME} "
                         f"not found in {part_config_dir}. "
-                        "This is expected for older models (trained with version < 0.5). "
-                        "You will need to copy over the graph construction  "
+                        "This is expected for graph data processed with with version < 0.5. "
```
zhjwy9343, line 311:
```suggestion
                        "This is expected for graph data processed with version < 0.5. "
```

zhjwy9343 (approved):
LGTM except for one typo

PR #1320, opened 2025-08-07
[Bug] Fix feat_dim for multi-column feature transformation

jalencato:
*Issue #, if available:*

*Description of changes:*

* For GConstruct multi-column feature transformaion, current saved feat_dim_dict will only save the dimension of single column. Fix the code to save the concatenation of all the columns.
* Add the tests for multi-column feature transformation.

### DIFF
--- a/python/graphstorm/gconstruct/transform.py
+++ b/python/graphstorm/gconstruct/transform.py
@@ -1611,9 +1611,9 @@ def process_features(data, ops: List[FeatTransform], ext_mem_path=None):
 
             # Write feature dimension back to the feature config
             # Updates the feature dimension, provided as a tuple in op.feat_dim. If the feature name
-            # already exists, this logic assumes the features are being concatenated along the 0-th axis.
-            # It therefore constructs a new dimension tuple by accumulating the size of
-            # this first dimension and keeping all subsequent dimension sizes constant.
+            # already exists, this logic assumes the features are being concatenated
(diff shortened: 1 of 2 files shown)


python/graphstorm/gconstruct/transform.py
```diff
@@ -1609,8 +1609,12 @@ def process_features(data, ops: List[FeatTransform], ext_mem_path=None):
...
                     new_data[key] = val
 
-        # Write feature dimension back to the feature config
-        feat_dim_dict[op.feat_name] = op.feat_dim
+            # Write feature dimension back to the feature config
+            if op.feat_name not in feat_dim_dict:
+                feat_dim_dict[op.feat_name] = op.feat_dim
+            else:
+                feat_dim_dict[op.feat_name] = (op.feat_dim[0] + feat_dim_dict[op.feat_name][0],) + \
+                                        feat_dim_dict[op.feat_name][1:]
```
thvasilo, line 1617: Can we clarify what's happening here and the expected input/output? These are supposed to be integer values right?

jalencato: For two transformed value with Dimension (M, K, ..) and (N, K, ..), we directly concatenate the value, so the output will be (M + N, K, ..)

jalencato: Added comment for more explanation.

thvasilo (approved):
Changes LGTM now, should make dimension consistent for multi-col features.

PR #1321, opened 2025-08-07
[Bug] Fix real-time inference workflow when adding reverse edges

jalencato:
*Issue #, if available:*

*Description of changes:*
* Add `add_reverse_edge` flag to handle the reverse edge case for real time inference endpoint. It will resolve the model loading issue during real time inference.
* Update the version string for gconstruct/gsprocessing.
* Test successfully on sagemaker with acm data.

### DIFF
--- a/graphstorm-processing/graphstorm_processing/config/config_parser.py
+++ b/graphstorm-processing/graphstorm_processing/config/config_parser.py
@@ -360,7 +360,7 @@ class NodeConfig(StructureConfig):
         return self._node_col
 
 
-def update_gsprocessing_config(graph_config: Dict[str, Any], add_reverse_edge: bool = False):
+def update_gsprocessing_config(graph_config: Dict[str, Any], add_reverse_edges: bool = False):
     """Parses a GSProcessing JSON configuration dictionary and save flag.
     Automatically adjusting the node/edge type configurations for homogeneous graphs
     to ensure they are compatible with DGL requirements.
@@ -369,7 +369,7 @@ def update_gsprocessing_config(graph_config: Dict[str, Any], add_reverse_edge: b
     ----------
     graph_config : Dict[str, Any]
         A GSProcessing configuration dict, describing stored graph data.
-    add_reverse_edge: Bool
+    add_reverse_edges: Bool
         If add reverse edge
     """
     if "nodes" in graph_config and graph_config["nodes"] != [{}]:
@@ -398,7 +398,7 @@ def update_gsprocessing_config(graph_config: Dict[str, Any], add_reverse_edge: b
     else:
         graph_config["is_homogeneous"] = False
 
-    graph_config["add_reverse_edge"] = add_reverse_edge
+    graph_config["add_reverse_edges"] = add_reverse_edges
 
 
 def create_config_objects(graph_config: Dict[str, Any]) -> Dict[str, Sequence[StructureConfig]]:
--- a/graphstorm-processing/graphstorm_processing/config/config_parser.py
+++ b/graphstorm-processing/graphstorm_processing/config/config_parser.py
@@ -362,7 +362,7 @@ class NodeConfig(StructureConfig):
 
 def update_gsprocessing_config(graph_config: Dict[str, Any], add_reverse_edges: bool = False):
     """Updates input GSProcessing dictionary with runtime parameters to allow graph reconstruction.
-    
+
     Will update the edge and node type names to match DGL expecations when the graph is homogeneous,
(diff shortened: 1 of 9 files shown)


graphstorm-processing/graphstorm_processing/config/config_parser.py
```diff
@@ -360,14 +360,17 @@ def node_col(self):
         return self._node_col
 
 
-def update_dict_if_homogeneous(graph_config: Dict[str, Any]):
-    """Parses a GSProcessing JSON configuration dictionary and adjust the node/edge type
-    if it is a homogeneous graph to fit the DGL requirement.
+def update_gsprocessing_config(graph_config: Dict[str, Any], add_reverse_edges: bool = False):
+    """Parses a GSProcessing JSON configuration dictionary and save flag.
+    Automatically adjusting the node/edge type configurations for homogeneous graphs
+    to ensure they are compatible with DGL requirements.
```
thvasilo, line 366:
```suggestion
    """Updates input GSProcessing dictionary with runtime parameters to allow graph reconstruction.
    
    Will update the edge and node type names to match DGL expecations when the graph is homogeneous,
    and adds an `add_reverse_edges` boolean value to indicate whether reverse edges were created
    during graph construction.
```

graphstorm-processing/graphstorm_processing/config/config_parser.py
```diff
@@ -360,14 +360,17 @@ def node_col(self):
...
 
     Parameters
     ----------
     graph_config : Dict[str, Any]
         A GSProcessing configuration dict, describing stored graph data.
+    add_reverse_edges: Bool
+        If add reverse edge
```
thvasilo, line 373:
```suggestion
        Indicates whether reverse edges were added during processing. Used to recreate graph downstream.
```

tests/unit-tests/test_metadata.py
```diff
@@ -896,6 +900,12 @@ def test_config_json_santiy_check():
     with pytest.raises(AssertionError, match='A \"version\" field must be defined in the'):
         config_json_sanity_check(gcont_config_json)
 
+    gcont_config_json = build_gcons_json_example()
+    gcont_config_json['version'] = 'gconstruct-v1.0'
+    with pytest.raises(AssertionError, match='The \"version\" field must contain a \"runtime\" '
```
zhjwy9343, line 905: if commit the above change, need to change here too.

tests/unit-tests/test_metadata.py
```diff
@@ -906,6 +916,16 @@ def test_config_json_santiy_check():
     with pytest.raises(AssertionError, match='The value of \"is_homogeneous\" can only be'):
         config_json_sanity_check(gcont_config_json)
 
+    gcont_config_json = build_gcons_json_example()
+    gcont_config_json.pop('add_reverse_edges')
+    with pytest.raises(AssertionError, match='A \"add_reverse_edges\" field must be defined'):
```
zhjwy9343, line 921:
```suggestion
    with pytest.raises(AssertionError, match='An \"add_reverse_edges\" field must be defined'):
```

python/graphstorm/gconstruct/construct_graph.py
```diff
@@ -1049,6 +1049,39 @@ def print_graph_info(g, node_data, edge_data, node_label_stats, edge_label_stats
...
+
+def write_transformed_config(output_conf_file, output_dir, process_confs):
+    """
+    Write the transformed configuration file. If not specified by output_conf_file,
+    will write to the output graph directory.
+
+    Parameters:
+    ---------
+    output_conf_file: str
+        Path to the output configuration file, default: None
```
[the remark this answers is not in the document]
thvasilo: The function itself does not provide a default value, hence calling it without a value will trigger a runtime error, i.e. you can't do this

```python
write_transformed_config(output_dir, process_confs)
```

If you want the function to be callable without providing a value for `output_conf_file` you'll need to add a default value for it, otherwise, you can annotate the parameter as `Optional` which means it can take the value `None` as input, which is a different case and means you can call the function with


```python
write_transformed_config(None, output_dir, process_confs)
```
Here, since the intended "default" value is dynamic (depends on the output_dir argument) the correct annotation would be Optional, and you can note what happens when the user passes `None` in the input

zhjwy9343 (approved):
LGTM

python/graphstorm/gconstruct/construct_graph.py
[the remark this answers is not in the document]
jalencato: Redefine the function and fix the test in the latest commit.

PR #1333, opened 2025-08-30
[Feature] Add sanity check of tokenize transformation in real-time endpoint launch

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
This PR adds a sanity check of `tokenize_hf` transformation in graph construction JSON file. If detected, will stop launch and raise error message.

### DIFF
--- a/sagemaker/launch/launch_utils.py
+++ b/sagemaker/launch/launch_utils.py
@@ -19,6 +19,7 @@
 import logging
 import os
 import re
+import json
 import shutil
 import tarfile
 from argparse import ArgumentTypeError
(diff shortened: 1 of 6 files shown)


sagemaker/launch/launch_utils.py
```diff
@@ -238,3 +239,33 @@ def check_name_format(name_str):
...
+    Return
+    ------
+    has_tokenize_transform: bool
+        A boolean value. If the JSON oject contains at least one tokenize transformation, return
+        True, otherwise False.
+    """
+    tokenize_str = 'tokenize_hf'
+    # Convert json object to string, and then use string search
+    json_str = json.dumps(graph_config)
+    has_tokenize_transform = tokenize_str in json_str
```
thvasilo, line 268: This is a fragile implementation, we could have `tokenize_hf` in other parts of the config, e.g. a path or feature name. 

It's better to walk the JSON structure and specifically look for the key/value  `"name": "tokenize_hf"` within feature definitions.

thvasilo (changes requested):
We should use a more refined search for the transformation.

sagemaker/launch/launch_utils.py
zhjwy9343: If walk through the JSON structure, it will be same as the GraphStorm core code in `metadata.py`. Better not to duplicate the same function for only one field.

To make sure `tokenize_hf` is in transformation field, new implementation use regex instead of a simple `in` function.

thvasilo (approved):
LGTM now

Issue #1344, opened 2025-10-13
[Bug] The trained edge input layer is not saved and reloaded


zhjwy9343:
When use edge features, the `GSEdgeEncoderInputLayer` is trained. But in the `save_model()` function, only the Node encoder layer is saved. And during inference time, the `GSEdgeEncoderInputLayer` is restored by using the built-in initialization methods, but not able to load any parameters because not Edge encoder layers are saved. The model just use the initialized parameters for inference.

Suggested way to solve this issue:

1. in `models/utils.py`, modify the `save_model()` function to include edge encoders;
2. define a new constant to represent the encoder layer that include node encoder and edge encoder.

zhjwy9343:
Solved with the PR #1347 (https://github.com/awslabs/graphstorm/pull/1347)

PR #1345, opened 2025-10-15
[Bug] Add edge input layer saving and restoring

zhjwy9343:
*Issue #1344, if available:*

*Description of changes:*
This PR fixed the issue 1344 by adding the saving and restoring methods for edge input layer. It also added test cases for model saving and restoring.


tests/unit-tests/test_gnn.py
```diff
@@ -1059,8 +1059,10 @@ def create_mlp_node_model(g, lm_config):
```
jalencato: Do we test the save model - load model - do inference? I think it should change a lot to our embedded layer.

PR #1346, opened 2025-10-16
[Bug] Add saving and loading of edge encoder

zhjwy9343:
*Issue #1344, if available:*

*Description of changes:*
This PR fix the bug that no edge encoders were saved or loaded. To fix this issue, this PR break the backward compatibility in three perspectives:
1. Change the `embed_layer` in model saving from a nn.Module to be a dictionary in the format of `{'embed':'{'node_embed': node_input_encoder, 'edge_embed': edge_input_encoder}}` when do model saving and loading. This change is an internal API change. Users who directly call the  `save_model()` and `load_mode()` APIs in the `model.utils.py` will have compatibility issue. But for using `model.save_model(model_path)` and `model.load_model(model_path)`, it will not be impacted.
2. To avoid confusion of `embed`, `dense_embed` with the new `node_embed` and `edge_embed` naming, disable the `dense_embed` naming constant, and add `node_embed` and `edge_embed` constants.
3. Change the options of `restore_model_layer` argument from `'embed'|'dense_embed', 'gnn', 'decoder'` to `'node_embed', 'edge_embed', 'gnn', 'decoder'`. And correspondingly change the `restore_model()` behavior by removing the `embed` and `dense_embed` options.

**[Back Compatibility Break]**
1. No `embed` or `dense_embed` options in restore model layers.
2. Need to specify `node_embed` for restoring node input encoder module, and `edge_embed` for edge input module if have saved.

### DIFF
--- a/tests/unit-tests/test_gnn.py
+++ b/tests/unit-tests/test_gnn.py
@@ -1181,15 +1181,8 @@ def test_gnn_model_load_save():
         save_model(tmpdirname, embed_layer={'node_embed':model.node_input_encoder})
         model1 = copy.deepcopy(model)
         with pytest.raises(AssertionError, match="There is no edge_embed *"):
-            load_model(tmpdirname, embed_layer={"node_embed": model.node_input_encoder,
+            load_model(tmpdirname, embed_layer={"node_embed": model1.node_input_encoder,
                                                 "edge_embed": model1.edge_input_encoder})
-
-        # abnormal case 4, not save embed, but restore node_embed.
-        #                It will raise an Assertion error, saying not node_embed to load.
-        save_model(tmpdirname, gnn_model=model.gnn_encoder)
-        model1 = copy.deepcopy(model)
-        with pytest.raises(AssertionError, match="There is no edge_embedt *"):
-            load_model(tmpdirname, embed_layer={"node_embed": model.node_input_encoder})
 
     th.distributed.destroy_process_group()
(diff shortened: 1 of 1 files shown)


python/graphstorm/model/utils.py
```diff
@@ -81,27 +88,43 @@ def sparse_emb_initializer(emb):
...
+        Change the``embed_layer`` in v0.5.1 from a module into a dict with two elements to support
+        saving the edge encoder in a model.
+
+    Parameters
+    ----------
+    model_path: str
+        The path of the model is saved.
+    gnn_model: model
+        A (distributed) model of GNN
+    embed_layer: dict of model
```
classicsong: This is a breaking change.

classicsong: Why not just add an edge_embed_layer at the end of the arg-list? Then the code will be backward compatible.

tests/unit-tests/test_gnn.py
```diff
@@ -1121,7 +1125,10 @@ def test_gnn_model_load_save():
         np_data = GSgnnData(part_config=part_config,
                             node_feat_field='feat')
     model = create_mlp_node_model(g, lm_config)
-    dense_params = {name: param.data[:] for name, param in model.node_input_encoder.named_parameters()}
+    # add edge encoder to save and load
```
classicsong: Can you also add the node_emb only test?

PR #1347, opened 2025-10-17
[Bug] Add saving and loading of edge encoder

zhjwy9343:
*Issue #1344, if available:*

*Description of changes:*
This PR fix the bug that no edge encoders were saved or loaded. It adds a new argument `edge_embed_layer` of the `save_model()` and `load_mode()` APIs in the `.model.utils.py`. This change is an internal API change. The public facing API `model.save_model(model_path)` and `model.load_model(model_path)` will not be impacted.

**[Back Compatibility Break]**
To fix this issue, this PR breaks the backward compatibility in two perspectives:
- Change the options of `restore_model_layer` argument from `'embed'|'dense_embed', 'gnn', 'decoder'` to `'node_embed', 'edge_embed', 'gnn', 'decoder'`. And correspondingly change the `restore_model()` behavior by removing the `embed` and `dense_embed` options.
- To avoid confusion of `embed`, and `dense_embed` naming, disable the `dense_embed` naming constant, and add `node_embed` and `edge_embed` constants for `embed`.

### DIFF
--- a/python/graphstorm/model/utils.py
+++ b/python/graphstorm/model/utils.py
@@ -1564,24 +1564,20 @@ def load_model(model_path, gnn_model=None, embed_layer=None, decoder=None, edge_
 
     # some modules are non-parameterized, e.g., the Dot-Product decoder, hence
     # they were not saved, and no need to do assert check.
-    if gnn_model is not None:
-        if isinstance(gnn_model, nn.Module):
-            assert GRAPHSTORM_MODEL_GNN_LAYER in checkpoint, "There is no GNN module to be loaded."
+    if gnn_model is not None and isinstance(gnn_model, nn.Module):
+        assert GRAPHSTORM_MODEL_GNN_LAYER in checkpoint, "There is no GNN module to be loaded."
         gnn_model.load_state_dict(checkpoint[GRAPHSTORM_MODEL_GNN_LAYER])
-    if  embed_layer is not None:
-        if isinstance(embed_layer, nn.Module):
-            assert GRAPHSTORM_MODEL_NODE_EMBED_LAYER in checkpoint, ("There is no node encoder "
-                                                                     "module to be loaded.")
+    if  embed_layer is not None and isinstance(embed_layer, nn.Module):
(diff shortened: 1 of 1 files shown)


python/graphstorm/model/gnn.py
```diff
@@ -741,20 +743,27 @@ def unfreeze_input_encoder(self):
             self._edge_input_encoder.unfreeze()
 
     # pylint: disable=signature-differs
-    def restore_dense_model(self, restore_model_path,
-                            model_layer_to_load=None):
-        # TODO(zhengda) we need to load edge_input_encoder.
+    def restore_dense_model(self, restore_model_path, model_layer_to_load=None):
+        """ restore dense models
```
jalencato: What about adding .. versionchanged:: to the log here?

jalencato (commented):
One high level comment:
* Why do we need to specify the `node_embed` and `edge_embed` here? How about we reuse the `embed` and try to load node&edge encoder at the same time? (Do not load if there is no node/edge encoder saved)

python/graphstorm/model/gnn.py
zhjwy9343: Good call. Changed.

zhjwy9343:
> One high level comment:
> 
> * Why do we need to specify the `node_embed` and `edge_embed` here? How about we reuse the `embed` and try to load node&edge encoder at the same time? (Do not load if there is no node/edge encoder saved)

There are a few considerations about the current design:
1. edge encoder is optional in most of cases. If keep `embed` and by default load both node&edge encoder, we might see errors when there is no edge encoder to load. In addition, if the model has edge encoder but users do NOT want to load it on purpose, there is no way to perform this action.
2. if change `embed` from string to non-string, e.g., a dict, it will have backward compatibility. For the `load_model()` API, the current design does not.
3. the `embed` and `dense_embed` naming are confusing. So better to clarify with more concrete names although this will break backward compatibility.

classicsong (approved):
LGTM

PR #1363, opened 2025-12-05
[Enhancement] Support Learnable Embeddings for Real-time inference

zhjwy9343:
*Issue #, if available:*

*Description of changes:*
This PR adds support of a special node feature, `gs_embedding`, in the payload JSON for real-time inference case.

This works when the deployed GraphStorm model was trained with learnable embeddings in the two cases below:
1. some node types are featureless, hence GraphStorm automatically assigned learnable embeddings on them while training; and,
2. user set the `--use-node-embedding` configuration to be true while training.

After trained, these learnable embeddings will be saved under different folders named with node type names. In offline inference, GraphStorm will load them as part of model artifacts. Because size of these learnable embeddings are proportional to the number of nodes, it could be hundreds of MBs or a few GBs. Reload these learnable embeddings on SageMaker endpoints could cause potential risks, like OOM and OOD.

Major changes:
- create two new input layer encoders, `GSPureLearnableInputLayer4GraphFromMetaData`, and `GSNodeEncoderInputLayer4GraphFromMetadata` to handle models that cannot load learnable embeddings at SageMaker endpoints.
- modify the `prepare_batch_input()` method to extract the `gs_embedding` feature and prepare it for model's forward function.
- modify the `gsf.py` model building to replace node input layers when the initial graphs are coming from metadata. 
- modify the `process_json_payload_nodes()` function to allow payload graph to contains features that was not used in model training, which is a common case because model training uses `--node-feat-name` to control which features to be used.

Tests:
- in local SageMaker environment.
- in Amazon SageMaker environment.

### DIFF
--- a/python/graphstorm/gsf.py
+++ b/python/graphstorm/gsf.py
@@ -1226,6 +1226,7 @@ def set_encoder(model, g, config, train_task):
         gnn_encoder = SAGEEncoder(h_dim=config.hidden_size,
                                   out_dim=out_emb_size,
                                   num_hidden_layers=config.num_layers - 1,
+                                  edge_feat_name=config.edge_feat_name,
                                   dropout=dropout,
                                   aggregator_type='pool',
                                   num_ffn_layers_in_gnn=config.num_ffn_layers_in_gnn,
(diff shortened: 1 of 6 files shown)


jalencato (approved):
approve with minor comment.

